PR #6, opened 2020-08-06
PR Build / Unit Test + Functional Test workflows

jeffalder:
Adds a workflow for running Unit Tests in the PR Build workflow.

### DIFF
--- a/.github/workflows/unit_test.yml
+++ b/.github/workflows/unit_test.yml
@@ -1,0 +1,54 @@
+name: Java PR build (gradle)
+
+on:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: gradle/wrapper-validation-action@v1
+      # The first JDK set up gets to be "primary".
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1.4.0
+        with:
+          java-version: 8
+      - name: save JAVA_HOME as JDK8 for later usage
+        run: echo "::set-env name=ORG_GRADLE_PROJECT_jdk8::$JAVA_HOME"
+      # The second JDK is not first in the path but does reset JAVA_HOME
+      - name: Set up JDK 1.7
+        uses: actions/setup-java@v1.4.0
+        with:
+          java-version: 7
+      - name: save JAVA_HOME as JDK7 for later usage
+        run: echo "::set-env name=ORG_GRADLE_PROJECT_jdk7::$JAVA_HOME"
+      # Java is installed; document settings
+      - name: log project settings
+        run: |
+          echo "--- Java versions ---"
+          $ORG_GRADLE_PROJECT_jdk7/bin/java -version
+          $ORG_GRADLE_PROJECT_jdk8/bin/java -version
+      - name: correct JAVA_HOME
+        run: echo "::set-env name=JAVA_HOME::$ORG_GRADLE_PROJECT_jdk8"
+      # Restore the gradle cache
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          # The docs say to use hashfiles, but gradle itself is smart enough to
+          # re-download dependencies if it couldn't resolve them.
+          # Because our gradle cache is _so big_, I think a 90% cache is far
+          # better than a 0% cache.
+          key: ${{ runner.os }}-gradle
+          restore-keys: |
+            ${{ runner.os }}-gradle
+      # Start running the build.
+      - name: build the agent jar
+        run: ./gradlew --console=plain --parallel newrelicJar
+      - name: run the unit tests
+        run: ./gradlew --console=plain --parallel test -x functional_test:test -PnoInstrumentation --continue
+
+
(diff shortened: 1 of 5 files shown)


.github/workflows/unit_test.yml
```diff
@@ -0,0 +1,51 @@
...
+          # The docs say to use hashfiles, but gradle itself is smart enough to
+          # re-download dependencies if it couldn't resolve them.
+          # Because our gradle cache is _so big_, I think a 90% cache is far
+          # better than a 0% cache.
+          key: ${{ runner.os }}-gradle
+          restore-keys: |
+            ${{ runner.os }}-gradle
+      # Start running the build.
+      - name: run the unit tests
+        run: ./gradlew --console=plain --parallel test -x functional_test:test -PnoInstrumentation --continue
```
XiXiaPdx, line 50: It looks like this is running both functional tests and unit tests. Why is that?

.github/workflows/unit_test.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: save JAVA_HOME as JDK7 for later usage
+        run: echo "::set-env name=ORG_GRADLE_PROJECT_jdk7::$JAVA_HOME"
+      # Java is installed; document settings
+      - name: log project settings
+        run: |
+          echo "--- Java versions ---"
+          $ORG_GRADLE_PROJECT_jdk7/bin/java -version
+          $ORG_GRADLE_PROJECT_jdk8/bin/java -version
+      - name: correct JAVA_HOME
+        run: echo "::set-env name=JAVA_HOME::$ORG_GRADLE_PROJECT_jdk8"
```
XiXiaPdx, line 34: If jdk7 was first setup as "primary" and then jdk8, all subsequent steps would be under jd8.  Then would this step to Correct Java Home still be necessary?

.github/workflows/unit_test.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Set up JDK 1.7
+        uses: actions/setup-java@v1.4.0
+        with:
+          java-version: 7
+      - name: save JAVA_HOME as JDK7 for later usage
+        run: echo "::set-env name=ORG_GRADLE_PROJECT_jdk7::$JAVA_HOME"
+      # Java is installed; document settings
+      - name: log project settings
+        run: |
+          echo "--- Java versions ---"
```
XiXiaPdx, line 30: Confirming the reason for setting jdk7 and jdk8 gradle properties is because `java.gradle` script looks for them?

.github/workflows/unit_test.yml
XiXiaPdx: Using act, I've locally tested this current flow and also a different one with Jdk7 first, then jdk8.  They both run but I'm getting failed tests in both runs.

.github/workflows/unit_test.yml
jeffalder: > I am surprised ./gradlew --console=plain --parallel test runs both unit and functional tests. Didn't know that.

Yes - running an task like `test` that isn't qualified with a project runs that task in all the projects it exists.

> So is functional_test with -PnoInstrumentation a way to disable the functional tests? I see its applied to every instrumentation module...I don't get how that disables a functional test.

The parameters break down like this:
1. `-x functional_test:test` is standard gradle syntax for "don't run the `test` task in the `functional_test` project".
2. `-PnoInstrumentation` is something written in _our_ gradle files that says "don't run the `test` task in any project under `instrumentation`". 

These two options are distinct. The latter doesn't relate to functional tests. Does that answer your question?

.github/workflows/unit_test.yml
jeffalder: Yes. In fact, it requires those two properties.

.github/workflows/unit_test.yml
jeffalder: I quickly tested installing java 7 and then java 8; it seems like java 7 was first in the path when I ran `java`. Reversing the order means that java 8 seemed to appear on the path first, so I went with this order.

jeffalder: I'm curious what tests failed. It passed for me locally and it appeared to pass on GH as well.

.github/workflows/unit_test.yml
XiXiaPdx: Ok, that clears it up!

.github/workflows/unit_test.yml
XiXiaPdx: Ok, I think I got it now. I ran it again and printed out the path. Your intention was to have jdk8 first on path.

JAVA_HOME and path to jdk are two separate concerns. The JAVA_HOME is an environment variable other java applications in our tests rely on. 

I'm running it again now, with the jdk8 first on path, and I'm getting this exception during the run:

```
 com.newrelic.agent.datastore.DatastoreMetricsTest > testBuildInstanceIdentifier FAILED
|     java.net.UnknownHostException at DatastoreMetricsTest.java:65
|         Caused by: java.net.UnknownHostException at DatastoreMetricsTest.java:65
```

thrown by:

        `final String hostname = InetAddress.getLocalHost().getHostName();`

I'm looking into it and I'm guessing its my machine and not anything to do with the test.

jeffalder: > Ok, I think I got it now. I ran it again and printed out the path. Your intention was to have jdk8 first on path.

Yes, precisely. Thanks!

PR #8, opened 2020-08-07
Move THIRD_PARTY_NOTICES

### LINKED ISSUE #7: Move THIRD_PARTY_NOTICES to the root of the repo
THIRD_PARTY_NOTICES should be in the root of the repo so it's easily discovered when the repo is checked out.

It should continue to be included in the agent jar, so don't forget that.

jeffalder:
Moves to root of the checkout. After a clean build, I verified that the file was in the zip and in newrelic.jar.

Fixes #7


jeffalder:
@breedx-nr We don't need to include it with newrelic-api because that lib is only our code. The notices only apply to code we distribute, which is only distributed with the zip and with the agent jar itself.

PR #9, opened 2020-08-08
Fix gradle errors

jeffalder:
Gradle was occasionally tossing out a deprecation warning about resolving the runtime classpath of another module during the build step. I ended up significantly re-writing the process of gathering cached weave attributes.

**Current behavior**
If the `jar` task for an instrumentation module runs, there's a `doLast` step that reads the jar, determines the cached weave attributes, and adds those attributes to the jar via `ant.jar`. There are a few things that are problematic:
1. The gradle warning, obviously.
   > The configuration :instrumentation-build:runtimeClasspath was resolved without accessing the project in a safe manner.  This may happen when a configuration is resolved from a different project. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0. See https://docs.gradle.org/6.5/userguide/viewing_debugging_dependencies.html#sub:resolving-unsafe-configuration-resolution-errors for more details.

   > If your build has an unsafe access deprecation warning, it needs to be fixed. It’s a symptom of these bad practices and can cause strange and indeterminate errors.
2. I suspect this was the cause of the weird closure errors we were getting, since the warning was from a similar generated name, `instrumentation_bvauobm1yl2kg6l2i76gae4v4$_run_closure1$_closure3.doCall`
3. Tacked-on doLasts that edit the task output make it difficult to know what order things will happen in.

**New behavior**
There is a new task, `writeCachedWeaveAttributes`, which writes the cached weave attributes to `build/weaveAttributes/MANIFEST.MF`. This manifest is then merged with the jar manifest when the jar is built the first time.

To handle shadowed dependencies, there's a new configuration for modules to use that should have exactly one artifact. This is used by the postgres instrumentation, which shadows json-simple.

The new code is extensively commented, and I've added a README that describes the whole process.

To test this, I downloaded 5.14.0 and extracted all the instrumentation manifests. I ran this code with `clean newrelicJar`, extracted all those manifests, and compared them. Here's the code that checks:

```java
    public static void main(String[] args) throws Exception {
        Manifest mpre, mpost;
        try (InputStream is = new FileInputStream(args[0])) {
            mpre = new Manifest(is);
        }
        try (InputStream is = new FileInputStream(args[0].replace("prechange", "postchange"))) {
            mpost = new Manifest(is);
        }

        mpre.getMainAttributes().remove(new Attributes.Name("Ant-Version"));
        mpre.getMainAttributes().remove(new Attributes.Name("Created-By"));

        if (!mpre.equals(mpost)) {
            System.out.println(args[0] + " unequal :-( ");
        } else {
            System.out.print("x");
        }
    }
```

This did correctly fail when there were differences (I verified that manually). The only current difference is netty-reactor because a change snuck in when the license headers were added.

I feel very confident in this. You might approach this commit-by-commit to handle it in digestible pieces.


gradle/script/instrumentation.gradle
```diff
@@ -1,32 +1,4 @@
-// update manifest in the doLast phase of jar.
-jar {
-    dependsOn ':instrumentation-build:compileJava'
-
-    doLast {
-        def out = new ByteArrayOutputStream()
-        javaexec {
-            classpath = project(":instrumentation-build").sourceSets.main.runtimeClasspath
```
jeffalder, line 8: This is the problem - accessing a classpath from another project.

instrumentation/jdbc-postgresql-8.0-312.jdbc3/build.gradle
```diff
@@ -1,13 +1,4 @@
-buildscript {
-    repositories {
-        jcenter()
-    }
-    dependencies {
-        classpath 'com.github.jengelman.gradle.plugins:shadow:5.2.0'
-    }
-}
```
jeffalder, line 8: shadow is already on the build path from `buildSrc`. It's redundant here.

PR #15, opened 2020-08-11
Updated reactor-netty matches the latest versions

tspring:
This updates the instrumentation verifier since the [FluxRecieve](https://github.com/newrelic/newrelic-java-agent/blob/main/instrumentation/netty-reactor-0.8.0/src/main/java/reactor/netty/channel/FluxRecieve_Instrumentation.java#L22) instrumentation point was changed to match the latest versions of reactor-netty.


jeffalder:
I'm checking into the CLA.

PR #16, opened 2020-08-12
Add okhttp-4.4 instrumentation

jeffalder:
Changes from 4.0:
1. `getResponseWithInterceptorChain` is now `internal`, so the Java method name changed.
2. `RealCall` moved from `okhttp3` to `okhttp3.internal.connection`


ahornerr:
I was able to confirm this works correctly with OkHTTP 4.7.2. 

image

PR #20, opened 2020-08-14
Make jar collector asynchronous

jeffalder:
Safely delegate the work of analyzing jars to a separate thread.


newrelic-agent/src/test/java/com/newrelic/agent/service/module/TrackedAddSetSynchronizationTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+        StringBuilder builder = new StringBuilder("Missing expected values:");
+
+        for (String expected : expectedValues) {
+            if (!allDeltas.contains(expected)) {
+                builder.append(" ").append(expected);
+            }
+        }
+
+        assertEquals(builder.toString(), allDeltas, fullSet);
```
breedx-nr, line 106: This test goes deep!  The complexity of testing in this particular case I think is worth the trade-off.

PR #21, opened 2020-08-17
*DO NOT MERGE*.  Opening this in order to facilitate conversation about dynamic attach.

breedx-nr:
*DO NOT MERGE*.

This is being opened just to allow some discussion to happen.  Doing this will allow me to take notes inline.


breedx-nr:
I'm going to submit a review and then immediately close the PR.  This is in hopes of having some discussion points in the history without this being considered an active merge request.  We should also probably not reopen, but instead open a new one.

PR #28, opened 2020-08-24
Expose the mbean that contains the linking metadata.

breedx-nr:
We have a need in JFR to be able to fetch the entity guid from the agent.  This change exposes a new MBean that contains the linking metadata, so that the JFR Daemon may leverage the entity guid.

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/jmx/LinkingMetadata.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/jmx/LinkingMetadata.java
@@ -1,0 +1,19 @@
+package com.newrelic.agent.jmx;
+
+import com.newrelic.api.agent.Agent;
+import com.newrelic.api.agent.Logger;
+import com.newrelic.api.agent.NewRelic;
+
+import java.util.Map;
+import java.util.logging.Level;
+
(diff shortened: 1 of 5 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/jmx/LinkingMetadata.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.Map;
+import java.util.logging.Level;
+
+public class LinkingMetadata implements LinkingMetadataMBean {
+
+    @Override
+    public Map<String, String> readLinkingMetadata() {
+        Agent agent = NewRelic.getAgent();
+        Logger logger = agent.getLogger();
+        logger.log(Level.INFO, "JmxLinkingMetadata: Fetching linking metadata from the agent...");
```
jasonjkeller, line 16:
```suggestion
        logger.log(Level.INFO, "JMX LinkingMetadata: Fetching linking metadata from the agent...");
```

I guess we should be consistent with our logging and use `JMX LinkingMetadata`

jasonjkeller:
As long as the tests pass this looks good to merge.

PR #33, opened 2020-08-27
change version to create dynamic attach snapshot repo

XiXiaPdx:
setup dynamic attach snapshot repo


XiXiaPdx:
will merge after #34  is merged...otherwise, I think the publish action will run twice

PR #42, opened 2020-08-28
Add a test to confirm socket timeout behavior

breedx-nr:
I think that adding a unit test to cover the new behavior is helpful.  

This also lays the foundation of possibly leveraging mocks in the other tests in the future.


breedx-nr:
NOTE: This applies onto #41

tspring:
Thanks!

PR #54, opened 2020-09-15
Add newrelic ssl certificate if ca_bundle_path is not specified. Reso…

tspring:
The removal of nrcerts and use_private_ssl configuration in 6.0.0 is preventing applications with custom trust stores from connecting to New Relic. This change always adds the minimal *.newrelic.com SSL certificate needed to connect unless a specific ca_bundle_path is specified.


breedx-nr:
What if we pulled the cert as a built-time step?  Does that help to reduce the maintenance burden?

jasonjkeller (approved):
Approving this PR but not merging as there is still some ongoing discussion about the best path forward.

PR #70, opened 2020-09-21
Fixes #69 by adding addCustomParameter(String, Boolean) api

### LINKED ISSUE #69: API for addCustomParameter(String, Boolean)
We should add an API for NewRelic.addCustomParameter(String, Boolean). As far as I can tell, this has never existed because the original API predated support for Boolean event types

tspring:
See #69

### DIFF
--- a/agent-bridge/src/main/java/com/newrelic/agent/bridge/PublicApi.java
+++ b/agent-bridge/src/main/java/com/newrelic/agent/bridge/PublicApi.java
@@ -147,7 +147,7 @@ public interface PublicApi {
      * @param key Custom parameter key.
      * @param value Custom parameter value.
      */
-    void addCustomParameter(String key, Boolean value);
+    void addCustomParameter(String key, boolean value);
 
     /**
      * Add key/value pairs to the current transaction. These are reported in errors and transaction traces.
--- a/agent-bridge/src/main/java/com/newrelic/agent/bridge/PublicApi.java
+++ b/agent-bridge/src/main/java/com/newrelic/agent/bridge/PublicApi.java
@@ -142,6 +142,14 @@ public interface PublicApi {
     void addCustomParameter(String key, String value);
 
     /**
+     * Add a key/value pair to the current transaction. These are reported in errors and transaction traces.
+     *
+     * @param key Custom parameter key.
+     * @param value Custom parameter value.
+     */
+    void addCustomParameter(String key, Boolean value);
+
+    /**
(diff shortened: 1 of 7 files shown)


agent-bridge/src/main/java/com/newrelic/agent/bridge/PublicApi.java
```diff
@@ -141,6 +141,14 @@
      */
     void addCustomParameter(String key, String value);
 
+    /**
+     * Add a key/value pair to the current transaction. These are reported in errors and transaction traces.
+     *
+     * @param key Custom parameter key.
+     * @param value Custom parameter value.
+     */
+    void addCustomParameter(String key, Boolean value);
```
breedx-nr, line 150: Can we simplify this to `addCustomParameter(String key, boolean value)` to avoid the case of somebody accidentally passing null?  This makes for a stronger API in my opinion.

newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
```diff
@@ -237,6 +237,17 @@ public void addCustomParameter(String key, Number value) {
         attributeSender.addAttribute(key, value, "addCustomParameter");
     }
 
+    /**
+     * Add a key/value pair to the current transaction. These are reported in errors and transaction traces.
+     *
+     * @param key
+     * @param value @
```
breedx-nr, line 244: Fix or remove javadoc.

newrelic-api/src/main/java/com/newrelic/api/agent/NewRelic.java
```diff
@@ -234,6 +234,21 @@ public static void addCustomParameter(String key, Number value) {
...
+     * <p>
+     * <b>Note:</b> The key and value pair will only be reported if the key argument can be represented in 255 bytes
+     * when encoded with UTF-8 encoding.
+     * </p>
+     *
+     * @param key Custom parameter key.
+     * @param value Custom parameter value.
+     * @since 6.1.0
+     */
+    public static void addCustomParameter(String key, Boolean value) {
```
breedx-nr, line 249: Just like on the api, I would strongly prefer to use a primitive `boolean` here instead of `Boolean` to avoid the null case.

breedx-nr (changes requested):
Thanks for jumping on this so quickly.

breedx-nr (approved):

PR #73, opened 2020-09-25
Akka http fix

tspring:
Transaction context can get lost between Directives because of Akka-Https use CallbackRunnables. This change attempts to propagate that context by weaving the CallbackRunnable constructor to create a token and link it on the run() method.

Due to performance concerns, I added a trackCallbackRunnable api to the Tracer, so that we only create/link the token when coming directly from AkkaHttp. We may consider putting this behind a config and/or traversing the tracer stack and not just checking the current tracer.

Thoughts?

### DIFF
--- a/instrumentation/scala-2.12.0/src/main/java/scala/concurrent/impl/CallbackRunnable_Instrumentation.java
+++ b/instrumentation/scala-2.12.0/src/main/java/scala/concurrent/impl/CallbackRunnable_Instrumentation.java
@@ -34,7 +34,9 @@ public class CallbackRunnable_Instrumentation<T> {
 
     public CallbackRunnable_Instrumentation(final ExecutionContext executor, final Function1 onComplete) {
         Transaction transaction = AgentBridge.getAgent().getTransaction(false);
-        if (AgentBridge.activeToken.get() == null && transaction != null) {
+        if (AgentBridge.activeToken.get() == null &&
+                transaction != null &&
+                AgentBridge.getAgent().getTracedMethod().trackCallbackRunnable()) {
             token = transaction.getToken();
         }
     }
--- a/instrumentation/scala-2.12.0/src/main/java/scala/concurrent/impl/CallbackRunnable_Instrumentation.java
+++ b/instrumentation/scala-2.12.0/src/main/java/scala/concurrent/impl/CallbackRunnable_Instrumentation.java
@@ -47,8 +47,13 @@ public class CallbackRunnable_Instrumentation<T> {
      * Override the setter for "value" so we can replace it with our wrapped version
      */
 
+    @Trace(async = true, excludeFromTransactionTrace = true)
     public void value_$eq(Try<T> original) {
         Weaver.callOriginal();
+        if (token != null) {
+            token.linkAndExpire();
+            token = null;
+        }
         AgentBridge.TokenAndRefCount tokenAndRefCount = AgentBridge.activeToken.get();
         if (tokenAndRefCount == null) {
             Transaction tx = AgentBridge.getAgent().getTransaction(false);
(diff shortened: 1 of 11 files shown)


instrumentation/scala-2.12.0/src/main/java/scala/concurrent/impl/CallbackRunnable_Instrumentation.java
```diff
@@ -25,15 +29,31 @@
...
     /**
      * Override the setter for "value" so we can replace it with our wrapped version
      */
+
+    @Trace(async = true, excludeFromTransactionTrace = true)
     public void value_$eq(Try<T> original) {
         Weaver.callOriginal();
-
+        if (token != null) {
+            token.linkAndExpire();
```
breedx-nr, line 54: Is it safe for multiple threads to link and expire at the same time?  If not, there might be a race condition here and we may need to introduce a semaphore or other concurrent safety net.

newrelic-agent/src/main/java/com/newrelic/agent/tracers/AbstractTracer.java
```diff
@@ -50,6 +50,7 @@
     private String customPrefix = "Custom";
     // doesn't need to be thread safe since this flag affects the decision to registerAsync
     private Boolean trackChildThreads = null;
+    private Boolean trackCallBackRunnable = false;
```
breedx-nr, line 53: `AtomicBoolean` here to help with threading issues?

newrelic-agent/src/main/java/com/newrelic/agent/tracers/AbstractTracer.java
```diff
@@ -231,6 +232,24 @@ public boolean trackChildThreads() {
...
+        this.trackCallBackRunnable = shouldTrack;
+    }
+
+    @Override
+    public boolean trackCallbackRunnable() {
+        if (!this.trackCallBackRunnable) {
+            TracedMethod parent = this.getParentTracedMethod();
+            if (null == parent) {
+                return false;
+            } else {
```
breedx-nr, line 246: remove redundant `else`

newrelic-agent/src/main/java/com/newrelic/agent/tracers/AbstractTracer.java
```diff
@@ -231,6 +232,24 @@ public boolean trackChildThreads() {
...
     }
 
+    @Override
+    public void setTrackCallbackRunnable(boolean shouldTrack) {
+        this.trackCallBackRunnable = shouldTrack;
+    }
+
+    @Override
+    public boolean trackCallbackRunnable() {
+        if (!this.trackCallBackRunnable) {
```
breedx-nr, line 242: If you factor out the guts of the `if` into a new private method called 'parentTrackCallbackRunnable()` or similar, then this collapses to a one-liner like
```
    @Override
    public boolean trackCallbackRunnable() {
        return this.trackCallBackRunnable || this.parentTrackCallbackRunnable();
    }
```

breedx-nr (changes requested):
Looks pretty good.  We talked before about the idea that callbackrunnable sneaking all the way into the tracer feels a bit heavy handed, but it seems workable.  I also like that there is a config check, but did I miss where that config was added?

instrumentation/scala-2.12.0/src/main/java/scala/concurrent/impl/CallbackRunnable_Instrumentation.java
tspring: Callbackrunnables can only be executed once, so this shouldn't be an issue

instrumentation/scala-2.12.0/src/main/java/scala/concurrent/impl/CallbackRunnable_Instrumentation.java
```diff
@@ -25,15 +29,31 @@
 @Weave(originalName = "scala.concurrent.impl.CallbackRunnable")
 public class CallbackRunnable_Instrumentation<T> {
     private final Function1<Try<?>, Object> onComplete = Weaver.callOriginal();
-
     private Try<T> value = Weaver.callOriginal();
+    @NewField
+    private Token token;
```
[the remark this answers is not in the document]
tspring: NewFields are extension classes, not actual Java fields so this should be ok

newrelic-agent/src/main/java/com/newrelic/agent/tracers/AbstractTracer.java
tspring: I don't think this needs to be thread safe as the tracer/TransactionActivity are thread specific

breedx-nr (approved):
Looks good. Thanks for the explanation and changes.

PR #76, opened 2020-10-02
add check for timeout exception

gfuller1:
Also manually tested and works.

### DIFF
--- a/infinite-tracing/src/main/java/com/newrelic/ResponseObserver.java
+++ b/infinite-tracing/src/main/java/com/newrelic/ResponseObserver.java
@@ -76,7 +76,7 @@ public class ResponseObserver implements StreamObserver<V1.RecordStatus> {
     /**
      * Detects if the error received was a connection timeout exception. This can happen if the agent hasn't sent any spans for more than 15 seconds.
      */
-    protected boolean isConnectionTimeoutException(Throwable t) {
+    private boolean isConnectionTimeoutException(Throwable t) {
         return t instanceof StatusRuntimeException
                 && t.getMessage().startsWith("INTERNAL: No error: A GRPC status of OK should have been sent");
     }
--- a/infinite-tracing/src/main/java/com/newrelic/ResponseObserver.java
+++ b/infinite-tracing/src/main/java/com/newrelic/ResponseObserver.java
@@ -43,7 +43,10 @@ public class ResponseObserver implements StreamObserver<V1.RecordStatus> {
             return;
         }
 
-        logger.log(Level.WARNING, t, "Encountered gRPC exception");
+        if (!isConnectionTimeoutException(t)) {
+            logger.log(Level.WARNING, t, "Encountered gRPC exception");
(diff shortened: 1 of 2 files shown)


infinite-tracing/src/main/java/com/newrelic/ResponseObserver.java
```diff
@@ -70,6 +73,14 @@ private boolean isChannelClosing(Throwable t) {
         return t instanceof StatusRuntimeException && t.getCause() instanceof ChannelClosingException;
     }
 
+    /**
+     * Detects if the error received was a connection timeout exception. This can happen if the agent hasn't sent any spans for more than 15 seconds.
+     */
+    protected boolean isConnectionTimeoutException(Throwable t) {
```
breedx-nr, line 79: If this is protected only for the test to have access, let's make it package-private and add a comment or annotation saying that it's available just for testing.

infinite-tracing/src/test/java/com/newrelic/ResponseObserverTest.java
```diff
@@ -115,5 +118,18 @@ public void shouldTerminateOnALPNError() {
...
+        DisconnectionHandler disconnectionHandler = mock(DisconnectionHandler.class);
+        MetricAggregator metricAggregator = mock(MetricAggregator.class);
+
+        ResponseObserver target = new ResponseObserver(
+                metricAggregator,
+                mock(Logger.class),
+                disconnectionHandler, shouldRecreateCall);
+
+        Throwable exception = new StatusRuntimeException(Status.fromCode(Status.Code.INTERNAL).withDescription("No error: A GRPC status of OK should have been sent\nRst Stream"));
+        assertTrue(target.isConnectionTimeoutException(exception));
```
breedx-nr, line 132: There are a few permutations that are missing from the test suite:
* type is not a `StatusRuntimeException` (tho it might be tested above/below, I didn't check)
* type IS `StatusRuntimeException` but the message doesn't match.
Would be nice to have proper coverage of the new behavior.

breedx-nr (changes requested):
Why do we want to special case this?  I fully grok what it's doing, but I don't immediately see why this is desirable.

I also asked for a small/nitpicky change.

gfuller1:
> Why do we want to special case this? I fully grok what it's doing, but I don't immediately see why this is desirable.
> 
> I also asked for a small/nitpicky change.

What happens is after 15 seconds if we don't send anything to infinite tracing it closes the channel and sends us an ugly exception that we print to the logs as a WARNING. The agent simply reconnect after that and there's nothing wrong with that, that's just how it is handled. 

We don't want to scare customers with this log message when nothing is actually wrong. So far it has scared customers into filing tickets with us. 

We special case this error because we don't want to avoid printing other errors that may occur which are actually serious.

breedx-nr:
Ok cool, thanks for the clarification.

breedx-nr (approved):

PR #80, opened 2020-10-06
ignore SocketExceptions when attempting to connect to cloud providers

tspring:
I noticed we'd log a SocketException when attempting to connect to AWS for cloud usage data


gfuller1 (approved):

gfuller1 (changes requested):
Can you add a test for this specific exception? Looks like CloudUtilityTest already has a test for a different exception.

gfuller1 (approved):

PR #83, opened 2020-10-10
Support Play Framework 2.8

tkawachi:
After upgrading to Play 2.8 in our environment, web transactions are now shown as non-web transactions. This is a problem and I would like to fix it.
I'm not familiar with the source code of the newrelic-java-agent, but I've rewritten it based on the Play 2.7 code until it passes the test.
I'd appreciate advice on what else needs to be fixed.


tspring:
Hi @tkawachi , Thanks for PR! I'd expect the Play-2.7 instrumentation to work on 2.8 and It doesn't look like there are any functional differences between it and this. Do you have an example of where the existing instrumentation is failing for 2.8?

tkawachi:
I don't understand exactly why we went from a web transaction to a non-web transaction in our environment, but based on [this comment](https://github.com/newrelic/newrelic-java-agent/blob/a9e42780feadcd3c0b3f555d2bef4b0b13d95598/instrumentation/play-2.7/src/test/java/com/nr/agent/instrumentation/play27/ServerInstrumentationTest.java#L82-L83), I understand that it is netty or akka that is recording the web transaction, correct?
If that is the case, then this PR is changing the wrong place.

tspring:
Yes, the WebTransaction is not going to be set in the Play instrumentation and in most cases should be getting set in either Akka-Http or Netty instrumentation. 

I'd suggest first checking that the same instrumentation has applied when running on 2.7 and 2.8. You can [grep validated](https://github.com/newrelic/newrelic-java-agent/blob/58bb62acbc91606ec5cb2aa7b29ecba6ab5dacbc/newrelic-agent/src/main/java/com/newrelic/agent/instrumentation/weaver/AgentWeaverListener.java#L74) in the agent log at FINE to determine this.

You could also set a breakpoint [here](https://github.com/newrelic/newrelic-java-agent/blob/58bb62acbc91606ec5cb2aa7b29ecba6ab5dacbc/newrelic-agent/src/main/java/com/newrelic/agent/TransactionApiImpl.java#L214) to determine what is setting it on Play-2.7 and use that to try and figure out what changed in 2.8. 

It can all be a bit of guess work, but if you have a simple testapp I'd be happy to look at it as well. Let me know if you have any questions. I'll leave this open for a few days before closing.

PR #87, opened 2020-10-16
Generalize the DOCKER_NATIVE_DRIVER_W_SYSTEMD regex to account for ot…

tspring:
There can be more than one runtime. K8s with crio for example looks like this:
1:name=systemd:/system.slice/crio-67f98c9e6188f9c1818672a15dbe46237b6ee7e77f834d40d41c5fb3c2f84a2f.scope


gfuller1 (changes requested):
Can you add a test for this new case in DockerDataTest?

gfuller1 (approved):

PR #94, opened 2020-10-27
Fix scala/play2.6 support

junder31:
### Overview
Monitoring on scala/play2.6.18 was not working when I initially installed the newrelic agent.  I found that the later versions of play2.6 deprecated the getHandlerFor method on the Server trait and used the Server object similar to how the play2.7 integration works.  Since the signatures for the getHandlerFor vary between the trait and object I had to do some refactoring to support both methods.

### Testing
I added a new tests for play2.6.18 which uses the object method rather than the trait.

### Checks
[ ] Are your contributions backwards compatible with relevant frameworks and APIs? No
[ ] Does your code contain any breaking changes? No
[ ] Does your code introduce any new dependencies? No


tspring:
@junder31 Thanks for PR! Looks like the Server trait doesn't exist in versions 2.6.0-2.6.11, which causes the instrumentation to not match those versions. If there's another common instrumentation point you could match that would be awesome, otherwise I think we'll need to split the module into a version for 2.6.0 and one for 2.6.12+. There's a @SkipIfPresent API you can use to avoid both loading. [Here's an example](https://github.com/newrelic/newrelic-java-agent/blob/main/instrumentation/slick-3.0.0/src/main/scala/slick/util/AsyncExecutor.scala#L31)
Let us know if you've any questions

junder31:
@tspring Thanks for the feedback.  I couldn't identify any good option to make use of the @SkipIfPresent annotation.  I decided to opt for splitting the integration into multiple modules.  It looks like the first play2.6 version that matched the signature I instrumented was 2.6.13 which was also when the deprecation warning was added.  The existing integration should work for all 2.6.0-2.6.12 and 2.6.13+ all should be supported with the new module I have contributed.

tspring:
@junder31 Thanks again, we appreciate it! I'll take a look at this tomorrow hopefully.

tspring:
@junder31 Apologies for the confusion and lack of documentation surrounding our instrumentation Weaver APIs. The @SkipIfPresent annotation is necessary to prevent both play-2.6 and play-2.6.13 from applying at the same time. If you could add a class like this to the old 2.6 instrumentation it should work: 

```package play.core.server.common;

import com.newrelic.api.agent.weaver.SkipIfPresent;

@SkipIfPresent
public class ServerDebugInfo$ {
}
```
FWIW, this project is what we use to check instrumentation matching: https://github.com/newrelic/newrelic-gradle-verify-instrumentation.

instrumentation/play-2.6.13/src/main/scala/play/core/server/Server_Instrumentation.scala
```diff
@@ -0,0 +1,41 @@
...
+
+@ScalaWeave(`type` = ScalaMatchType.Object, `originalName` = "play.core.server.Server")
+class Server_Instrumentation {
+  @Trace(dispatcher = true)
+  def getHandlerFor(
+    request: RequestHeader,
+    applicationProvider: ApplicationProvider
+  ): Either[Future[Result], (RequestHeader, Handler)] = {
+    val tempResult: Either[Future[Result], (RequestHeader, Handler)] = Weaver.callOriginal()
+    val result = tempResult.right.map { case (request, handler) =>
```
tspring, line 27: I'm not very familiar with Scala or Play. Is this assignment of result always modifying the original and is that safe?

tspring (commented):
Would you mind rebasing this on a branch and commit message reflective of the change?

instrumentation/play-2.6.13/src/main/scala/play/core/server/Server_Instrumentation.scala
junder31: Most of the default operations and data structures in scala are immutable.  This operation will create a new object entirely and will not modify tempResult at all.

junder31:
> ServerDebugInfo

@tspring Thanks for the help.  Is there an easy way to identify candidate classes for the @SkipIfPresent annotation?

PR #101, opened 2020-11-01
Reintroduce use_private_ssl config to prevent adding ssl certs by def…

tspring:
We removed `use_private_ssl`  in 6.0 as part of the OSS effort. Unfortunately this required users to always provide their own certs. The change to add newrelic certs to the Truststore by default, has caused various SSL errors. use_private_ssl is now required to load the bundled certificates

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/transport/apache/ApacheSSLManager.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/transport/apache/ApacheSSLManager.java
@@ -9,6 +9,7 @@ package com.newrelic.agent.transport.apache;
 
 import com.google.common.collect.ImmutableList;
 import com.newrelic.agent.Agent;
+import com.newrelic.agent.config.DataSenderConfig;
 import org.apache.http.ssl.SSLContextBuilder;
 
 import javax.net.ssl.SSLContext;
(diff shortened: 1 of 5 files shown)


jasonjkeller (approved):

newrelic-agent/src/main/java/com/newrelic/agent/transport/apache/ApacheSSLManager.java
```diff
@@ -36,12 +37,16 @@
...
-    public static SSLContext createSSLContext(String caBundlePath) {
+    public static SSLContext createSSLContext(DataSenderConfig config) {
         SSLContextBuilder sslContextBuilder = new SSLContextBuilder();
         try {
-            if (caBundlePath != null) {
-                sslContextBuilder.loadTrustMaterial(getKeyStore(caBundlePath), null);
-            } else {
+            if (config.getCaBundlePath() != null) {
+                if (config.getUsePrivateSSL()) {
+                   Agent.LOG.log(Level.FINE, "Ignoring use_private_ssl config." +
```
breedx-nr, line 45: Shouldn't we log this at WARN or ERROR?

jasonjkeller: I think it is probably ok to leave at `FINE`. It's not an error, or even problematic, condition. It really just indicates that one config option took precedent over another.

PR #106, opened 2020-11-07
Work around occasional ClassCircularityError 

breedx-nr:
We sometimes infrequently see a `ClassCircularityError` when weaving, and it shows up when loading `java.util.concurrent.locks.ReentrantReadWriteLock$Sync$HoldCounter`.

This adds to existing work-arounds by forcing an early, deterministic classload for this class.

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/InitProblemClasses.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/InitProblemClasses.java
@@ -73,6 +73,15 @@ public class InitProblemClasses {
             } catch (Throwable e) {
                 // This might happen on non-sun JVMs
             }
+
+            // This works around a ClassCircularityError with the doubly nested inner of ReentrantReadWriteLock.
+            try {
+                String holdCounterClassName = "java.util.concurrent.locks.ReentrantReadWriteLock$Sync$HoldCounter";
+                Class.forName(holdCounterClassName);
+                Agent.LOG.log(Level.FINE, "Worked around loading class " + holdCounterClassName);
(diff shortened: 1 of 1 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/InitProblemClasses.java
```diff
@@ -73,6 +73,15 @@ public static final void loadInitialClasses() {
             } catch (Throwable e) {
                 // This might happen on non-sun JVMs
             }
+
+            // This works around a ClassCircularityError with the doubly nested inner of ReentrantReadWriteLock.
+            try {
+                String holdCounterClassName = "java.util.concurrent.locks.ReentrantReadWriteLock$Sync$HoldCounter";
+                Class.forName(holdCounterClassName);
+                Agent.LOG.log(Level.FINE, "Worked around loading class " + holdCounterClassName);
```
jasonjkeller, line 81: We probably don't need to log here.

jasonjkeller (approved):

newrelic-agent/src/main/java/com/newrelic/agent/InitProblemClasses.java
breedx-nr: Agreed, but I kept it at `FINE` to not cloud the logs.  Since it only happens once at startup, probably ok right?  I can remove it if you think it's unhelpful.

XiXiaPdx: I like the log

PR #109, opened 2020-11-10
Add support for scala 2.13 async operations

junder31:
### Overview
This adds support for async operations in scala 2.13.  In scala 2.12 the mechanism that was instrumented was CallbackRunnable.  In 2.13 this class no longer exists.  Transformation now preforms a similar role.  The instrumentation on Transformation is fairly similar to the CallbackRunnable instrumentation from previous version of the scala instrumentation.

### Testing
This passes the same tests that the older instrumentation provided.  I have also deployed it with one of my scala2.13 applications and verified that the transaction traces contain all the asynchronous work I expected to be included.

### Checks
Are your contributions backwards compatible with relevant frameworks and APIs? N/A
Does your code contain any breaking changes? N/A
Does your code introduce any new dependencies? Just scala2.13 which is being instrumented.


XiXiaPdx:
Hi @junder31 , nice work!!  For the app you emailed me, I now see `scala.concurrent.impl.Promise$Transformation.run()`  and `scala.concurrent.impl.Promise$Transformation.submitWithValue()` reported in the Transacton segments.  Just confirming this app is now running with your latest Scala2.13 async changes?

junder31:
@XiXiaPdx yes the app I sent you includes all the changes I have been working on.  That includes the akka-http change I have a PR open for, this PR, and one more that fixes an issue with the scalapb-grpc clients that I have not submitted a PR for.  These three changes have all the scala/play applications at my company instrumented quite well.  We do have some other scala teams that use different frameworks and libraries that I will be working on in the coming weeks/months that may result in some more PRs.

instrumentation/scala-2.13.0/src/main/java/scala/concurrent/impl/Transformation_Instrumentation.java
```diff
@@ -0,0 +1,108 @@
...
+
+@Weave(originalName = "scala.concurrent.impl.Promise$Transformation")
+public class Transformation_Instrumentation<T>  {
+    private Try<T> _arg = Weaver.callOriginal();
+    @NewField
+    private Token token;
+    @NewField
+    private AgentBridge.TokenAndRefCount tokenAndRefCount;
+
+    private Transformation_Instrumentation(final Function1 _fun, final ExecutionContext _ec, final Try<T> _arg, final int _xform)  {
```
tspring, line 35: Hi @junder31, I think you could remove this weave method. We instrumented the constructor in 2.12 due to an issue with Akka-http linking. It would also allow removing the token newfield and @Trace on submitWithValue, the latter of which is causing the ProducerConsumerWithFutures.testProducerConsumer test to fail with extra segments.

junder31: @tspring I made the changes you suggested.  I'll give it a shot on one of my apps and see if instrumentation is still working.  The test case at least passing now.

junder31: @tspring looks like I'm still getting the transactions segments I was expecting on my app as well with these changes.

tspring:
Thanks @junder31 , we appreciate your work on this. Last thing, I fixed the github actions issue preventing the tests from passing but it looks like you'll need merge/rebase main.

tspring (approved):

instrumentation/scala-2.13.0/src/main/java/scala/concurrent/impl/Transformation_Instrumentation.java
```diff
@@ -0,0 +1,92 @@
...
+public class Transformation_Instrumentation<T>  {
+    private Try<T> _arg = Weaver.callOriginal();
+    @NewField
+    private AgentBridge.TokenAndRefCount tokenAndRefCount;
+
+    /**
+     * Override the submitWithValue so we can replace it with our wrapped version.
+     * It plays similar role to the 2.12 setter for "value" on the old CallbackRunnable
+     */
+    public final Promise.Transformation submitWithValue(final Try<T> resolved) {
```
vpavkin, line 37: I might be wrong, but shouldn't we have this annotation here?
```scala
@Trace(async = true, excludeFromTransactionTrace = true)
```
This way the segments are reported and it allows to align the `ProducerConsumerWithFutures.testProducerConsumer ` test with how it looks for 2.12.
Currently it has zero segments reported.

tspring: Hi, the annotation with async=true should only be necessary when calling Token::link. For the 2.12 instrumentation we had to create a second token in the CallbackRunnable constructor to fix an issue Akka-Http. It's not clear to me the Java/scala.concurrent.impl.CallbackRunnable/value_$eq segments are valuable and ideally they wouldn't be included in traces either, but I'd be interested to hear if you find them useful?

junder31: I don't believe they are useful.  Until I worked on the scala instrumentation I had no idea that method even existed.  I think most scala users would actually find those segments confusing because they are completely unaware that futures work like that under the hood.

vpavkin: I don't have a strong opinion here. If for practical purposes it's better to leave these segments out - I'm totally fine with it. Just pointed out the difference in behavior here.

PR #111, opened 2020-11-10
upgrade to latest version of embedded mongo, assign embedded mongo ex…

jack-berg:
### Overview
The mongo instrumentation module tests use the `de.flapdoodle.embed.mongo` library to spin up and test again embedded mongo db instances. Connecting to older versions of embedded mongo (< 3.6.0) produces pesky firewall dialogs. The executables are assigned names with UUIDs in them so you're constantly re-prompted to acknowledge the same connections. 

This PR updates the configuration to assign stable names to the embedded mongo executables. This allows developers to acknowledge the firewall dialogs once and only once.

### Related Github Issue
None applicable. 

### Testing
The code changed is test code. I've verified that the changes produce the intended results and reduce the burden of testing the mongo instrumentation modules. 

### Checks

### DIFF
--- a/instrumentation/mongodb-2.12/build.gradle
+++ b/instrumentation/mongodb-2.12/build.gradle
@@ -9,7 +9,7 @@ dependencies {
     implementation(project(":newrelic-weaver-api"))
     implementation("org.mongodb:mongo-java-driver:2.12.3")
 
-    testImplementation("de.flapdoodle.embed:de.flapdoodle.embed.mongo:1.50.2")
+    testImplementation("de.flapdoodle.embed:de.flapdoodle.embed.mongo:3.0.0")
 }
 
 verifyInstrumentation {
--- a/instrumentation/mongodb-2.12/src/test/java/com/nr/agent/instrumentation/mongodb/MongoDb212Test.java
+++ b/instrumentation/mongodb-2.12/src/test/java/com/nr/agent/instrumentation/mongodb/MongoDb212Test.java
@@ -7,59 +7,89 @@
 
 package com.nr.agent.instrumentation.mongodb;
 
-import static org.junit.Assert.assertEquals;
-
-import java.io.IOException;
-import java.net.ServerSocket;
-import java.util.ArrayList;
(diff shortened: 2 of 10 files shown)


XiXiaPdx:
Hi @jack-berg , those prompts get old fast so thanks for improving the test!  

I tested this locally and I'm still getting re-prompted during tests. I do see the name has changed though, for instance:

`Do you want the application “extract-Db214-extractmongod” to accept incoming network connections?` 

Maybe it's my firewall settings? I'm wondering if you encountered this at all during your tests.

jack-berg:
> Maybe it's my firewall settings? I'm wondering if you encountered this at all during your tests.

Yes, you still have to click "Allow" when prompted with these dialogs, but assuming you do, they should stop recurring. 

One thing I noticed while testing is that I think you have to click "Allow" fairly quickly or it doesn't seem to stick. Not sure what this is about, but I bet its due to the fact that if you don't click quickly the executable may already be shut down. 

Also, I did some research trying to figure out if there was any way to manually allow connections to a particular application by name. The firewall options only allow you to select Applications from the filesystem, and not manually type in executable names, which is doesn't work for us because the executables we're trying to allow are embedded. Anyway, I was hoping to find some way to manage firewall exceptions via the terminal, but couldn't find anything.

XiXiaPdx:
> One thing I noticed while testing is that I think you have to click "Allow" fairly quickly or it doesn't seem to stick. Not sure what this is about, but I bet its due to the fact that if you don't click quickly the executable may already be shut down.

Ahhh, that did it! I was not clicking fast enough. 

I really appreciate your detailed comment in the test. Would you mind adding a callout for the "fairly quickly" requirement? I can see it popping up for others in the future.

jack-berg:
> Would you mind adding a callout for the "fairly quickly" requirement? I can see it popping up for others in the future.

Noted and have pushed an additional commit with an improved comment @XiXiaPdx

XiXiaPdx (approved):
LGTM 🍏

PR #129, opened 2020-11-28
Wrong URL reporting for Spring WebClient when multiple threads make requests to different URLs in parallel

veklov:
The behavior being fixed: When multiple threads make HTTP calls to several URLs in parallel, the stats reported to NewRelic for one call may be recorded with URL from another call.

Root cause: The logic capturing request URLs was passing the URLs to the reporting logic (segment.reportAsExternal) using a static variable. Multiple threads were able to write and read that variable in parallel. Since a thread wrote its request URL to the variable, but before it read it back, other threads were able to overwrite it.

### DIFF
--- a/instrumentation/spring-webclient-5.0/src/main/java/com/nr/agent/instrumentation/spring_webclient/Util.java
+++ b/instrumentation/spring-webclient-5.0/src/main/java/com/nr/agent/instrumentation/spring_webclient/Util.java
@@ -39,6 +39,9 @@ public class Util {
     }
 
     public static Mono<ClientResponse> reportAsExternal(ClientRequest request, Mono<ClientResponse> response) {
+        if(request == null) {
+            return response;
+        }
         Segment segment = (Segment) request.attribute(SEGMENT_ATTRIBUTE).orElse(null);
         if (segment == null) {
             return response;
--- a/instrumentation/spring-webclient-5.0/src/main/java/com/nr/agent/instrumentation/spring_webclient/Util.java
+++ b/instrumentation/spring-webclient-5.0/src/main/java/com/nr/agent/instrumentation/spring_webclient/Util.java
@@ -8,11 +8,15 @@
 package com.nr.agent.instrumentation.spring_webclient;
 
 import com.newrelic.agent.bridge.AgentBridge;
+import com.newrelic.agent.bridge.Transaction;
 import com.newrelic.api.agent.GenericParameters;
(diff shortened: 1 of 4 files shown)


veklov:
I have just realized that new implementation expands URL templates and this may not be appropriate for newrelic (too many unique URLs)

veklov:
I have checked that the original implementation did not preserve the templated URL too. So it was a false alarm. It looks like the below annotation of external calls with target URL in templated form happens on the NewRelic side by matching data reported by client and server agents:
image

tspring:
@veklov Thanks for contribution! Would you be willing to add a test to WebClientTest that makes multiple requests to different URLs in parallel?

veklov:
@tspring Sure, I have added ConcurrencyTest. As with all such tests, it may require tuning to work in all environments. If it does not fail on the main branch in your env, please try to increase the number of iterations each thread does or add more tasks/threads.

instrumentation/spring-webclient-5.0/src/main/java/com/nr/agent/instrumentation/spring_webclient/Util.java
```diff
@@ -22,35 +26,49 @@
...
+        Transaction txn = AgentBridge.getAgent().getTransaction(false);
+        if (txn != null) {
+            Segment segment = txn.startSegment("WebClient.exchange");
+            segment.addOutboundRequestHeaders(new OutboundRequestWrapper(request));
+            request.attribute(SEGMENT_ATTRIBUTE, segment);
+        }
+    }
+
+    public static Mono<ClientResponse> reportAsExternal(ClientRequest request, Mono<ClientResponse> response) {
+        Segment segment = (Segment) request.attribute(SEGMENT_ATTRIBUTE).orElse(null);
```
tspring, line 42: I think you may need to do a null check on request, since this will run before the original code

tspring (changes requested):
Looks good, thanks again for your work on this!

instrumentation/spring-webclient-5.0/src/main/java/com/nr/agent/instrumentation/spring_webclient/Util.java
veklov: @tspring done

tspring (approved):
Looks good, thanks again for your work on this!

veklov:
@tspring Thank you for prompt review!

PR #133, opened 2020-12-04
cleanup a handful of small things

jack-berg:
### Overview
I noticed a few issues small opportunities for improvement while I was perusing the code. 

### Checks


newrelic-agent/src/main/java/com/newrelic/agent/Transaction.java
```diff
@@ -985,10 +975,6 @@ public TransactionTimer getTransactionTimer() {
     private void finishTransaction() {
         try {
             synchronized (lock) {
-                if (Agent.LOG.isFinestEnabled()) {
-                    threadAssertion();
-                }
```
jack-berg, line 990: All `threadAssertion()` does is assert that the current thread has a lock on `lock`, so calling it immediately after `synchronized (lock)` is redundant.

tspring (approved):

Issue #126, opened 2020-11-24
Spring Instrumentation Transaction Name is set to URI variable name from RequestMapping instead of the value


tbradellis:
## Description
The transaction name is set to the variable name for @RequestMapping instead of the resolved endpoint defined in application.properties or other config file.

## Steps to Reproduce

define a variable in `application.properties` for a Spring Boot Project

example:
```
    requestmapping.path=hello

```
Create a controller using a URI path variable

```

    @RestController
    public class HelloController {

	    @RequestMapping(path = "/${requestmapping.path}")
	    public String index() {
		    return "Greetings from Spring Boot!";
	    }

    }

```
Download and install the agent.
Run the application with the Java Agent.

## Expected Behavior
Transaction Name should resolve to the variable value not the string variable name.

## additional info
New Relic APM UI:

<img width="728" alt="varnameTransName" src="https://user-images.githubusercontent.com/7401258/100138396-d3bc8a00-2e42-11eb-9d06-877493e7ecff.png">

Debug

<img width="1270" alt="methodPath" src="https://user-images.githubusercontent.com/7401258/100138399-d4552080-2e42-11eb-89e8-644902290a9b.png">
<img width="1259" alt="helloPath" src="https://user-images.githubusercontent.com/7401258/100138405-d61ee400-2e42-11eb-8326-bf4955af537f.png">

Seeing this in spring-4.3 module, but this looks like it will replicate in earlier versions as well.
## Additional context
Can be easily replicated with a simple spring boot app as described above.
https://start.spring.io/

Somewhere here it looks like that we only read the value on RequestMapping, but it never resolves.
See debugging screenshot...

It seems that `methodPath` never resolves to anything:
https://github.com/newrelic/newrelic-java-agent/blob/ec16239ba1a498b3743654fa1d727c8818eac3e0/instrumentation/spring-4.3.0/src/main/java/com/nr/agent/instrumentation/SpringControllerUtility.java#L20

https://github.com/newrelic/newrelic-java-agent/blob/ec16239ba1a498b3743654fa1d727c8818eac3e0/instrumentation/spring-4.2.0/src/main/java/com/nr/agent/instrumentation/SpringControllerUtility.java#L55

https://github.com/spring-projects/spring-framework/blob/d8dafbc49d66738a20e389336ed8359bb7ed2c95/spring-web/src/main/java/org/springframework/web/bind/annotation/RequestMapping.java#L99

[...]

PR #146, opened 2020-12-10
add weave to get uri placeholder value

tbradellis:
### Overview
Describe the changes present in the pull request
Weave spring UrlPathHelper to set the Transaction Name to the actual path for a request when a path placeholder is used on `RequestMapping` 
e.g. `RequestMapping(path= "/${...}")`
https://docs.spring.io/spring-framework/docs/current/reference/html/web.html#mvc

To use the placeholder value instead of the String value `"/${...}")' a configuration option is required.
Requires configuration to change from previous behavior.  Default will not change previous behavior.

Adds optional configuration to use property file defined URI.

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/126

### Testing
ran existing instrumentation and functional tests
Validated fix with local Spring App

### Checks

[* ] Are your contributions backwards compatible with relevant frameworks and APIs?
Currently only changes spring-4.3.0
[* ] Does your code introduce any new dependencies? Please describe.
Yes.
`compile group: 'javax.servlet', name: 'javax.servlet-api', version: '3.0.1'`

transactionName

### DIFF
--- a/instrumentation/spring-4.3.0/src/main/java/com/nr/agent/instrumentation/SpringControllerUtility.java
+++ b/instrumentation/spring-4.3.0/src/main/java/com/nr/agent/instrumentation/SpringControllerUtility.java
@@ -79,9 +79,11 @@ public class SpringControllerUtility {
         if (rootPath == null && methodPath == null) {
             AgentBridge.getAgent().getLogger().log(Level.FINE, "No path was specified for SpringController {0}", matchedAnnotationClass.getName());
         } else {
-            String fullPath = SpringControllerUtility.getPath(rootPath, methodPath, httpMethod);
-            transaction.setTransactionName(TransactionNamePriority.FRAMEWORK_HIGH, true, "SpringController",
-                    fullPath);
+            if(!SpringPlaceholderConfig.springPlaceholderValue){
+                String fullPath = SpringControllerUtility.getPath(rootPath, methodPath, httpMethod);
+                transaction.setTransactionName(TransactionNamePriority.FRAMEWORK_HIGH, true, "SpringController",
+                        fullPath);
+            }
         }
     }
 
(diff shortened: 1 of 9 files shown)


instrumentation/spring-4.3.0/src/main/java/com/nr/agent/instrumentation/SpringControllerUtility.java
```diff
@@ -17,6 +17,8 @@
 
 public class SpringControllerUtility {
 
+    public static String placeholderValue;
```
jack-berg, line 20: This strategy isn't threadsafe and needs some rethinking.

The flow works as follows:

`UrlPathHelper#getLookupPathForRequest` calls `SpringControllerUtility#setPlaceholderValue`, which sets `SpringControllerUtility.placeholderValue` to be path.

Later in the flow, `SpringControllerUtility#getPath` checks if `SpringControllerUtility.placeholderValue` has a value, and if so, uses it to construct the transaction path.

The problem is that multiple threads are sharing the single static instance of `SpringControllerUtility#placeholderValue` to maintain state. I've spun up a test app locally and setup a test to execute requests in parallel and have demonstrated that this produces corrupt transactions.

The logic also breaks if only a single thread is used to make requests. Suppose you have two routes `GET /route1` and `GET /${ENV_VAR}`, and an env var named `ENV_VAR` is set to `route2` so that you effectively have `GET /route2`. If you call `GET /route2`, `SpringControllerUtility#placeholderValue` will be set to `route2`, and subsequent calls to `GET /route1` will see the value of `route2` in `SpringControllerUtility#placeholderValue` and report their transaction path as `/route2`, which is not good.

jack-berg (changes requested):
In order for this type of strategy to work, you need to store the transaction path in a way that is thread safe.

instrumentation/spring-4.3.0/src/main/java/com/nr/agent/instrumentation/SpringControllerUtility.java
tbradellis: Ah, understand. Thank you for the feedback!

tbradellis:
Womp. Will investigate.

tbradellis:
@jack-berg I've still got some issues to sort out, but does this look like it's going in the right direction to you?

tbradellis:
oh gosh..nm.  I see this still does not solve the concurrency problem.

tbradellis:
Adding some tests for this.

PR #148, opened 2020-12-10
Make sure NettyDispatcher gets initialized

tspring:
When testing the spring-webflux-5.3.1 changes, @jack-berg and I found that the initial Transaction created by Netty-4.0.8 instrumentation was never getting created. The issue is that AbstractServerBootstrap isn't always used, consequently never triggering the (re)instrumentation of NettyDispatcher::channelRead , which is the Transaction start poing

### DIFF
--- a/instrumentation/netty-4.0.8/src/main/java/io/netty/bootstrap/NettyDispatcher.java
+++ b/instrumentation/netty-4.0.8/src/main/java/io/netty/bootstrap/NettyDispatcher.java
@@ -16,6 +16,7 @@ import com.newrelic.api.agent.TracedMethod;
 import io.netty.channel.ChannelHandlerContext_Instrumentation;
 import io.netty.handler.codec.http.HttpRequest;
 
+import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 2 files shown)


instrumentation/netty-4.0.8/src/main/java/io/netty/bootstrap/NettyDispatcher.java
```diff
@@ -26,12 +27,14 @@
 public class NettyDispatcher {
 
     private static volatile NettyDispatcher instance = null;
+    public static volatile AtomicBoolean instrumented = new AtomicBoolean(false);
```
jack-berg, line 30: Let's make this final.

jack-berg (approved):
I've tested this and it works.

instrumentation/netty-4.0.8/src/main/java/io/netty/bootstrap/NettyDispatcher.java
tspring: Ah, good call 👍

PR #149, opened 2020-12-11
Akka http core 2 13 10 2

tspring:
This is largely a copy/paste of akka-http-core-2.13_10.1.8, only it weaves [singleRequest](https://github.com/newrelic/newrelic-java-agent/compare/akka-http-core-2_13_10_2?expand=1#diff-027192b84c806f33d3a82cbe9fde34c2c4d148bcb0096ddf7f40c17f589424f8R55) as the singleRequestImpl was removed in 10.2.0 . See also https://github.com/newrelic/newrelic-java-agent/tree/main/instrumentation/akka-http-core-10.2.0


instrumentation/akka-http-core-2.13_10.2.0/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+
+verifyInstrumentation {
+    fails('com.typesafe.akka:akka-http-core-experimental_2.13:[1.0,10.1.8)')
+    passesOnly('com.typesafe.akka:akka-http-core_2.13:[10.2.0-RC1,)') {
+        compile("com.typesafe.akka:akka-stream_2.13:2.5.23")
+    }
+    fails('com.typesafe.akka:akka-http-core_2.12:[10.2.0-RC1,)') {
+        compile("com.typesafe.akka:akka-stream_2.12:2.5.23")
+    }
+
```
jack-berg, line 23:
```suggestion
```

tspring:
#152

jack-berg (approved):

ryanb93:
Does this mean it will _only_ work for `10.2.0` or would `10.2.2` work as well?

tspring:
@ryanb93 The instrumentation should match on 10.2.2 as well. Each module's build.gradle contains a verifyInstrumentation block that checks the matched versions like here: https://github.com/newrelic/newrelic-java-agent/pull/149/files#diff-e4cdea3b7561b480821e0b8e0229e081549a591277fb2627b7a4abf68f42e3bcR17

The syntax is a range where `[` is inclusive and `)` is exclusive. So currently this instrumentation will apply to all akka-http-core_2.13 versions starting with 10.2.0-RC1.

If you aren't getting transactions I'd suspect the problem is that your code is not going through either of these A[kkaHttp methods ](https://github.com/newrelic/newrelic-java-agent/pull/149/files#diff-94b8661e5aa87320e45eb226717dea594d2e4274e589c3c7627a8800d5139fb7R21-R29). 

If you don't mind, identifying the correct transaction start point or filing an issue with a simple example project that demonstrates the problem would be a great help.

ryanb93:
I'm on `akka-http` version `10.2.2` and set the logging to finest, I see this in the logs:

```
2021-01-15T16:22:14,228+0000 [1 22] com.newrelic FINEST: Skipping instrumentation module com.newrelic.instrumentation.akka-http-core-10.2.0. The most likely cause is that com.newrelic.instrumentation.akka-http-core-10.2.0 shouldn't apply to this application.
```

PR #162, opened 2020-12-16
fix scala test compile flakes by forcing zinc scala version in select…

jack-berg:
The builds have been flakely lately with several modules' test scala compilation failing. After quite a bit of digging and trial and error, I've determined that the root cause is that for certain modules, the version of scala that the zinc incremental compiler uses somehow is resolved incorrectly. The scala version resolution failure for zinc is not deterministic, and seems isolated to just a couple modules.

If we see this type of error again, we should be able to fix it in a similar fashion by forcing the scala library version used for the `zinc` configuration.


build.gradle
```diff
@@ -36,6 +36,7 @@ subprojects {
     repositories {
         mavenLocal()
         mavenCentral()
+        jcenter()
```
jack-berg, line 39: I routinely get resolution errors locally and adding jcenter() resolves them. I see that jcenter() is added as a repository elsewhere in the project, so I don't see any issue in adding another option for resolving dependencies.

instrumentation/akka-http-core-10.0/build.gradle
```diff
@@ -9,7 +9,6 @@ dependencies {
     implementation("org.scala-lang:scala-library:2.11.12")
     implementation("com.typesafe.akka:akka-http-core_2.11:10.0.0")
     implementation("com.typesafe.akka:akka-actor_2.11:2.3.14")
-    testImplementation(project(":instrumentation:akka-2.2")) { transitive = false }
```
jack-berg, line 12: Unused.

instrumentation/akka-http-core-2.11_10.0.11/build.gradle
```diff
@@ -8,8 +8,6 @@ dependencies {
     implementation(project(":agent-bridge"))
     implementation("com.typesafe.akka:akka-http-core_2.11:10.0.11")
     implementation("com.typesafe.akka:akka-stream_2.11:2.5.11")
-
-    testImplementation(project(":instrumentation:akka-2.2")) { transitive = false }
```
jack-berg, line 12: Unused.

instrumentation/akka-http-core-10.0.11/src/main/scala/akka/http/DefaultParsingErrorHandler.scala
```diff
@@ -1,8 +0,0 @@
-package akka.http
-
-import com.newrelic.api.agent.weaver.SkipIfPresent
-
-@SkipIfPresent
-class DefaultParsingErrorHandler {
-
-}
```
jack-berg, line 8: This file was a remnant from a deleted module.

XiXiaPdx (approved):
Great find!!! Needle in a haystack.  

Any idea how it may have been introduced?

tspring (approved):

Issue #164, opened 2020-12-17
Improve backoff strategy for Infinite Tracing


XiXiaPdx:
## Is your feature request related to a problem? Please describe.
The current backoff strategy is a repeating 15 second backoff in `DefaultBackOffPolicy` of Infinite Tracing.

## Feature Description
As suggested in the [agent spec](https://source.datanerd.us/agents/agent-specs/blob/master/Infinite-Tracing.md#failed_precondition), for the `Failed_Precondition` error, the backoff strategy should be the same as connect requests. 

The Java Agent has a [working implementation](https://github.com/newrelic/newrelic-java-agent/blob/6d3dceb8143c639331eb3ecb40fc2cbdab8720c1/newrelic-agent/src/main/java/com/newrelic/agent/rpm/RPMConnectionServiceImpl.java#L105-L116) of the back off strategy.

Issue #160, opened 2020-12-16
Respond to `Failed_Precondition` response code by disconnecting and connecting to Trace Observer


XiXiaPdx:
## Feature Description

The java agent should respond to a 8T `Failed_Precondition` response status by closing the current channel fully and starting a new channel by resolving the DNS record to the Trace Observer again. This will shift the agent's traffic to a different cell.

If the first attempt to reconnect with a new channel is unsuccessful, subsequent failures should retry with a backoff strategy.  

While reconnecting, spans should be stored and then sent when reconnected.

More info in [the current agent spec](https://source.datanerd.us/agents/agent-specs/blob/master/Infinite-Tracing.md#failed_precondition). 

## Additional context
Reason and research here - #157

Issue #157, opened 2020-12-15
Research and produce stories for 8T polish EPIC


XiXiaPdx:
Started 12/14 . research #134 

produce stories and analysis for the work to be done

XiXiaPdx:
There are two pieces of work surfaced in #134 

1.  Add additional metadata to each outbound span or spanbatch for Infinite Tracing.

This is already code-complete ✅  Here is where [we grab the map and pass it to Infinite Tracing](https://github.com/newrelic/newrelic-java-agent/blob/70950a1c7bd4fc4bd5de469ac965995c829868e4/newrelic-agent/src/main/java/com/newrelic/agent/RPMService.java#L279-L281).  

2.  Agent must respond to 8T response code by connecting to new IP address for Trace Observer

[From this agent spec](https://source.datanerd.us/agents/agent-specs/blob/master/Infinite-Tracing.md#failed_precondition) for infinite tracing, the java agent should respond to a 8T `Failed_Precondition` response status by closing the  channel fully and starting a new channel to new DNS record(IP address) for the same Trace Observer host name.  This will shift the agent's traffic to a different cell. 

In order to shift agent traffic to a different cell, the 8T team will change the DNS record (IP address) for Trace Observers. 8T will then send a `Failed Precondition` status code to agents and that will be the signal for agents to connect to the new IP address.  

With Java, the one challenge I see to successfully connecting to the new IP address is DNS Time to Live (aka TTL or DNS caching) by the JVM. The TTL could be set to _ forever _ and the agent will never resolve to the new IP Address. 

For example:

The first time the JVM is asked to resolve `nr.trace.observer`  to an IP address to connect to 8T and send spans,  

* it will look in the cache and fail to find the IP because it is the first time resolving `nr.trace.observer`
* it will obtain the IP address from its default name service and store the value in the cache.  
* It connects, sends spans.

Sometime later, 8T wants to shift agent traffic to a different cell. 

* 8T sets a new IP address record for `nr.trace.observer`. 
* 8T sends the `Failed Precondition` signal to agents.
* The agent disconnects and tries to reconnect to `nr.trace.observer`.  
* JVM will resolve `nr.trace.observer`  and looks in the cache. 
* Since the cache is set to _forever_, the old IP address is used again.  
* The agent will fail to connect because of the bad IP and repeatedly fail according to the backoff policy. 

[...]

PR #168, opened 2020-12-18
Infinite tracing failed precondition

XiXiaPdx:
### Overview

When the agent receives a `Failed_Precondition` gRPC response code from Infinite Tracing, the agent will disconnect to the Trace Observer by closing the existing channel fully and clear the stream observer. The agent will then immediately try to reconnect to the Trace Observer. The reason we are doing this is here #157 

General code context:

The [entry point for building](https://github.com/newrelic/newrelic-java-agent/blob/6d3dceb8143c639331eb3ecb40fc2cbdab8720c1/newrelic-agent/src/main/java/com/newrelic/agent/service/ServiceManagerImpl.java#L227) Infinite Tracing from the agent side.
When [agent successfully connects](https://github.com/newrelic/newrelic-java-agent/blob/6d3dceb8143c639331eb3ecb40fc2cbdab8720c1/newrelic-agent/src/main/java/com/newrelic/agent/RPMService.java#L278-L282), that's when  Infinite Tracing [actually starts.](https://github.com/newrelic/newrelic-java-agent/blob/6d3dceb8143c639331eb3ecb40fc2cbdab8720c1/newrelic-agent/src/main/java/com/newrelic/agent/service/UpdateInfiniteTracingAfterConnect.java#L28-L34)

Once it starts, `LoopForever` gets looping and regularly checks `ConnectionStatus` to determine if a channel and stream observer needs to be built. It does this by going through a couple of layers to `ChannelSupplier`

After a channel and stream observer are connected, one way they are disconnected is due to receiving certain status codes. The status codes are received in `ResponseObserver` and the `DisconnectionHandler` is the bulk of the logic to change connection status. 

### Related Github Issue
#160  and #164

### DIFF
--- a/infinite-tracing/src/main/java/com/newrelic/ResponseObserver.java
+++ b/infinite-tracing/src/main/java/com/newrelic/ResponseObserver.java
@@ -17,7 +17,7 @@ public class ResponseObserver implements StreamObserver<V1.RecordStatus> {
     private final AtomicBoolean shouldRecreateCall;
 
     public ResponseObserver(MetricAggregator metricAggregator, Logger logger, DisconnectionHandler disconnectionHandler,
-            AtomicBoolean shouldRecreateCall) {
+                            AtomicBoolean shouldRecreateCall) {
         this.metricAggregator = metricAggregator;
         this.logger = logger;
         this.disconnectionHandler = disconnectionHandler;
@@ -64,6 +64,7 @@ public class ResponseObserver implements StreamObserver<V1.RecordStatus> {
         logger.log(Level.FINE, "Completing gRPC call.");
         shouldRecreateCall.set(true);
         metricAggregator.incrementCounter("Supportability/InfiniteTracing/Response/Completed");
+        disconnectionHandler.resetConnectBackoffPolicy();
     }
 
     /**
--- a/infinite-tracing/src/main/java/com/newrelic/SpanEventConsumer.java
+++ b/infinite-tracing/src/main/java/com/newrelic/SpanEventConsumer.java
@@ -18,6 +18,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.logging.Level;
 
 /**
  * Accepts a {@link SpanEvent} for publishing to the Trace Observer.
@@ -25,6 +26,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
  * Not thread-safe.
  */
 public class SpanEventConsumer implements Consumer<SpanEvent> {
+    private final Logger logger;
     private final BlockingQueue<SpanEvent> queue;
     private final MetricAggregator aggregator;
     private final ConnectionHeaders connectionHeaders;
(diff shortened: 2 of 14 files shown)


infinite-tracing/src/main/java/com/newrelic/SpanEventConsumer.java
```diff
@@ -45,7 +48,9 @@ private SpanEventConsumer(BlockingQueue<SpanEvent> queue, MetricAggregator aggre
     @Override
     public void accept(SpanEvent spanEvent) {
         aggregator.incrementCounter("Supportability/InfiniteTracing/Span/Seen");
-        queue.offer(spanEvent);
+            if(!queue.offer(spanEvent)) {
```
XiXiaPdx, line 51: Currently, we create a metric for `Seen`  vs `Sent` .  The difference between the two are the number of spans that are presumably dropped. 

I think it would be nice to have a log message to indicate if the queue is full

infinite-tracing/src/main/java/com/newrelic/ResponseObserver.java
```diff
@@ -43,6 +46,14 @@ public void onError(Throwable t) {
             return;
         }
 
+        if (firstFailedPrecondition.get() && isFailedPrecondition(t)) {
```
XiXiaPdx, line 49: This handles the first `Failed_Precondition`, the one that should trigger an immediate reconnect

infinite-tracing/src/main/java/com/newrelic/ResponseObserver.java
```diff
@@ -15,18 +15,21 @@
     private final Logger logger;
     private final DisconnectionHandler disconnectionHandler;
     private final AtomicBoolean shouldRecreateCall;
+    private final AtomicBoolean firstFailedPrecondition;
 
     public ResponseObserver(MetricAggregator metricAggregator, Logger logger, DisconnectionHandler disconnectionHandler,
-            AtomicBoolean shouldRecreateCall) {
+                            AtomicBoolean shouldRecreateCall, AtomicBoolean firstFailedPrecondition) {
```
XiXiaPdx, line 21: We treat the first failed condition a bit differently, immediate reconnect with no backoff.

infinite-tracing/src/main/java/com/newrelic/ResponseObserver.java
```diff
@@ -15,18 +15,21 @@
...
         this.logger = logger;
         this.disconnectionHandler = disconnectionHandler;
         this.shouldRecreateCall = shouldRecreateCall;
+        this.firstFailedPrecondition = firstFailedPrecondition;
     }
 
     @Override
     public void onNext(V1.RecordStatus value) {
         metricAggregator.incrementCounter("Supportability/InfiniteTracing/Response");
+        firstFailedPrecondition.compareAndSet(false,true);
```
XiXiaPdx, line 32: According to the agent spec, receiving gRPC `Ok` codes are an acceptable indicator to reset  failed precondition  so the next time the error occurs, it is considered the first one again.

infinite-tracing/src/main/java/com/newrelic/StreamObserverFactory.java
```diff
@@ -23,6 +28,7 @@ public StreamObserverFactory(
         ClientCallStreamObserver<V1.Span> streamObserver = (ClientCallStreamObserver<V1.Span>) ingestServiceFutureStub.recordSpan(responseObserver);
 
         aggregator.incrementCounter("Supportability/InfiniteTracing/Connect");
+        logger.log(Level.FINE, "Connected to Trace Observer.");
```
XiXiaPdx, line 31: We should log when we are connected for better troubleshooting experience.

XiXiaPdx:
Disconnecting and reconnecting after receiving the error

https://user-images.githubusercontent.com/24482401/102809016-23c13a80-4376-11eb-9b0f-5ded6ba17abf.mov

Logs showing the disconnect and connect behavior. 

Screen Shot 2020-12-20 at 2 52 55 PM

PR #174, opened 2020-12-21
Weave WorkerTask and link onSubscribe to support subscribeOn with cus…

tspring:
…tom schedulers

There was a gap in our Webflux instrumentation where we weren't linking on custom schedulers. This should fix it so supported external calls like this get linked to the Transaction:
```
Mono.fromCallable(() ->
					myExternalCall()
					.subscribeOn(jdbScheduler);
```


jack-berg:
I've built the jar locally with this and validated that database calls _are not_ linked without this code, and _are_ linked with this code.

jack-berg (approved):

PR #175, opened 2020-12-31
Refactor infinite tracing

jack-berg:
- Rewrite ConnectionStatus to remove pesky busy waiting. 
- Refactor infinite tracing error handling to simplify and improve clarity.


infinite-tracing/src/main/java/com/newrelic/ChannelSupplier.java
```diff
@@ -12,47 +12,42 @@
...
     }
 
     @Override
     public ManagedChannel get() {
-        ConnectionStatus.BlockResult blockResult = getBlockResult();
+        boolean requiresReconnect = awaitRequiresReconnect();
 
-        if (blockResult == ConnectionStatus.BlockResult.GO_AWAY_FOREVER) {
-            throw new RuntimeException("No longer attempting to connect.");
-        }
```
jack-berg, line 33: Generating this exception is now handled within `ConnectionSynchronizer`. This enables the response to be simplified from an 4 value enum to a boolean, which makes it easier to reason about what a caller should do.

infinite-tracing/src/main/java/com/newrelic/ConnectionHeaders.java
```diff
@@ -24,13 +19,11 @@ public void set(String newRunToken, Map<String, String> headers) {
         newHeaders.put("agent_run_token", newRunToken);
         newHeaders.put("license_key", licenseKey);
         this.headers = newHeaders;
-
-        logger.log(Level.INFO, "New Relic connection successful. Attempting connection to the Trace Observer.");
-        connectionStatus.reattemptConnection();
```
jack-berg, line 29: This was a leaky abstraction. ConnectionHeaders should be a simple data carrier. Its methods should not be concerned with connection management.

infinite-tracing/src/main/java/com/newrelic/ConnectionSynchronizer.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Provides utilities for communicating connection status information across the various threads responsible for
+ * connecting and responding to connection errors.
+ *
+ * This class is thread safe.
+ */
+class ConnectionSynchronizer {
```
jack-berg, line 13: This class is a rewrite of `ConnectionStatus` and represents the bulk of the change of this PR. 

The contract of the infinite trace gRPC server is a bidirection stream, with the client sending spans, and the server sending response status messages. Sending spans happens in a single thread initialized in `SpanEventConsumer`. This thread also ends up managing the connection to the gRPC server. The responses sent by the server are handled by threads of cached thread pool, which is the default behavior of the `ManagedChannel` produced by `ChannelFactory#createChannel`.

This class's role is to synchronize connection status between these threads. There are two main improvements in this implementation of its `ConnectionStatus` predecessor:
1. The `awaitRequiresReconnect` method, equivalent to `ConnectionStatus#blockOnConnection`, uses a latch to block instead of busy waiting.
2. The API is simplified, with the notable removal of enum responses for callers to handle.

infinite-tracing/src/main/java/com/newrelic/ResponseObserver.java
```diff
@@ -96,4 +101,42 @@ private boolean isOkHttpALPNException(RuntimeException cause) {
         // It's the only mechanism we have to identify the problem.
         return cause.getMessage() != null && cause.getMessage().startsWith("TLS ALPN negotiation failed with protocols");
     }
+
+    /**
+     * Handle the error {@code status}. If irrecoverable, close the connection. Else, backoff and initiate reconnection.
+     *
+     * @param status the error status
+     */
+    void handleErrorStatus(Status status) {
```
jack-berg, line 110: The previous setup with a `BackoffPolicy` interface, implemented by `ConnectBackoffPolicy` and `DefaultBackoffPolicy`, and managed by `DisconnectionHandler` had poor "power to weight ratio". It resulted in the relatively simple workflow of handling status errors being distributed across 5 classes, and hard to put together. 

The `ResponseObserver#onError()` already does most of the error management, so why not include this last little bit so that its all in one spot?

PR #180, opened 2021-01-05
refactor infinite tracing for improved clarity and simplicity, add su…

### LINKED ISSUE #160: Respond to `Failed_Precondition` response code by disconnecting and connecting to Trace Observer

### LINKED ISSUE #164: Improve backoff strategy for Infinite Tracing

jack-berg:
Refactor infinite tracing for improved clarity and simplified. Refactor implies that the code has been reorganized but is functionally equivalent. This is _mostly_ true for this PR. The changes in behavior include:

- Backoff sequence added when `FAILED_PRECONDITIONS` status is returned.
- When a timeout error occurs due to 15 seconds of inactivity, reconnect immediately instead of backing off. This should have been the behavior but I think it slipped through the cracks when the change was made to detect timeout errors.
- Log messages have been adjusted slightly and standardized. 
- Added support for `_flakyCode` config, which allows for debugging error conditions with `_flakyPercentage`.
- Removed busy waiting while backoff is occurring in favor of a proper synchronizer mechanism. 

Closes #164.
Closes #160.


infinite-tracing/src/main/java/com/newrelic/SpanEventSender.java
```diff
@@ -0,0 +1,138 @@
...
+        try {
+            return queue.poll(250, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException("Thread interrupted while polling for spans.");
+        }
+    }
+
+    @VisibleForTesting
+    static V1.Span convert(SpanEvent spanEvent) {
```
XiXiaPdx, line 91: I would prefer the span conversion process to remain in a separate class. I know we discussed it is a straight forward and simple conversion. I think it is separate enough of a responsibility to organize on its own.

jack-berg: Sure I can do that, but before I do, let me just explain my reasoning for why I didn't do it initially. 

IMO, decomposition is about breaking large problems into smaller problems of singular concern, and a separate java class files isn't always the best way to accomplish that. They can reduce readability (i.e. having to look at 5 files to figure out what's happening is harder to undertstand than 5 functions in a single file). And they can encourage consumption outside of code outside its intended use (i.e. a private function in a class file sends a clear signal to maintainers that only the owning class should consume it while a separate class might more readily reused). They can help reduce clutter, but at 138 lines I don't think that this class is suffering from that. 

Anyway, that's my point of view in general, but in this case I don't think breaking it out is such a bad thing.

XiXiaPdx: Appreciate the insight and this will help me in the future! At times, I've gone through what seems a bit much of code indirection and agree with your point it is hard to read.

XiXiaPdx (approved):
I might be a bit too close to these changes as @jack-berg  and have discussed them at length.  I think these changes do improve the code quality, readability, and reduces complexity. 

If you plan to manual test for proper behavior, here are some notes that might be helpful.

1.  Use the [`flaky_code`](https://source.datanerd.us/agents/agent-specs/blob/master/Infinite-Tracing.md#metadata) agent config setting for testing. 

This header metadata is only observed by the staging Trace Observer

2. Testing `FAILED_PRECONDITION` will be just verifying we disconnect from the Trace Observer and reconnect to the same one.

infinite-tracing/src/main/java/com/newrelic/ResponseObserver.java
```diff
@@ -1,98 +1,117 @@
...
+        aggregator.incrementCounter("Supportability/InfiniteTracing/Span/gRPC/" + status.getCode());
+        aggregator.incrementCounter("Supportability/InfiniteTracing/Response/Error");
 
-        Status status = null;
-        if (t instanceof StatusRuntimeException) {
-            StatusRuntimeException statusRuntimeException = (StatusRuntimeException) t;
-            status = statusRuntimeException.getStatus();
-            metricAggregator.incrementCounter("Supportability/InfiniteTracing/Span/gRPC/" + status.getCode());
+        if (status.getCode() == Status.Code.UNIMPLEMENTED) {
+            // See: https://source.datanerd.us/agents/agent-specs/blob/master/Infinite-Tracing.md#unimplemented
```
jasonjkeller, line 55: ugg I wish these specs were public

newrelic-agent/src/main/java/com/newrelic/agent/config/InfiniteTracingConfigImpl.java
```diff
@@ -19,6 +19,7 @@
     public static final String TRACE_OBSERVER = "trace_observer";
     public static final String SPAN_EVENTS = "span_events";
     public static final String FLAKY_PERCENTAGE = "_flakyPercentage";
+    public static final String FLAKY_CODE = "_flakyCode";
```
jasonjkeller, line 22: Does the `_flakyCode` config need to be publicly documented for customer use?

jasonjkeller (approved):
I didn't have much exposure to this part of the codebase before the refactoring so I can't say that I have much of a feel for why it was refactored. That said, the refactored code looks good to me. Overall it looks to have reduced the number of classes a good amount without creating "god" classes, which is nice.

Has this been manually tested and verified to work (old and new functionality alike)? And have the AITs been run against this branch? If so, then let's go ahead and merge this.

newrelic-agent/src/main/java/com/newrelic/agent/config/InfiniteTracingConfigImpl.java
jack-berg: Good question! No, the headers included by `_flakyCode` and `_flakyPercentage` only impact interactions with the staging environment.

infinite-tracing/src/main/java/com/newrelic/ResponseObserver.java
jack-berg: > Has this been manually tested and verified to work (old and new functionality alike)? And have the AITs been run against this branch? 

Yes - lots of manual testing, and the AITs have passed.

PR #183, opened 2021-01-07
add configurable rate limiter for jar collector processing

### LINKED ISSUE #179: Rate limit jar collection processing
Jar collection processing can use high amounts of CPU at application startup. We should adjust the code to rate limit this process and effectively draw out jar processing over a longer period of time.

A rudimentary version of this strategy is available in this branch: https://github.com/newrelic/newrelic-java-agent/compare/rate-limit-jar-collector

Ideally the allowable rate would have a reasonable default, but also be configurable.

jack-berg:
Closes #179.

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/service/module/JarCollectorServiceProcessor.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/service/module/JarCollectorServiceProcessor.java
@@ -7,11 +7,12 @@
 
 package com.newrelic.agent.service.module;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
+import com.google.common.util.concurrent.RateLimiter;
 import com.newrelic.Function;
 import com.newrelic.agent.bridge.ManifestUtils;
 import com.newrelic.agent.config.AgentConfig;
-import com.newrelic.api.agent.Config;
 import com.newrelic.api.agent.Logger;
 
 import java.io.File;
@@ -28,6 +29,8 @@ import java.util.jar.JarEntry;
 import java.util.jar.JarInputStream;
 import java.util.jar.Manifest;
 import java.util.logging.Level;
+
+import static com.newrelic.agent.config.JarCollectorConfigImpl.DEFAULT_JARS_PER_SECOND;
 
 /**
  * Attempts to open jars and obtain version information from manifests.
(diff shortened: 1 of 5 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/service/module/JarCollectorServiceProcessor.java
```diff
@@ -55,32 +58,29 @@
...
         this.logger = logger;
-        this.skipTempJars = config.getValue("jar_collector.skip_temp_jars", true);
+        this.skipTempJars = agentConfig.getJarCollectorConfig().skipTempJars();
         if (!skipTempJars) {
             logger.log(Level.FINEST, "temporary jars will be transmitted to the host");
         }
+        this.ignoreJars = new ArrayList<>(agentConfig.getIgnoreJars());
+        int jarsPerSecond = agentConfig.getJarCollectorConfig().getJarsPerSecond();
+        if (jarsPerSecond <= 0) {
+            logger.log(Level.INFO, "Jars per second must be greater than 0. Defaulting to {0}.", DEFAULT_JARS_PER_SECOND);
```
tspring, line 75: Should we disable if 0 and use the default if negative?

jack-berg: I could do that. I was thinking through that question myself and came to the conclusion that you could effectively disable by setting the config to really high number. Like, if someone set it to 1_000_000_000, there's no way the single thread processing these jars could run into that bound.

jasonjkeller: I'm ok with leaving this behavior as is and providing public documentation for the `jar_collector` config so that customers can disable it if desired.

jasonjkeller (approved):
This looks great to me. Let's ship it.

Along with this PR I think we should update the [public agent config doc](https://docs.newrelic.com/docs/agents/java-agent/configuration/java-agent-configuration-config-file#Error_Collector) to include a section on the `jar_collector` config options:

```
  jar_collector:
    enabled: true # default true
    max_class_loaders: 300 # default 5000
    skip_temp_jars: true # default true
    jars_per_second: 5 # default 10
```

It doesn't look like we publicly document any of these config settings.

tspring (approved):

PR #187, opened 2021-01-12
add Transaction insertDistributedTraceHeaders and acceptDistributedTr…

jack-berg:
Working branch to add w3c trace header APIs.

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/OutboundHeadersMap.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/OutboundHeadersMap.java
@@ -8,17 +8,22 @@
 package com.newrelic.agent;
 
 import java.util.HashMap;
+import java.util.Map;
 
 import com.newrelic.api.agent.HeaderType;
 import com.newrelic.api.agent.OutboundHeaders;
 
-@SuppressWarnings("serial")
-public class OutboundHeadersMap extends HashMap<String, String> implements OutboundHeaders {
+/**
+ * An {@link OutboundHeaders} that is backed by a map.
+ */
+public class OutboundHeadersMap implements OutboundHeaders {
+
     private final HeaderType type;
+    private final Map<String, String> map;
 
     public OutboundHeadersMap(HeaderType type) {
-        super();
         this.type = type;
+        this.map = new HashMap<>();
     }
 
     @Override
@@ -26,8 +31,19 @@ public class OutboundHeadersMap extends HashMap<String, String> implements Outbo
         return type;
     }
 
+
     @Override
     public void setHeader(String name, String value) {
-        put(name, value);
+        map.put(name, value);
     }
+
+    /**
+     * Obtain a copy of the headers backing map.
+     *
+     * @return the headers map
+     */
+    public Map<String, String> asMap() {
+        return new HashMap<>(map);
+    }
+
 }
--- a/newrelic-api/src/main/java/com/newrelic/api/agent/NoOpAgent.java
+++ b/newrelic-api/src/main/java/com/newrelic/api/agent/NoOpAgent.java
@@ -168,6 +168,14 @@ class NoOpAgent implements Agent {
         @Override
         public void acceptDistributedTracePayload(DistributedTracePayload payload) {
         }
+
+        @Override
+        public void insertDistributedTraceHeaders(Map<String, String> headersMap) {
+        }
+
+        @Override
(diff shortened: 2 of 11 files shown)


newrelic-api/src/main/java/com/newrelic/api/agent/NoOpAgent.java
```diff
@@ -168,6 +168,14 @@ public void acceptDistributedTracePayload(String payload) {
         @Override
         public void acceptDistributedTracePayload(DistributedTracePayload payload) {
         }
+
+        @Override
+        public void insertDistributedTraceHeaders(Map<String, String> headersMap) {
+        }
+
+        @Override
+        public void acceptDistributedTraceHeaders(HeaderType headerType, Map<String, String> headersMap) {
```
jack-berg, line 177: We've discussed the accepting `TransportType` here, which is available in the agent bridge and would have to be promoted to be in the agent api module. I found the `HeaderType` enum already in the agent api module, and with a partially overlapping set of values as `TransportType`, and thought it might be more appropriate. 

Thoughts?

XiXiaPdx: I wonder why the TransportType has not been promoted already?  It offers more options to be specific and that would seem to be a good thing.  

I guess we could always add more constants to the HeaderType enum but that's potentially further repeating what's already on the TransportType.

jack-berg: Another thing worth nothing is that `HeaderType` is already part of the contract of `InboundHeaders` and `OutboundHeaders`, which are used quite a bit throughout the API an the implementation.

newrelic-agent/src/main/java/com/newrelic/agent/OutboundHeadersMap.java
```diff
@@ -8,26 +8,42 @@
...
 
 import com.newrelic.api.agent.HeaderType;
 import com.newrelic.api.agent.OutboundHeaders;
 
-@SuppressWarnings("serial")
-public class OutboundHeadersMap extends HashMap<String, String> implements OutboundHeaders {
+/**
+ * An {@link OutboundHeaders} that is backed by a map.
+ */
+public class OutboundHeadersMap implements OutboundHeaders {
```
XiXiaPdx, line 19: @jack-berg  Did you remove the extends HashMap and add returning a copy of the internal map because it's more idiomatic or was this necessary for the API to work?    I'm trying to piece together why `CrossProcessTransactionStateImpl` needed to change and this is the only clue I see so far

jack-berg: Yup we should favor composition over inheritance. Just a bit of improvement while I was in the neighborhood.

XiXiaPdx (approved):
LGTM!  Once these tests pass, can we run the AITs on this branch?

newrelic-api/src/main/java/com/newrelic/api/agent/NoOpAgent.java
jasonjkeller: In this case would it make sense to enhance `HeaderType` (so that it has parity with `TransportType`) and just replace usages of `TransportType` with it?

jasonjkeller: hmmm actually, I'm not sure if using `HeaderType` is the right choice as it deviates from the spec: https://source.datanerd.us/agents/agent-specs/blob/master/distributed_tracing/Distributed-Tracing.md#7-parent-transport-type

At a minimum `HeaderType` needs to have parity with all of the types declared in `TransportType`

Also this kind of convolutes matters given that all of our [metrics](https://source.datanerd.us/agents/agent-specs/blob/master/distributed_tracing/Distributed-Tracing.md#metrics) and [attributes](https://source.datanerd.us/agents/agent-specs/blob/master/distributed_tracing/Distributed-Tracing.md#attributes) explicitly refer to `transportType`.

I think promoting `TransportType` to the public API is probably the correct path forward.

jack-berg: Yeah, `TransportType` seems like the right fit, but we may have to do some other not-so-minor refactors to accommodate it, possibly making breaking changes to the `InboundHeaders` and `OutboundHeaders` interfaces which are part of the public API.

jack-berg:
Closing in favor of this alternative: https://github.com/newrelic/newrelic-java-agent/pull/188

Issue #125, opened 2020-11-23
Spring-Webflux, webclient doesn't capture Mono::switchIfEmpty external calls


tspring:

jasonjkeller:
The New Relic Java agent (as of 6.2.1) currently isn't instrumenting some parts of the Spring WebFlux reactive code.

Namely, the Spring `WebClient` instrumentation doesn't capture:
* `Mono::switchIfEmpty` external calls
* `Flux::flatMap` external calls

It looks like the latter (`Flux::flatMap`) is a direct result of not instrumenting `switchIfEmpty` which is eventually called in the [DefaultWebClient](https://github.com/spring-projects/spring-framework/blob/master/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClient.java#L551-L563):

```java
		private <T> Mono<T> handleBodyMono(ClientResponse response, Mono<T> body) {
			body = body.onErrorResume(WebClientUtils.WRAP_EXCEPTION_PREDICATE, exceptionWrappingFunction(response));
			Mono<T> result = applyStatusHandlers(response);
			return (result != null ? result.switchIfEmpty(body) : body);
		}

		private <T> Publisher<T> handleBodyFlux(ClientResponse response, Flux<T> body) {
			body = body.onErrorResume(WebClientUtils.WRAP_EXCEPTION_PREDICATE, exceptionWrappingFunction(response));
			Mono<T> result = applyStatusHandlers(response);
			return (result != null ? result.flux().switchIfEmpty(body) : body);
		}
```

Here's an app that repros the issue: https://github.com/jasonjkeller/spring-webflux-repro

tspring:
Closing as this is resolved in the current main branch

PR #190, opened 2021-01-15
Spring webflux subscribe

tspring:
When calling subscribe on a Mono/Flux, a LambdaSubscriber is created with a new Context. This change inserts a newrelic-token on said Context for the TokenLinkingSubscriber. 

I had to split the netty-reactor-0.8.0 instrumentation and create one for 0.9.0, because the earlier versions do not provide an impl of currentContext and we can't indirect weave inherited methods.

I also removed the writeWith and writeAndFlushWith weave methods from spring-webflux, as they were expiring the Token too early and allowing the Transaction to finish before the task was fully complete.


instrumentation/netty-reactor-0.9.0/src/main/java/reactor/netty/http/server/HttpTrafficHandler_Instrumentation.java
```diff
@@ -0,0 +1,27 @@
+/*
+ *
+ *  * Copyright 2020 New Relic Corporation. All rights reserved.
+ *  * SPDX-License-Identifier: Apache-2.0
+ *
+ */
```
jasonjkeller, line 6: This header needs to be added to all source files.

jasonjkeller (approved):
Based on previous discussions and positive feedback on the changes I'm ok with these changes. 

Out of curiosity, have any kind of performance tests been run to determine the impact of the additional token creation?

Also, before closing this out it would be great to determine whether or not it also addresses this issue: https://github.com/newrelic/newrelic-java-agent/issues/125

instrumentation/netty-reactor-0.8.0/src/main/java/reactor/netty/http/server/HttpServerMetricsHandler.java
```diff
@@ -0,0 +1,7 @@
+package reactor.netty.http.server;
+
+import com.newrelic.api.agent.weaver.SkipIfPresent;
+
+@SkipIfPresent
```
jack-berg, line 5: What's going on here?

instrumentation/netty-reactor-0.9.0/src/main/java/reactor/core/publisher/LambdaMonoSubscriber_Instrumentation.java
```diff
@@ -0,0 +1,45 @@
...
+    private Context nrContext;
+    final Context initialContext = Weaver.callOriginal();
+
+    @WeaveAllConstructors
+    protected LambdaMonoSubscriber_Instrumentation() {
+        // LamdaMonoSubscriber creates a new Context, so we create a new token and put it on the Context
+        // to be linked by TokenLinkingSubscriber but expired on onComplete here
+        if (AgentBridge.getAgent().getTransaction(false) != null
+                && initialContext.getOrDefault("newrelic-token", null) == null) {
+            nrContext = Context.of("newrelic-token", NewRelic.getAgent().getTransaction().getToken());
```
jack-berg, line 24: I believe we discussed naming this `newrelic-token` field to something else to disambiguate it from the key of the other context.

instrumentation/netty-reactor-0.9.0/src/main/java/reactor/netty/http/server/HttpTrafficHandler_Instrumentation.java
jack-berg: Does it? Some of this code was based off open telemetry instrumentation (as called out in comments and the `NOTICE.txt` in the root of these modules.  Was the omission of these headers in the `0.8.0` module accidental or intentional?

instrumentation/spring-webflux-5.1.0/src/main/java/org/springframework/http/ReactiveHttpOutputMessage_Instrumentation.java
```diff
@@ -23,32 +23,6 @@
...
-            if (this.token != null) {
-                this.token.expire();
-                this.token = null;
-            }
-        } catch (Throwable t) {
-            AgentBridge.instrumentation.noticeInstrumentationError(t, Weaver.getImplementationTitle());
-        }
-        return Weaver.callOriginal();
-    }
-
```
jack-berg, line 51: This is the expiring of tokens you said was being done defensively, but which not actually needed, correct?

instrumentation/netty-reactor-0.8.0/src/main/java/reactor/netty/http/server/HttpServerMetricsHandler.java
tspring: In reactor-netty versions prior to 0.9.0, LambdaSubscriber::currentContext is implemented in the super class, which we can't weave so we needed to split the instrumentation. This is to prevent netty-reactor-0.8 from applying to 0.9+ and effectively double weaving.

instrumentation/netty-reactor-0.9.0/src/main/java/reactor/core/publisher/LambdaMonoSubscriber_Instrumentation.java
tspring: Doing so made the code considerably more complicated and brittle. This way we continue to use the TokenLinkingSubscriber rather a new Subscriber.

instrumentation/spring-webflux-5.1.0/src/main/java/org/springframework/http/ReactiveHttpOutputMessage_Instrumentation.java
tspring: Yes, it turns out that the writeWith and writeAndFlushWith can get called during the reactive flow which prevents further operations from linking the token and effectively ending the Transaction early.

instrumentation/netty-reactor-0.9.0/src/main/java/reactor/netty/http/server/HttpTrafficHandler_Instrumentation.java
tspring: It was a bit of both, I'll add the headers to the classes that aren't derived from Otel

Issue #128, opened 2020-11-27
Support for java.net.http.HttpClient


mpostupalski:
## Is your feature request related to a problem? Please describe.
`java.net.http.HttpClient` is a standard Java http client so it would be nice to have it instrumented automatically

## Feature Description
`java.net.http.HttpClient` is instrumented automatically along with other 3rd party libraries

## Describe Alternatives
So far we are using custom instrumentation which is confusing. We can also switch to some 3rd party http client but we want to keep our client lightweight.

## Additional context
_none_

## Priority
Please help us better understand this feature request by choosing a priority from the following options: 
Really Want

Aha! Link: https://newrelic.aha.io/features/JAVA-192

tspring:
Hi @mpostupalski , Thanks for creating this. Instrumentation for java.net.http.HttpClient has been on our roadmap for some time but I'm hopeful we'll get to it sometime in the next few weeks.

tbradellis:
I've started work on this one.

PR #191, opened 2021-01-19
Jdk11 httpclient

XiXiaPdx:
This adds instrumentation for the Java 11 HttpClient, for #128 

Non-instrumentation changes:

- for `getRuntimeMaxSupportedClassVersion` in WeaveUtils,  increased the max supported version to Java 11.
- added `addReadUnnamedModuleToHttpModule` so java.net.http module can call into the unnamed module that holds the agent instrumentation utility classes.  This required updating the module-utils dependency with this PR - https://github.com/newrelic/newrelic-module-util-java/pull/13
- Update Github actions to include JDK11 

For the instrumentation:

- A user of the HttpClient would build a HttpRequest. There is where we instrument and add headers. 
- The request would then be sent with the HttpClient. Whether by `send` or `sendAsync`, under the hood, both call into `sendAsync`  on the HttpClientImpl.  This is where the agent reports as external, using a segment
- The HttpClient throws `UnresolvedAddressException`instead of `UnknownHostException` 

Works from manual testing 

client.sendAsync
Screen Shot 2021-01-19 at 9 42 58 AM

client.send
Screen Shot 2021-01-19 at 9 43 20 AM

The test app executes 1:1  of each send type from above.
Screen Shot 2021-01-19 at 9 46 12 AM

1 external for each transaction above
Screen Shot 2021-01-19 at 9 45 44 AM


XiXiaPdx:
The actions failure is because I changed the agent's build.gradle to test with a local jar of module-utils.  Once the new verson of module utils is released, I'll update this PR

XiXiaPdx:
@tspring , I wanted to write some instrumentation tests but instrumentation was not applying to the test class. After much digging, I see the reason is that the custom InstrumentationClassLoader does not load classes from the jdk, so the weave classes are not applied.   

In this situation, what do you suggest for test coverage?

XiXiaPdx:
re-opened in  #251

PR #206, opened 2021-02-02
fix infinite tracing race condition regarding span observer

jack-berg:
Fix a race condition in infinite tracing involving `SpanEventSender` and `ResponseObserver`:
- SpanEventSender obtains an observer via `ChannelManager#getSpanObserver()`
- SpanEventSender does some work, and eventually writes a Span to the observer
- However, the observer has already been cancelled by  ResponseObserver#onCompleted, which cancels the observer

The fix is to have ResponseObserver not directly cancel the observer, but instead mark it as needing recreation, where its recreated the next time SpanEventSender gets an observer from the channel manager. In doing so, SpanEventSender shouldn't be able to write to a cancelled observer, since any cancelling and recreation of observers happens in a blocking fashion during its call to ChannelManager#getSpanObserver(), and never in a separate thread.


jasonjkeller (approved):
Good catch!

PR #221, opened 2021-02-18
Dynamic attach support for the Java agent. Removal of the agent jar utility for dynamically attaching the agent.

jasonjkeller:
Dynamic attach support for the Java agent. However, this PR removes the agent jar utility for dynamically attaching the agent. In order to dynamically attach the agent you must currently use the [nri-introspector-java](https://github.com/newrelic/nri-introspector-java) tool.

For a clearer idea of what was removed from the original dynamic attach prototype see [this comparison](https://github.com/newrelic/newrelic-java-agent/compare/dynamic-attach...dynamic-attach-no-jar-util). Removing the agent jar utility for dynamically attaching the agent reduced the number of files changes from 60+ to less than 30.

Resolves #209

PR #228, opened 2021-02-25
Use connection listener for connection status

tspring:


XiXiaPdx (changes requested):
I see where you are going with this and now I see the pattern in the agent for using the connection listener. 

I tested this branch with the introspector. Although the agent does attach, there are JSON output issues. The CLI invoking the introspector is dependent on the JSON output.

Right now, testing with NEW_RELIC_SYNC_STARTUP=true in my env, the output is:

```
[{"agentArgs":"eJx1kM1qwzAQhN9FZzc4OP69ObFoRVPV2AohJ2HL6yBQpCCp6SH43atLoQn0sAu7+83Azh1N0glzA4uqeVAOIgT6Jq3RF9AeVXdE8ZF3eE92vCE968j2wHDDWVfvCH3lmNbbPW5Qhbz9AhT9wd8+exb2zg9nqc8vwigFwhu70vBtQUmxEubyoKjbltP6AweVcpJ7cP7pTp5c//cKhWmP+Ts+BXzOiyyBTBRCpGMB600ichizWcSbqRjjNMnHKYZhfrDoTzQ0Vnfs0P4+uERITmFYl1lZBtqBDdm1xoasijQuI3S15grWS3AhvWX5ATFBbxw=",
"success":false,"messages":[{"level":"INFO","label":"Msg","message":"Initializing agent"}],"pid":19699,"command":null}]
```
That was unexpected. Even before the busy wait code was added, the attach agent would indictate success in this test scenario.  This would need to be fixed since the agent logs indicate success but the JSON output does not. 

I tested with NEW_RELIC_SYNC_STARTUP=false in my env, the output is the same as above, which is another problem. 

It's the same issue Jason raised  here - https://github.com/newrelic/newrelic-java-agent/pull/221#discussion_r578079696

and is the reason the busy wait code was added.

Issue #229, opened 2021-02-25
JBoss Wildfly throws error on startup


Steve24by7:
# Wildfly startup shows the following error when NewRelic is enabled
2016-11-29 10:50:04,223 ERROR [org.jboss.as.controller.management-operation] (New Relic Harvest Service) WFLYCTL0013: Operation ("read-attribute") failed - address: ([("subsystem" => "transactions")]) - failure description: "WFLYCTL0216: Management resource '[(\"subsystem\" => \"transactions\")]' not found"

Tracked in New Relic Ticket 437213

## Description
#  https://discuss.newrelic.com/t/wildfly-startup-throws-an-error-for-transactions-subsystem/44417 

## Expected Behavior
Startup with no console errors.

## Troubleshooting
In the previous cases, as an another workaround to ignore the message, we suggested to disable jboss-jmx-7 instrumentation as tracking through the source code, it does look like the jboss is part of the below instrumentation package:

com.newrelic.instrumentation.jboss-jmx-7

This does work, but has the unfortunate side-effect of losing jmx monitoring.

## Additional context

(Migrate to Jira)

jasonchester:
The startup errors are causing the health endpoint to report as down in wildfly for us making this a blocking issue which we cannot ignore as we are using the health endpoint for our liveness checks.


```sh
❯ curl http://localhost:9990/health -sS | jq
```

```json
{
  "status": "DOWN",
  "checks": [
    {
      "name": "server-state",
      "status": "UP",
      "data": {
        "value": "running"
      }
    },
    {
      "name": "deployments-status",
      "status": "UP",
      "data": {
        "instant-impact-keycloak-bundle-1.0.0-SNAPSHOT.ear": "OK",
        "saml-group-idp-mapper-1.0.0-SNAPSHOT.jar": "OK",
        "client-management-keycloak-bundle-1.0.0-SNAPSHOT.ear": "OK",
        "keycloak-server.war": "OK",
        "set-existing-user-info-authenticator-1.0.0-SNAPSHOT.jar": "OK"
      }
    },
    {
      "name": "boot-errors",
      "status": "DOWN",
      "data": {
[...]

PR #230, opened 2021-02-26
JBoss Wildfly throws error on startup

Steve24by7:
### Overview
By allowing a specific delay to be set in the yml file, using the new parameter jboss7_jmxService_delay an int that is a child of common, the number of seconds to delay monitoring of the jboss jmx service, the admin can ensure that no error is thrown in the JBoss console during startup. This change has been implemented successfully at a New Relic customer's enterprise JBoss 7 implementation.

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/229

### Testing
Set log_level to finer to introduce two new log entries, for example:
2021-02-25T16:03:01,020-0500 [17938 1] com.newrelic FINE: JBoss7 invoking JMX service with delay 60 sec
2021-02-25T16:04:01,011-0500 [17938 36] com.newrelic FINE: JBoss7 JMX monitoring service has been installed

### Checks

### DIFF
--- a/instrumentation/jboss-jmx-7/src/main/java/org/jboss/modules/ModuleLoader_Instrumentation.java
+++ b/instrumentation/jboss-jmx-7/src/main/java/org/jboss/modules/ModuleLoader_Instrumentation.java
@@ -7,16 +7,39 @@
 
 package org.jboss.modules;
 
+import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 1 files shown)


jasonjkeller:
Closing this as we're concerned about starting threads in instrumentation and we'd prefer to find a solution that avoids that. Additionally any solution would also need to apply to the Wildfly instrumentation.

PR #234, opened 2021-03-04
Fix weaver jax rs issue

### LINKED ISSUE #233: Weaver is dependent on jax-rs instrumentation module classes being read from the module jar in a specific order
The weaver loads & caches each instrumentation module in [CachedWeavePackage.load](https://github.com/newrelic/newrelic-java-agent/blob/main/newrelic-weaver/src/main/java/com/newrelic/weave/weavepackage/CachedWeavePackage.java#L162-L189) by locating the instrumentation module jar and reading in the bytes for the all of the classes contained in the module.

It appears that the weaver depends on the classes being read in from the `JarInputStream` from the jar (`newrelic.jar!/instrumentation/jax-rs-1.0-1.0.jar`) in a specific and predictable order for the `jax-rs-1.0` instrumentation.

Specifically, the `JavaxWsRsApi_Instrumentation` class needs to be read in before `JavaxWsRsApi_Subresource_Instrumentation` or the instrumentation will fail to apply with a `WeaveViolation` (see [CLASS_MISSING_REQUIRED_ANNOTATIONS](https://github.com/newrelic/newrelic-java-agent/blob/main/newrelic-weaver/src/main/resources/com/newrelic/weave/violation/CLASS_MISSING_REQUIRED_ANNOTATIONS.md)) in cases where jax-rs sub-resources are being used and do not include the `@Path` annotation.

The reason is that a map of all method annotations is updated in [WeavePackage.processWeaveBytes](https://github.com/newrelic/newrelic-java-agent/blob/main/newrelic-weaver/src/main/java/com/newrelic/weave/weavepackage/WeavePackage.java#L146) when each of the classes mentioned above are processed. In the map (`Map<String, ClassNode>`) annotations are mapped to a `weaveNode` (either `JavaxWsRsApi_Instrumentation` or `JavaxWsRsApi_Subresource_Instrumentation`). 

As it currently works `JavaxWsRsApi_Subresource_Instrumentation` needs to be processed last so that it replaces the `weaveNode` values in the map for the sub-resource instrumentation to work properly.

As it stands, this approach is brittle and subject to breaking because there is no guarantee that the classes will be added and read from the `jax-rs-1.0-1.0.jar` in a deterministic manner given that the build process can vary depending on the environment that it is run in.

jasonjkeller:
- Refactors `WeavePackage` to process annotation based instrumentation classes properly when multiple classes with the same annotations exist
- Adds tests

This fixes jax-rs based transactions that were previously being named incorrectly when using sub-resources:

Screen Shot 2021-03-03 at 3 21 41 PM

Resolves #233


jasonjkeller:
AIT run: https://javaagent-build.pdx.vm.datanerd.us/view/All/job/JavaAgent_PR_AIT/48/

newrelic-weaver/src/main/java/com/newrelic/weave/weavepackage/WeavePackage.java
```diff
@@ -142,8 +142,25 @@ public WeavePackage(WeavePackageConfig config, List<byte[]> weavePackageBytes) {
                     requiredClassAnnotationsLookup.put(weaveNode.name, weave.getRequiredClassAnnotations());
                 }
                 if (isMethodAnnotationMatch) {
```
jasonjkeller, line 144: Not sure if it is worth adding similar logic to the above condition when `isClassAnnotationMatch` is `true`? I was trying to be minimally invasive and currently there don't appear to be any instrumentation modules that have a similar situation with class annotations but technically the same issue could happen there.

newrelic-weaver/src/test/java/com/newrelic/weave/weavepackage/WeavePackageAnnotationProcessingTest.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+        // all method annotations should map to the JaxRsSubresourceAnnotations ClassNode, regardless of which JaxRs class bytes were processed first
+        for (String annotation : expectedJaxRsMethodAnnotationClasses) {
+            String weaveNode = allMethodAnnotationWeaves.get(annotation).name;
+            Assert.assertEquals(JAX_RS_SUBRESOURCE_ANNOTATIONS_CLASS_NODE_NAME, weaveNode);
+        }
+    }
+
+    @Test
+    public void testJaxRsAnnotationProcessingOppositeClassByteOrder() throws IOException {
```
XiXiaPdx, line 94: Could you explain how this test is different than the one above? I'm having trouble seeing the difference.

XiXiaPdx (approved):
:sushi:  Awesome work, looks great!

newrelic-weaver/src/test/java/com/newrelic/weave/weavepackage/WeavePackageAnnotationProcessingTest.java
jasonjkeller:
```
        List<byte[]> weaveBytes = new ArrayList<>();
        weaveBytes.add(WeaveTestUtils.getClassBytes(JAX_RS_ANNOTATIONS_CLASS_NAME));
        weaveBytes.add(WeaveTestUtils.getClassBytes(JAX_RS_SUBRESOURCE_ANNOTATIONS_CLASS_NAME));
```

The two `add` lines are swapped which reverses the order that the `weaveBytes` are read in. This effectively mimics generating an instrumentation module jar and gives us control over the order that the classes are read from it.

newrelic-weaver-api/src/main/java/com/newrelic/api/agent/weaver/WeavePriorityOrder.java
```diff
@@ -0,0 +1,31 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Sets the priority order that a weave instrumentation class should be processed in when processing all classes in a given {@code WeavePackage}.
+ *
+ * This can be used to force weave instrumentation classes to be processed in a particular order, if necessary.
+ *
```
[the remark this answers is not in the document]
jasonjkeller: haha it was MIN at one point but I changed it last night.

> What do you think about calling out explicitly in the comments here that WeavePriorityOrder is Integer.MAX_VALUE on instrumentation classes that don't have this annotation?

I don't think that this needs to be called out on the javadoc here as it's not behavior of the annotation, it's behavior of the `WeaveUtils` helper method that returns that value and it's already documented there. If a class doesn't have the annotation then they won't ever hit the logic that parses them.

newrelic-weaver/src/test/java/com/newrelic/weave/weavepackage/WeavePackageAnnotationProcessingTest.java
```diff
@@ -0,0 +1,145 @@
...
+        expectedSpringMethodAnnotationClasses.add("org.springframework.web.bind.annotation.RequestMapping");
+        expectedSpringMethodAnnotationClasses.add("org.springframework.web.bind.annotation.PatchMapping");
+        expectedSpringMethodAnnotationClasses.add("org.springframework.web.bind.annotation.PutMapping");
+        expectedSpringMethodAnnotationClasses.add("org.springframework.web.bind.annotation.GetMapping");
+        expectedSpringMethodAnnotationClasses.add("org.springframework.web.bind.annotation.PostMapping");
+        expectedSpringMethodAnnotationClasses.add("org.springframework.web.bind.annotation.DeleteMapping");
+    }
+
+    @Test
+    public void testJaxRsAnnotationProcessing() throws IOException {
```
XiXiaPdx, line 67: Suggestion for the test name. I think it'll make it easier for our future selves to see the difference in the test

`testJaxRsSubResourceAnnotationsProcessedFirst`

newrelic-weaver/src/test/java/com/newrelic/weave/weavepackage/WeavePackageAnnotationProcessingTest.java
XiXiaPdx: Thanks!  So same suggestion as above with the test name. 

`testJaxRsSubResourceAnnotationsProcessedLast`

newrelic-weaver/src/test/java/com/newrelic/weave/weavepackage/WeavePackageAnnotationProcessingTest.java
jasonjkeller: Done, for both tests

jasonjkeller:
AITs are green after latest changes: https://javaagent-build.pdx.vm.datanerd.us/view/All/job/JavaAgent_PR_AIT/51/

PR #237, opened 2021-03-07
Fix of transaction propagation for empty mono/flux and retry operators

veklov:
This PR fixes two issues with the project reactor instrumentation
- Transaction is not propagated to Mono.doOnSuccess and Flux.doOnComplete
- Transaction is lost after Mono.retry* re-subscribes while handling an error

### DIFF
--- a/instrumentation/netty-reactor-0.8.0/src/main/java/com/nr/instrumentation/reactor/netty/TokenLinkingSubscriber.java
+++ b/instrumentation/netty-reactor-0.8.0/src/main/java/com/nr/instrumentation/reactor/netty/TokenLinkingSubscriber.java
@@ -48,7 +48,7 @@ public class TokenLinkingSubscriber<T> implements CoreSubscriber<T> {
 
     @Override
     public void onComplete() {
-        subscriber.onComplete();
+        withNRToken(subscriber::onComplete);
     }
 
     @Override
@@ -67,8 +67,10 @@ public class TokenLinkingSubscriber<T> implements CoreSubscriber<T> {
     @Trace(async = true, excludeFromTransactionTrace = true)
     private void withNRError(Runnable runnable, Throwable throwable) {
         if (token != null && token.isActive()) {
-            token.linkAndExpire();
+            token.link();
             if (NettyReactorConfig.errorsEnabled) {
+                // I believe 100% of users disable this as it makes NewRelic to report caught and handled
+                // exceptions as errors in APM. Is there a real use case for this?
                 NewRelic.noticeError(throwable);
             }
         }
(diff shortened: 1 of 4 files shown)


instrumentation/netty-reactor-0.8.0/src/main/java/com/nr/instrumentation/reactor/netty/TokenLinkingSubscriber.java
```diff
@@ -67,8 +67,10 @@ private void withNRToken(Runnable runnable) {
     @Trace(async = true, excludeFromTransactionTrace = true)
     private void withNRError(Runnable runnable, Throwable throwable) {
         if (token != null && token.isActive()) {
-            token.linkAndExpire();
+            token.link();
             if (NettyReactorConfig.errorsEnabled) {
+                // I believe 100% of users disable this as it makes NewRelic to report caught and handled
+                // exceptions as errors in APM. Is there a real use case for this?
```
tspring, line 73: Would you be OK with setting the default to false and leaving the config in case someone finds it useful

instrumentation/netty-reactor-0.8.0/src/test/java/com/nr/agent/instrumentation/TransactionPropagationTest.java
```diff
@@ -0,0 +1,254 @@
...
+    public static final String A_VALUE = "";
+
+    @BeforeClass
+    public static void init() {
+        Hooks.onEachOperator(TokenLinkingSubscriber.class.getName(), tokenLift());
+        DebugFlag.tokenEnabled.set(true);
+    }
+
+    @Test
+    @Ignore("A test harness check")
```
tspring, line 45: Can you remove or un-ignore all of the ignored tests

tspring (changes requested):
Thanks @veklov . This all looks good. Just two minor changes if you don't mind

instrumentation/netty-reactor-0.8.0/src/main/java/com/nr/instrumentation/reactor/netty/TokenLinkingSubscriber.java
veklov: I just wanted to draw your attention to this. No change is necessary if you think it may be useful for someone.

Please also note that there is a similar lifter in spring-webflux instrumentation module. It reports such error unconditionally and expires tokens immediately after linking (one of the issues being fixed in this PR).
https://github.com/newrelic/newrelic-java-agent/blob/2ff0a81c4903854ea64f313422c9e520fa0cdf94/instrumentation/spring-webflux-5.3.0/src/main/java/com/nr/agent/instrumentation/spring/reactive/Util.java#L74-L92
I use spring webflux server and spring webclient with reactor-netty within the same application. Given that neither token expiration nor error reporting happens after I apply this PR and configure reactor-netty instrumentation to not report errors, it seems that the webflux's lifter is not used and creates a lot of confusion.

instrumentation/netty-reactor-0.8.0/src/test/java/com/nr/agent/instrumentation/TransactionPropagationTest.java
veklov: Done

instrumentation/netty-reactor-0.8.0/src/main/java/com/nr/instrumentation/reactor/netty/TokenLinkingSubscriber.java
tspring: Thanks for the pointer, I'll revisit that instrumentation and see if we can't clean it up some

tspring (approved):

Issue #244, opened 2021-03-15
 Remove nrcerts and the use_private_ssl config 


XiXiaPdx:
This will address #235 

There is prior art in the internal archived java agent, JAVA-6240.

PR #245, opened 2021-03-15
remove SSL Config and NR certs

XiXiaPdx:
For #244 

1. removes NR bundled certs
2. removes `use_private_ssl`
3. adds log at info level when `ca_bundle_path` is used
4. adds log specifically for CertificateException, calling out the header might be malformed.

Manually tested [against this internal endpoint](https://newrelic.slack.com/archives/C01R23NQH5W/p1615224255064300) with the following settings:
```
ca_bundle_path:  null
use_private_ssl: true
```

As expected, agent successfully negotiated SSL handshake and logged use_private_ssl has been removed. 

```
ca_bundle_path:  /path/to/current/newrelic-com.pem 
use_private_ssl: true
```

As expected, agent was unsuccessful in negotiating SSL handshake.  Also, `Using ca_bundle_path: ....` was logged as well.


XiXiaPdx:
AITS look good - https://javaagent-build.pdx.vm.datanerd.us/job/JavaAgent_PR_AIT/56/

newrelic-agent/src/main/java/com/newrelic/agent/transport/apache/ApacheSSLManager.java
```diff
@@ -125,6 +56,9 @@ private static KeyStore getKeyStore(String caBundlePath)
                 while (is.available() > 0) {
                     try {
                         caCerts.add((X509Certificate) cf.generateCertificate(is));
+                    } catch (CertificateException e) {
+                        Agent.LOG.log(Level.SEVERE, "Failed to generate ca_bundle_path certificate. Verify the certificate format. For instance, is it this issue JDK-8208602 ?", e);
+                        break;
                     } catch (Throwable t) {
                         Agent.LOG.log(Level.SEVERE,
                                 "Unable to generate ca_bundle_path certificate. Will not process further certs.", t);
```
jasonjkeller, line 64: Adding the new catch block above will prevent the code from getting into this catch block and logging `Unable to generate ca_bundle_path certificate. Will not process further certs` since the exception that was making it here was `CertificateException`.

```
2021-03-11T15:51:17,781-0800 [89136 1] com.newrelic ERROR: Unable to generate ca_bundle_path certificate. Will not process further certs.
java.security.cert.CertificateException: Could not parse certificate: java.io.IOException: Illegal header: -----BEGIN CERTIFICATE----- 
```

Why don't we get rid of the new catch and just merge the two log lines together in the old catch block:

```
} catch (Throwable t) {
                        Agent.LOG.log(Level.SEVERE,
                                "Unable to generate ca_bundle_path certificate. Please verify the certificate format. Will not process further certs.", t);
}
```

In general it's a terrible idea to `catch (Throwable t)` and catching a more specific exception (like you were doing with `CertificateException`) would be preferred but I don't know if `Throwable` is used here for a specific reason or not and I'd prefer to change as little behavior as possible around this logic.

jasonjkeller (changes requested):

newrelic-agent/src/main/java/com/newrelic/agent/config/AgentConfigImpl.java
```diff
@@ -356,7 +356,7 @@ private String initSSLConfig() {
             if (caBundlePath != null) {
                 Agent.LOG.log(Level.INFO, "use_private_ssl configuration setting has been removed.");
             } else {
-                Agent.LOG.log(Level.SEVERE, "use_private_ssl configuration setting has been removed. Please use ca_bundle_path instead.");
+                Agent.LOG.log(Level.SEVERE, "The use_private_ssl configuration setting has been removed and will be ignored. The agent will use the JVM/JRE truststore by default unless you configure ca_bundle_path to use a different truststore.\n");
```
jasonjkeller, line 359: Is the newline character (`\n`) needed here or is it just a copy & paste artifact? Let's remove it if it isn't intentional.

jasonjkeller (approved):
One minor nitpick about her newline character otherwise looks good to go.

PR #251, opened 2021-03-22
Java11 httpclient instrumentation

XiXiaPdx:
This adds instrumentation for the Java 11 HttpClient, for #128 

ISSUE:   The transactions take a long time to report to NR because there is an unexpired token in every. transaction. That token takes a default 180 seconds to timeout, finishing the transaction. 

https://staging.one.nr/01qwL4EArw5

The token is created in the process of CompletableFutures. The issue is being tracked here https://github.com/newrelic/newrelic-java-agent/issues/344

Non-instrumentation changes:

- for `getRuntimeMaxSupportedClassVersion` in WeaveUtils,  increased the max supported version to Java 11.
- added `addReadUnnamedModuleToHttpModule` so java.net.http module can call into the unnamed module that holds the agent instrumentation utility classes.  This required updating the module-utils dependency with this PR - https://github.com/newrelic/newrelic-module-util-java/pull/13
- Update Github actions to include JDK11 

For the instrumentation:

- A user of the HttpClient would build a HttpRequest. There is where we instrument and add headers. 
- The request would then be sent with the HttpClient. Whether by `send` or `sendAsync`, under the hood, both call into `sendAsync`  on the HttpClientImpl.  This is where the agent reports as external, using a segment
- The HttpClient throws `UnresolvedAddressException`instead of `UnknownHostException`


XiXiaPdx:
1. This bug fix PR #250  in the Java Agent needs to merged first. It affects the performance of this instrumentation and the integration test built for it.  
2. After the bug fix is merged, the integration test should be merged.
https://source.datanerd.us/java-agent/agent-integration-tests/pull/42
3. Then run the internal AITs against this PR branch to verify it passes.

welintoncoradini:
Early tests yielded positive results, NR trace.id is now being passed to the downstream entities. About a ~1/5 of the requests could be seen in the portal (10min delay).

XiXiaPdx:
This is 🟢  for the AITs - https://javaagent-build.pdx.vm.datanerd.us/job/JavaAgent_PR_AIT/113/

But this 🟢  does not include the  Java11 HttpClient specific AIT.  That AIT is on the "stop list" because the instrumentation here has not been merged yet.     

However, I've tested this branch against the specific Java11 HttpClient AIT  locally and it passes.

jasonjkeller (approved):

instrumentation/httpclient-jdk11/src/main/java/com.nr.agent.instrumentation.httpclient/InboundWrapper.java
```diff
@@ -0,0 +1,26 @@
...
+
+import com.newrelic.api.agent.HeaderType;
+import com.newrelic.api.agent.InboundHeaders;
+
+import java.net.http.HttpResponse;
+import java.util.function.Function;
+
+public class InboundWrapper implements InboundHeaders {
+
+    HttpResponse delegate;
```
twcrone, line 11: Why is this package private?

instrumentation/httpclient-jdk11/src/main/java/com.nr.agent.instrumentation.httpclient/Util.java
```diff
@@ -0,0 +1,99 @@
...
+
+import java.net.ConnectException;
+import java.net.URI;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.util.concurrent.CompletionException;
+import java.util.function.BiConsumer;
+import java.util.logging.Level;
+
+public class Util {
```
twcrone, line 15: Might not be a big deal, but we might want to qualify our "Util" classes with a semi-unique prefix so when searching in IntelliJ its easier to resolve.  Once again, maybe not a big deal for our work but in other projects I've been on it was.

instrumentation/httpclient-jdk11/src/main/java/jdk/internal/net/http/HttpClientImpl_Instrumentation.java
```diff
@@ -0,0 +1,51 @@
...
+
+@Weave(originalName = "jdk.internal.net.http.HttpClientImpl", type = MatchType.ExactClass)
+final class HttpClientImpl_Instrumentation {
+
+    @Trace
+    public <T> HttpResponse<T> send(HttpRequest req, HttpResponse.BodyHandler<T> responseHandler)
+            throws IOException, InterruptedException {
+        URI uri = req.uri();
+        Segment segment = startSegment(uri);
+        HttpResponse<T> response;
```
twcrone, line 26: This block would be one of those opportunities we discussed where a private (maybe static) method might help document what it does real quick.  'tryToProcessResponse' perhaps is quicker to understand than scanning the code.  Just a thought, not meant to be a criticism by ANY means.

newrelic-agent/src/main/java/com/newrelic/bootstrap/BootstrapAgent.java
```diff
@@ -208,10 +206,30 @@ private static ClassLoader getPlatformClassLoaderOrNull() {
...
         try {
             ModuleUtil util = new ModuleUtilImpl();
-            util.redefineModules(inst, agentClassLoader);
+            util.redefineJavaBaseModule(inst, agentClassLoader);
+        } catch (Throwable t) {
+            System.err.println("The agent failed to redefine modules as necessary. " + t);
+        }
+    }
+
+    /**
```
twcrone, line 218: Does Javadoc even get generated on private methods?  If not, you can probably boil this one down to a regular comment.

instrumentation/httpclient-jdk11/src/main/java/com.nr.agent.instrumentation.httpclient/InboundWrapper.java
XiXiaPdx: Good catch, will fix in update PR

instrumentation/httpclient-jdk11/src/main/java/com.nr.agent.instrumentation.httpclient/Util.java
XiXiaPdx: Sounds good. What do you suggest here?

newrelic-agent/src/main/java/com/newrelic/bootstrap/BootstrapAgent.java
XiXiaPdx: Good to know on the javadoc for private methods.  I’ve never thought about it so I did some research and it seems there are good opinions on this.  I'm going to keep it as is with the current comment style. There are other private method javadocs in that class and probably throughout the agent. Whether right or wrong, there appears to be precedent in the agent.  

Happy to change it if the team decides this format is not preferred!

instrumentation/httpclient-jdk11/src/main/java/com.nr.agent.instrumentation.httpclient/Util.java
twcrone: Once again, not suggesting you change now after the fact, but if you think its helpful in the future `SomethingSpecificUtil` might make it easier to find and distinguish.

instrumentation/httpclient-jdk11/src/main/java/jdk/internal/net/http/HttpClientImpl_Instrumentation.java
twcrone: Yeah, I think it would not work or be helpful here now that I look at more closely.  The `Weaver.callOriginal()` and segment stuff probably works better inline here rather than pulled out.  I'll try to give a better example at some point.

newrelic-agent/src/main/java/com/newrelic/bootstrap/BootstrapAgent.java
twcrone: I guess my only follow up point here is if our javadoc generator does not run against private methods currently, there is little point in using the verbose Javadoc formatting here.

```
/* 
  Modify the java.net.http module so that it can read from the platform classloader's
  unnamed module. The agent http client instrumentation utility classes are
  in this specific unnamed module.
  
  ModuleUtil is compiled in a multi-release jar. In Java < 11, this
  results in a no-op implementation
*/
```
is probably sufficient without extra @param etc. that probably just clutters things up.

vijaykramesh:
this seems to have broken scala on java 11 - see https://github.com/newrelic/newrelic-java-agent/issues/438

PR #263, opened 2021-04-05
add service_name attribute value

XiXiaPdx:
@jasonjkeller 

I think this is what you mean by consistency for service name, in the JFR service, as it is in the jfr daemon?

I could have used jfrconfig.  The daemonConfig is exactly how it is done in the daemon so I opted for that syntax.


jasonjkeller:
Closing. This is now handled elsewhere.

Issue #225, opened 2021-02-24
Update http Response code attribute names


jodeev:
<p style="text-align:start;">Implement the last piece of <a href="https://newrelic.atlassian.net/browse/AGENTS-201">AGENTS-201: Span Event Improvements</a>.</p><p style="text-align:left;">Add the following Span attributes on external spans, with the http response code for each external call made:</p><ul>
<li><p><code>http.statusCode</code></p></li>
<li><p><code>http.statusText</code></p></li>
</ul><p style="text-align:start;">These attributes should be added when using any supported http client. These attributes should also be applied to the root span.</p><p style="text-align:left;">The above attributes will be replacing the following <i>deprecated</i> attributes:</p><ul>
<li><p><code>http.responseCode</code></p></li>
<li><p><code>response.status</code></p></li>
<li><p><code>response.statusMessage</code></p></li>
</ul><p><br></p><p> <code>http.statusCode</code>  and <code>http.statusText</code>  attribute names should also be added to transactions, and old attribute names deprecated.</p><p style="text-align:start;"><br></p><p style="text-align:start;"><b>Engineering notes:</b></p><ul>
<li><p>Have to modify HttpParameters to accept a shim that might have status code.</p></li>
<li><p>Opportunistic: Redesign HttpParameters entirely; shim it in the API implementation so that the agent code only has to deal with one non-deprecated version.</p></li>
</ul><p><br></p><p><b>Acceptance criteria</b></p><ul>
<li><p>Conform to <a href="https://source.datanerd.us/agents/agent-specs/blob/master/Span-Events.md#http"><u><span style="color:#1155CC;">agent spec for attribute naming, specifically pay attention to http.statusCode and http.statusText, and make sure that this has been implemented on spans</span></u></a></p></li>
<li><p>Attribute name of <code>http.statusCode</code> and <code>http.statusText</code> on spans needs to match the attribute name on transactions</p></li>
</ul><p><br></p><p><b>Notes</b></p><p>Release notes examples: <a href="https://docs.newrelic.com/docs/release-notes/agent-release-notes/php-release-notes/php-agent-960255">https://docs.newrelic.com/docs/release-notes/agent-release-notes/php-release-notes/php-agent-960255</a></p>

Aha! Link: https://newrelic.aha.io/features/JAVA-260

jasonjkeller:
When a transaction ends it processes the transaction and notifies its listeners:

[...]

PR #272, opened 2021-04-09
Update http attrs

meiao:
### Overview
Adds http status code/text to external spans and web transaction spans.

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/225

### Checks

Not entirely.

Yes.
There is a change in the signature of one the HttpParameters constructors. The constructor is protected, so it would only affect people who subclassed HttpParameters. This can be avoided by creating a new constructor.
A method was added to the Build interface of the HttpParameters$Builder. This will only affect people that implemented that interface.

No.

### DIFF
--- a/instrumentation/httpclient-4.0/src/main/java/org/apache/http/client/HttpClient.java
+++ b/instrumentation/httpclient-4.0/src/main/java/org/apache/http/client/HttpClient.java
@@ -63,6 +63,7 @@ public abstract class HttpClient {
                 .uri(requestURI)
                 .procedure(PROCEDURE)
                 .inboundHeaders(inboundCatWrapper)
+                .status(response.getStatusLine().getStatusCode(), response.getStatusLine().getReasonPhrase())
                 .build());
     }
 
@@ -219,6 +220,7 @@ public abstract class HttpClient {
                     .uri(requestURI)
                     .procedure(PROCEDURE)
                     .inboundHeaders(inboundCatWrapper)
+                    .status(response.getStatusLine().getStatusCode(), response.getStatusLine().getReasonPhrase())
                     .build());
         }
     }
--- a/newrelic-api/src/main/java/com/newrelic/api/agent/HttpParameters.java
+++ b/newrelic-api/src/main/java/com/newrelic/api/agent/HttpParameters.java
@@ -31,6 +31,10 @@ public class HttpParameters implements ExternalParameters {
      */
     private final String procedure;
 
+    private final Integer statusCode;
+
+    private final String statusText;
+
     /**
      * The headers from the response of the external call.
      */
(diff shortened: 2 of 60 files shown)


instrumentation/httpclient-4.0/src/main/java/org/apache/http/client/HttpClient.java
```diff
@@ -63,6 +63,7 @@ private static void processResponse(URI requestURI, HttpResponse response) {
                 .uri(requestURI)
                 .procedure(PROCEDURE)
                 .inboundHeaders(inboundCatWrapper)
+                .status(response.getStatusLine().getStatusCode(), response.getStatusLine().getReasonPhrase())
```
meiao, line 66: This is what every instrumentation will have to add.
The step builder still works because status() was added to the Build interface which is the last step.

A previous design used a new interface that extended Build. But it would break if a client assigned the steps to variables. e.g.
```
UriParameter uriParam = HttpParameters.library("gRPC");
ProcedureParameter procParam = uriParam.uri(uri);
// supposing the new interface was returned by uri(), the line above would break.
```

newrelic-api/src/main/java/com/newrelic/api/agent/HttpParameters.java
```diff
@@ -42,13 +46,16 @@
     private final ExtendedInboundHeaders extendedInboundResponseHeaders;
 
     protected HttpParameters(String library, URI uri, String procedure, InboundHeaders inboundHeaders) {
-        this(library, uri, procedure, inboundHeaders, null);
+        this(library, uri, procedure, null, null, inboundHeaders, null);
     }
 
-    protected HttpParameters(String library, URI uri, String procedure, InboundHeaders inboundHeaders, ExtendedInboundHeaders extendedInboundHeaders) {
+    protected HttpParameters(String library, URI uri, String procedure, Integer statusCode, String statusText, InboundHeaders inboundHeaders,
+            ExtendedInboundHeaders extendedInboundHeaders) {
```
meiao, line 53: This constructor was changed and would break anyone subclassing HttpParameters and using it.
There is a simple workaround which is to create a constructor that receives a Builder.

newrelic-api/src/main/java/com/newrelic/api/agent/HttpParameters.java
```diff
@@ -212,6 +251,12 @@ public static UriParameter library(String library) {
 
     public interface Build {
 
+        /**
+         * Set the status code/text of the HTTP response.
+         * @return the builder in a buildable state.
+         */
+        Build status(Integer statusCode, String statusText);
```
meiao, line 258: This is another breaking change. But should only cause problems to someone implementing this interface.
Not sure why one would do it.

PR #276, opened 2021-04-13
Build with java8

XiXiaPdx:
`./gradlew build. -x test` and  `./gradlew jar` were successful locally. 

1. Any mentions to `JavaVersion.VERSION_1_7` in gradle files are now 
```
sourceCompatibility = JavaVersion.VERSION_1_8
targetCompatibility = JavaVersion.VERSION_1_8
```

2. Updated guava dependency to be the non-android flavor. The agent was using guava-android because it is Java 7 compatible. 
3.  There were 4 failing unit tests that required minor tweaks to resolve. 
4. Locally, the only instrumentation test failures are with the RabbitMQ instrumentation but I don't think they are of concern. The failures existed before making the Java 8 changes to the agent and I believe are related to my local network configuration.

5. The 9 failing AITS - https://javaagent-build.pdx.vm.datanerd.us/job/JavaAgent_PR_AIT/62/

It is the same error for all of the tests, timed out waiting for the server to start up. 

These AITS are resolved through this internal PR here - https://source.datanerd.us/java-agent/agent-integration-tests/pull/45


newrelic-agent/src/main/java/com/newrelic/agent/config/UtilizationDataConfig.java
```diff
@@ -35,7 +35,7 @@
 
     private String parseConfigVal(String key) {
         if (getProperty(key) != null) {
-            return String.valueOf(getProperty(key));
+            return getProperty(key).toString();
```
XiXiaPdx, line 38: On Java 8 Agent, String.valueOf was causing a ClassCastException with a cross agent unit test.   

```
java.lang.ClassCastException: 
class java.lang.Long cannot be cast to class 
[C (java.lang.Long and 
[C are in module java.base of loader 'bootstrap')
```

When debugging, the more readable reason is :

```
Type mismatch Can't assign non-array value to an array
```

On Java 7 Agent, the byte code for this method is

```
String.valueOf:(Ljava/lang/Object;)Ljava/lang/String;
```

On Java 8, the byte code is:

```
String.valueOf:([C)Ljava/lang/String;
```

I don't know why this compiles differently. I tried a scratch program with just a psvm and the same behavior.  

This took way too long to track down. In the interest of time, this was the quickest fix and I'm not sure if it is the best one.  

This method `parseConfigValue` is only called in this class so the blast radius appears small.

infinite-tracing/src/test/java/com/newrelic/ResponseObserverTest.java
```diff
@@ -98,7 +90,7 @@ void shutdownChannelAndBackoff_ConnectTimeoutBackoffZeroSeconds() {
 
         target.shutdownChannelAndBackoff(status);
 
-        verify(logger).log(eq(Level.FINE), any(Throwable.class), anyString(), any());
+        verify(logger).log(eq(Level.FINE), any(Throwable.class), anyString(), anyString());
```
XiXiaPdx, line 93: any() was failing.  anyString() was a better arguement matcher and passed the test.

newrelic-agent/src/main/java/com/newrelic/agent/TransactionService.java
```diff
@@ -71,12 +63,7 @@ public TransactionService(int numMaintenanceThreads, long initialDelay, long del
 
         // run as daemon to not prevent shutdown of application
         scheduler = Executors.newScheduledThreadPool(numMaintenanceThreads, new DefaultThreadFactory(TRANSACTION_SERVICE_PROCESSOR_THREAD_NAME, true));
-        scheduler.scheduleWithFixedDelay(new Runnable() {
-            @Override
-            public void run() {
-                processQueue();
-            }
-        }, initialDelayMilli, delayMilli, TimeUnit.MILLISECONDS);
+        scheduler.scheduleWithFixedDelay(() -> TransactionService.this.processQueue(), initialDelayMilli, delayMilli, TimeUnit.MILLISECONDS);
```
jasonjkeller, line 66: OMG is this a lambda in the Java agent code base!!! It's happening!!!

XiXiaPdx: Yes, that was my little test.  That caused a bug that took a while to resolve. The error suggested the android guava dependency might be the issue.   I ended up deleting my gradle cache and starting from scratch again.

jasonjkeller:
Local testing all test suites passed:

Functional tests
```
./gradlew --console=plain --parallel functional_test:test --continue
```

Unit tests
```
./gradlew --console=plain --parallel clean test -x functional_test:test -PnoInstrumentation --continue
```

Instrumentation tests
```
./gradlew --console=plain instrumentation:test --continue
```

AITs passed except for Play2 and SNAP flickers: https://javaagent-build.pdx.vm.datanerd.us/job/JavaAgent_PR_AIT/74/testReport/

jasonjkeller (approved):

Issue #293, opened 2021-05-05
Update jdbc-mariaDB-1.3.0 instrumentation


XiXiaPdx:
## Is your feature request related to a problem? Please describe.
jdbc-mariaDB-1.3.0 instrumentation fails against  org.mariadb.jdbc:mariadb-java-client:3.0.0-alpha

It fails because 3.0.0-alpha has two `connect()` metho

https://github.com/mariadb-corporation/mariadb-connector-j/blob/60d535247e3b1fcd7c67a1456d2cef68dd850039/src/main/java/org/mariadb/jdbc/Driver.java#L32-L65d signatures in Driver.java.  1.3.0 to 2.7.2 only has one. 

## Feature Description
Once 3.0.0 settles, see if instrumentation needs to be updated to support. 

## Describe Alternatives
Capped instrumentation verification of jdbc-mariaDb-1.3.0 to  2.7.2, last version before 3.0.0-alpha.

kford-newrelic:
Not selected for the agent roadmap.

PR #294, opened 2021-05-05
cap to last version successfully instrumented

XiXiaPdx:
Why this is necessary is in this issue #293

### DIFF
--- a/instrumentation/jdbc-mariadb-1.3.0/build.gradle
+++ b/instrumentation/jdbc-mariadb-1.3.0/build.gradle
@@ -12,7 +12,7 @@ jar {
 }
 
 verifyInstrumentation {
-    passesOnly 'org.mariadb.jdbc:mariadb-java-client:[1.3.0,2.7.2]'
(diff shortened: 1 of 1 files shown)


instrumentation/jdbc-mariadb-1.3.0/build.gradle
```diff
@@ -12,7 +12,7 @@ jar {
 }
 
 verifyInstrumentation {
-    passesOnly 'org.mariadb.jdbc:mariadb-java-client:[1.3.0,)'
+    passesOnly 'org.mariadb.jdbc:mariadb-java-client:[1.3.0,2.7.2]'
```
jasonjkeller, line 15: You should use the `)` closing parenthesis to cap the range so that it's obvious which version the instrumentation fails on. 

So if it starts failing at 3.0.0 the following indicates that it supports up to, but not including, 3.0.0.

```
passesOnly 'org.mariadb.jdbc:mariadb-java-client:[1.3.0,3.0.0)'
```

jasonjkeller (changes requested):

jasonjkeller (approved):

PR #296, opened 2021-05-06
Changed dropped events to log at FINE instead of WARN

jasonjkeller:
We already log about dropped span events at FINE, this PR does the same for error and insights/custom events instead of logging them at WARN to decrease log verbosity.

```
"Dropped {0} span events out of {1}."
"Dropped {0} error events out of {1}."
"Dropped {0} custom events out of {1}."
```


meiao (approved):
LGTM

PR #297, opened 2021-05-07
Prevent JFR Service tests from running on Java 7/9/10

jasonjkeller:
JFR daemon doesn't support Java 7, 9, or 10. Just Java 8 and 11+.


meiao (approved):
LGTM

Issue #305, opened 2021-05-17
noticeError -  update javadocs, last error wins instead of first


XiXiaPdx:
Todo: 

1.  Update noticeError() javadocs. 
https://docs.newrelic.com/docs/agents/java-agent/api-guides/guide-using-java-agent-api/

2. Update public docs for another config
`ignoreErrorPriority: false`

Context:

[The docs](https://docs.newrelic.com/docs/agents/java-agent/api-guides/guide-using-java-agent-api/) (search for "the first call to noticeError wins") and [javadocs](https://newrelic.github.io/java-agent-api/javadoc/com/newrelic/api/agent/NewRelic.html#noticeError(java.lang.String)) state that the last error reported in a transaction wins.  This is the correct behavior.  The Javadocs still say first error wins.

Old slack conversation about this discrepancy with the docs and how to switch this behavior with a config setting.  

https://newrelic.slack.com/archives/G96BPDE4Q/p1585583512173900

To switch this behavior,  so first noticeError wins and is reported, a config setting in the agent can be changed to `false`.  This does not switch the behavior though for errors seen by tracers versus the noticeError API. Last tracer error still wins with this config setting. 

    `ignoreErrorPriority: false`

Draft PR in docs to add this config - https://github.com/newrelic/docs-website/pull/2353#issuecomment-844446268

tbradellis:
I'll add the ingoreErrorPriority documentation. Once done with that piece I'll close this out.

tbradellis:
Added configuration property PR to public docs:
https://github.com/newrelic/docs-website/pull/2607

PR #306, opened 2021-05-17
switch default ignore error priority

XiXiaPdx:
addresses bug in #305 

One of the functional tests failed because the test was asserting the last error noticed in a transaction would be reported.   With this change, it is the first error that is noticed.

### DIFF
--- a/functional_test/src/test/java/test/newrelic/test/agent/api/ApiTest.java
+++ b/functional_test/src/test/java/test/newrelic/test/agent/api/ApiTest.java
@@ -8,11 +8,8 @@
 package test.newrelic.test.agent.api;
 
 import com.google.common.collect.Maps;
-import com.newrelic.agent.MetricNames;
 import com.newrelic.agent.Transaction;
-import com.newrelic.agent.TransactionData;
-import com.newrelic.agent.TransactionDataList;
-import com.newrelic.agent.TransactionListener;
+import com.newrelic.agent.*;
 import com.newrelic.agent.bridge.AgentBridge;
 import com.newrelic.agent.bridge.TransactionNamePriority;
 import com.newrelic.agent.browser.BrowserConfigTest;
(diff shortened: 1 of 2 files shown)


jasonjkeller:
@XiXiaPdx 

>One of the functional tests failed because the test was asserting the last error noticed in a transaction would be reported. With this change, it is the first error that is noticed.

When did this test start failing exactly?

[This test](https://source.datanerd.us/java-agent/java_agent/blame/61b62c1dc75bdb0beae6d8bd0cd50e4818c6267b/core/functional_test/test/newrelic/test/agent/api/ApiTest.java#L474) was put in place a decade ago, why would it only recently start failing? This implies that agent behavior has recently changed regarding this API, which I wouldn't expect.

It doesn't feel right to me to change this behavior after customers have gotten used to it working a certain way for such a long time. I don't really understand why this issue suddenly manifested and would like to understand that better before we change anything. Do agent specs tell us what the expected behavior should be? Do we not have an AIT in place for this scenario? Perhaps the docs are just wrong?

XiXiaPdx:
When I first submitted this PR, one of the functional tests failed. The test that failed is here - https://github.com/newrelic/newrelic-java-agent/blob/f18215d145bd6992c0fe74a8c503459799e108ca/functional_test/src/test/java/test/newrelic/test/agent/api/ApiTest.java#L941

 It failed because this PR changes the default behavior to report the first error but the test was asserting for the last error.  Before this PR, the agent reported the last error and that is why that test passed. 

I looked through these agent specs and I don't see any specs about this situation - if two errors are reported in a transaction, should first or last be reported.

This issue came up in a GTSE escalation. The customer was expecting the first error to be reported because that is what they saw in our docs and javadocs. 

> @XiXiaPdx
> 
> > One of the functional tests failed because the test was asserting the last error noticed in a transaction would be reported. With this change, it is the first error that is noticed.
> 
> When did this test start failing exactly?
> 
> [This test](https://source.datanerd.us/java-agent/java_agent/blame/61b62c1dc75bdb0beae6d8bd0cd50e4818c6267b/core/functional_test/test/newrelic/test/agent/api/ApiTest.java#L474) was put in place a decade ago, why would it only recently start failing? This implies that agent behavior has recently changed regarding this API, which I wouldn't expect.
> 
> It doesn't feel right to me to change this behavior after customers have gotten used to it working a certain way for such a long time. I don't really understand why this issue suddenly manifested and would like to understand that better before we change anything. Do agent specs tell us what the expected behavior should be? Do we not have an AIT in place for this scenario? Perhaps the docs are just wrong?
> 
> I might just be missing some context but this doesn't feel like something that should suddenly be changing.
> 
> I kicked off an AIT run here: https://javaagent-build.pdx.vm.datanerd.us/job/JavaAgent_PR_AIT/83/

XiXiaPdx:
update docs to last error wins

PR #323, opened 2021-06-11
Weaver violation logging

dhilpipre:
The Java Agent is supposed to log Weaver Violations to be logged to the agent log if the instrumentation package is not part of the Java Agent.  But this does not work.  The custom flag in WeavePackageConfig always defaults to false.   A code change to WeavePackageConfig sets custom to true if the Vendor Id is set to a value.   This allows Weaver violations to be be logged for field instrumentation (or a third party instrumentation vendor).


tbradellis:
A short while back we moved `WeaveViolations` behind the debug flag. `debug: true` in the newrelic.yml.  
Can you confirm those WeaveViolations aren't showing with debug enabled?

However if that is indeed the case and they do not write to the agent log with debug enabled, then we'd want to take this PR without the Lettuce changes and just based off of main.  

Can you confirm the first question and then if the change is still needed push the logging changes based off of main?

dhilpipre:
I did test it with logging_level set to debug and the violations are logged.   But this deviates from what was first told to FIT when it was decided that Weaver Violations would still be logged for external vendors (e.g FIT).   Also if debug does the same thing then the custom flag is useless and should be removed.

jasonjkeller:
@dhilpipre I'm closing this PR as the agent already has the ability to log WeaveViolations. Additionally this PR appears to include many more changes related to lettuce instrumentation which should not be included. Please feel free to open any new PRs if need be and make sure that they're cleaned up to only include proposed changes.

Issue #198, opened 2021-01-26
Spring Reactive DB Drivers - MongoDB Support


tspring:
Add out-of-the-box Java agent support for MongoDB under the Spring Reactive framework

See MMF [doc](https://docs.google.com/document/d/1wrM0B9puN0YR3Wn9jGbNrCu-fw63Di856wh3hlwhYF0/edit?usp=sharing)

Add support for Reactive calls for MongoDB so that database interactions are instrumented by the Java Agent.

MongoDB, Redis, and Cassandra all have native reactive support in [Spring Data](https://spring.io/projects/spring-data). Many relational databases (Postgres, Microsoft SQL Server, MySQL, H2, and Google Spanner) have reactive support via [R2DBC](https://github.com/r2dbc).

Confirm support for async.client.MongoClient

Particularly Mongo will report incorrect metrics for Async which has been an issue for a while.

Review [https://github.com/topics/nrlabs](https://github.com/topics/nrlabs) instrumentation for Mongo async, Lettuce, etc.

Aha! Link: https://newrelic.aha.io/features/JAVA-295

Write instrumentation and instrumentation tests for [MongoDB reactive streams](https://mongodb.github.io/mongo-java-driver/).

jasonjkeller:
The existing mongo instrumentation seems to be partially applying when mongo reactive is being used. This is causing a number of issues (skewed response times, memory pressure) as those instrumentation modules aren't intended to support mongo reactive. See this internal [slack thread](https://newrelic.slack.com/archives/G96BPDE4Q/p1623360414414400) for some additional context.

To prevent such issues when using mongo reactive the existing instrumentation should be disabled.

```
  class_transformer:
    com.newrelic.instrumentation.mongo-3.7:
      enabled: false
    com.newrelic.instrumentation.mongo-3.1:
      enabled: false
    com.newrelic.instrumentation.mongo-3.0:
      enabled: false
```

As part of the work to officially support mongo reactive we need to determine whether or not the old instrument should apply and be supplemented by an additional instrumentation specific to mongo reactive. Or we need to ensure that the old instrumentation modules don't apply at all and only the new mongo reactive instrumentation does.

Stephan202:
@jasonjkeller the instrumentation artifact IDs use `mongodb` rather than `mongo`. Updating your suggestion accordingly, I can confirm that the following works:
```
  class_transformer:
    com.newrelic.instrumentation.mongodb-3.7:
      enabled: false
[...]

PR #341, opened 2021-07-08
Disable weaving package when reactive driver (Mongo 4.x+) is detected…

nathankooij:
… to avoid segment timeouts.

### Overview
Disables `mongodb-3.7` instrumentation when the reactive driver is detected. Previously [`NettyMongoClients`](https://github.com/newrelic/newrelic-java-agent/blob/main/instrumentation/mongodb-3.7/src/main/java/com/mongodb/async/client/NettyMongoClients.java) was used for this purpose, but this class was removed in MongoDB 4.x+ drivers. As such, the instrumentation becomes active again (when using MongoDB 4.x+) and we start seeing segment timeouts. Instead we can use the new [`MongoClients`](https://github.com/mongodb/mongo-java-driver/blob/master/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/MongoClients.java) class from the reactive driver.

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/198 (since it could be solved by adding instrumentation for the reactive driver)

### Checks

[ x ] Are your contributions backwards compatible with relevant frameworks and APIs?
~[ ] Does your code contain any breaking changes? Please describe.~
~[ ] Does your code introduce any new dependencies? Please describe.~

If there are any questions or this is not the right approach let me know, as right now we have to disable the MongoDB instrumentation altogether as pointed out [here](https://github.com/newrelic/newrelic-java-agent/issues/198#issuecomment-876322287).

### DIFF
--- a/instrumentation/mongodb-3.7/src/main/java/com/mongodb/reactivestreams/client/MongoClients.java
+++ b/instrumentation/mongodb-3.7/src/main/java/com/mongodb/reactivestreams/client/MongoClients.java
@@ -1,0 +1,14 @@
+/*
+ *
+ *  * Copyright 2020 New Relic Corporation. All rights reserved.
+ *  * SPDX-License-Identifier: Apache-2.0
+ *
+ */
+
+package com.mongodb.reactivestreams.client;
+
+import com.newrelic.api.agent.weaver.SkipIfPresent;
+
+// This instrumentation will cause Segment timeouts if it applies to the reactive driver
+@SkipIfPresent
+public class MongoClients {}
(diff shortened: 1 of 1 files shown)


Stephan202 (approved):
LGTM, based on https://github.com/newrelic/newrelic-java-agent/issues/198#issuecomment-876587083.

jasonjkeller:
I tested this with an [app using the Mongo 4.x sync drivers](https://github.com/jasonjkeller/spring-boot-mongodb-base-project) and our 7.0.0 agent versus an agent with the changes proposed here:

```
mongodb-driver-core | 4.1.1
mongodb-driver-sync | 4.1.1
```

The `mongodb-3.7` instrumentation loaded and weaved the following classes in both cases:

```
Supportability/WeaveInstrumentation/Loaded/com.newrelic.instrumentation.mongodb-3.7/1

Supportability/WeaveInstrumentation/WeaveClass/com.newrelic.instrumentation.mongodb-3.7/com/mongodb/MongoClientSettings$Builder
Supportability/WeaveInstrumentation/WeaveClass/com.newrelic.instrumentation.mongodb-3.7/com/mongodb/MongoClientSettings
```

So it doesn't look like this change prevents the `mongodb-3.7` instrumentation from working with 4.x non-reactive Mongo, which is awesome.

jasonjkeller (approved):
LGTM 🚀

PR #348, opened 2021-07-21
remove bintray and jcenter, update dependency versions

XiXiaPdx:
anorm and play instrumentation module dependencies do not resolve through bintray and the agent does not compile and build

This removes dl.bintray.  Also, this removes jcenter because it goes away Feb 2022. 

A few of the module's dependencies had to be updated. They are not available on jcenter or mavenCentral.

### DIFF
--- a/instrumentation/anorm-2.0/build.gradle
+++ b/instrumentation/anorm-2.0/build.gradle
@@ -1,17 +1,11 @@
 apply plugin: 'scala'
 
-repositories {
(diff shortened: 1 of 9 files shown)


XiXiaPdx:
AIT passed - https://javaagent-build.pdx.vm.datanerd.us/job/JavaAgent_PR_AIT/102/

meiao (approved):

PR #367, opened 2021-08-06
Remove potential NPE

twcrone:


tbradellis (approved):
One quick question for my own understanding...
Would it make sense to use the same convention here for consistency (or is it not worth changing the code since the Node is already null checked)?
https://github.com/newrelic/newrelic-java-agent/blob/53cbca0a6a6c7a2215ea6e672f3cb871d3689245/newrelic-agent/src/main/java/com/newrelic/agent/tracers/jasper/GeneratorVisitBodyTracerFactory.java#L77
otherwise, lgtm

PR #371, opened 2021-08-16
Enabling configuration of jfr params

meiao:
### Overview
Enables configuration of queue size and harvest interval for JFR data.

### Related Github Issue
https://github.com/newrelic/newrelic-jfr-core/issues/203

### Checks

Yes
No
No

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/config/JfrConfigImpl.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/config/JfrConfigImpl.java
@@ -7,6 +7,8 @@
 
 package com.newrelic.agent.config;
 
+import com.newrelic.jfr.daemon.DaemonConfig;
+
 import java.util.Collections;
 import java.util.Map;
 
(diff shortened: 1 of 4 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/config/JfrConfigImpl.java
```diff
@@ -18,12 +20,19 @@
     public static final String AUDIT_LOGGING = "audit_logging";
     public static final Boolean AUDIT_LOGGING_DEFAULT = Boolean.FALSE;
     public static final Boolean USE_LICENSE_KEY_DEFAULT = Boolean.TRUE;
+    public static final String HARVEST_INTERVAL = "harvest_interval";
+    public static final String QUEUE_SIZE = "queue_size";
```
jasonjkeller, line 24: If the intent is for these to be publicly configurable then we'll probably want them in the default yaml: https://github.com/newrelic/newrelic-java-agent/blob/main/newrelic-agent/src/main/resources/newrelic.yml#L274-L289

And also in the usage section of the readme: https://github.com/newrelic/newrelic-jfr-core#usage-for-new-relic-java-agent-jfr-service

Thats said, I'm not sure if the intent is for these to be publicized as of yet...

jasonjkeller (approved):

kford-newrelic:
Feature Request submitted by a customer:

> We would like to have the option to control the JFR EVENTS QUEUE SIZE on the Java APM agent side via an environment variable or newrelic.yml config file attribute.
> 
> This is so that apps that generate more than the New Relic designated default JFR event queue size don't lose that information before it's sent over to NewRelic.
> 
> If we can control the JFR Event Queue Size, we could also effectively stop the NR java agent spamming logs with "Getting error Rejecting RecordedEvent – queue is full" messages without having to disable JFR.
> 
> We are looking for functionality similar to the one described in (this PR)

jasonjkeller:
When deciding if the agent should impose limits on these new config options we should consider the limits imposed by the Metric API and Event API that JFR is sent to so that we can minimize the likelihood of data being rejected by the backend:
- https://docs.newrelic.com/docs/data-apis/ingest-apis/metric-api/metric-api-limits-restricted-attributes/
- https://docs.newrelic.com/docs/data-apis/ingest-apis/event-api/introduction-event-api/#limits

Also consider adding supportability metrics to record the values configured by customers to make troubleshooting easier.

PR #401, opened 2021-09-02
Fix errors related to newrelic.com domain

jasonjkeller:
These tests were making external calls to the `newrelic.com` domain and failing on Java 9 & 10.

The assertion failures were because not all of the expected `External` metrics were being generated:

```
External/allOther
External/newrelic.com/all
External/all
External/newrelic.com/HttpURLConnection
```

The underlying cause seemed to be related to SSL cert issues only present on Java 9 & 10.

```
javax.net.ssl.SSLHandshakeException: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
```

It's a bad idea to repeatedly hit the `newrelic.com` domain in tests in the first place and switching it to `example.com`, which is a domain dedicated for testing, resolved the issues.


tbradellis (approved):
lgtm

PR #412, opened 2021-09-09
refactor util name, private final fields

XiXiaPdx:
pr feedback also remove test folder (that shouldn't have been there)


instrumentation/httpclient-jdk11/src/test/java/com/nr/agent/instrumentation/httpclientjdk/HttpClientTest.java
```diff
@@ -1,70 +0,0 @@
-package com.nr.agent.instrumentation.httpclientjdk;
```
twcrone, line 1: So I guess this test should be removed?

twcrone (approved):

PR #423, opened 2021-09-10
Prevent CompileTestScala from running under noInstrumentation

meiao:
### Overview
Prevents compileTestScala from running when noInstrumentation profile is on.

### Checks

Yes

No

No


meiao:
This did not work as intended.

PR #426, opened 2021-09-13
Remove anorm-2.0 instrumentation module

jasonjkeller:
This deletes the `anorm-2.0` instrumentation module as the artifacts that it instrumented can no longer be publicly downloaded. Furthermore, the range of artifacts that it instrumented (`[2.0,2.3-M1)`) were all released in 2015 or prior and are likely not worth supporting any longer.

Supportability metrics for instrumentation modules across all accounts during the past month: https://onenr.io/0nVjYVW4eQ0

Usage of this version of the anorm instrumentation is fairly low:

Screen Shot 2021-09-13 at 4 22 22 PM

Compared to more popular instrumentation modules:

Screen Shot 2021-09-13 at 4 25 07 PM


instrumentation/anorm-2.0/build.gradle
```diff
@@ -1,39 +0,0 @@
-apply plugin: 'scala'
-
-dependencies {
-    implementation(project(":newrelic-api"))
-    implementation(project(":agent-bridge"))
-    implementation(project(":newrelic-weaver-api"))
-    implementation("org.scala-lang:scala-library:2.10.7")
-    implementation("com.typesafe.play:anorm_2.10:2.3.9")
```
jasonjkeller, line 8: This was incorrectly updated to compile against `com.typesafe.play:anorm_2.10:2.3.9` because that is the oldest version available on maven. However, this instrumentation module is only supposed to instrument anorm versions `2.0` up to, but not including `2.3-M1`. Thus it was compiling against an unsupported version.

This meant that the instrumentation was weaving a method that no longer existed in version `2.3.9` which was leading to the following `WeaveViolation`s:

```
     WeaveViolation{type=METHOD_NEW_CALL_ORIGINAL_UNSUPPORTED, clazz=anorm/Sql$, method=as(Lanorm/ResultSetParser;Ljava/sql/ResultSet;)Ljava/lang/Object;}
     WeaveViolation{type=METHOD_NEW_NON_PRIVATE_UNSUPPORTED, clazz=anorm/Sql$, method=as(Lanorm/ResultSetParser;Ljava/sql/ResultSet;)Ljava/lang/Object;}
```

XiXiaPdx (approved):
🪓

PR #431, opened 2021-09-14
Exclude problematic artifact for jedis-3.0.0 instrumentation verifier

jasonjkeller:
Exclude `redis.clients:jedis:3.6.2` from verification as there is something weird about that version that causes failures.


tbradellis (approved):
lgtm

PR #432, opened 2021-09-14
Correctly verify against Scala 2.13 instead of 2.12

jasonjkeller:
This instrumentation module was supposed to be verifying against Scala 2.13 instead of 2.12. All versions pass now.


tbradellis (approved):
lgtm

Issue #413, opened 2021-09-09
Enable DT by default


jasonjkeller:
Distributed tracing should be [enabled by default in the agent](https://github.com/newrelic/newrelic-java-agent/blob/main/newrelic-agent/src/main/java/com/newrelic/agent/config/DistributedTracingConfig.java#L14) and the [default yaml](https://github.com/newrelic/newrelic-java-agent/blob/main/newrelic-agent/src/main/resources/newrelic.yml#L207-L214) that ships with the agent.

```
  distributed_tracing:
    enabled: true
```

What is the behavior when the customer has CAT enabled via server side config? Does that always override any other settings?

Ensure that all test suites (unit, functional, instrumentation, and AITs) pass after the defaults have been changed.

PR #439, opened 2021-09-15
Enable DT by default

jasonjkeller:
Turns on DT by default and updates the yaml

https://javaagent-build.pdx.vm.datanerd.us/job/JavaAgent_PR_AIT/122/

Resolves https://github.com/newrelic/newrelic-java-agent/issues/413


jasonjkeller:
Lots of test failures to resolve with this change.

Issue #410, opened 2021-09-08
Incorrect Transaction Name with AbstractController @RequestMapping


twcrone:
## Description
In a Spring web app with controllers that extend an `abstract` controller that has an `@RequestMapping` that is shared among multiple controllers, the transaction name for the controller endpoints lose the `method` and `path` as they would show normally.

## Expected Behavior
Transaction names should still be `/path/to/resource (HTTP_METHOD)`

## Steps to Reproduce
[Spring Pet Clinic with Abstract controller request mapping](https://github.com/twcrone/spring-pet-clinic-abstract-controller-issue)
Details on repro in the README.md.

<img width="681" alt="Screen Shot 2021-09-08 at 1 24 47 PM" src="https://user-images.githubusercontent.com/653013/132558308-ffa8a102-38fc-49fb-a34c-9aee6933dbdc.png">

twcrone:
My research thus far finds that it need not be an `abstract` controller nor must it be a controller endpoint that extends a controller.  The mere fact that a controller extends a base class with a route messes up other routes.

twcrone:
Conducted a week-long research spike. Findings:

1. Having a base controller class with routes that are inherited is not what the instrumentation was originally designed to support.
2. Adding this functionality would require significant changes to Spring instrumentation that might negatively impact other more commonly seen use cases.
3. More significant research is necessary to consider deep level changes to the core instrumentation packages.
4. These changes need to be weighed against impact to the vast number of users already relying on the Spring instrumentation as it is today.

Bringing this issue to management attention for review. No instrumentation changes made.

Note : A viable workaround is moving shared controller functionality into a utility class that can be wired into controllers that need it (i.e. “Favor composition over inheritance”).  This would allow the instrumentation to work as it was originally designed.

PR #467, opened 2021-10-03
fixes issue with transaction naming for abstract spring controller cl…

tspring:
…asses that are not annotated. GH issue #410

Hey old agent friends @XiXiaPdx @jasonjkeller @tbradellis, hope all is well with y'all! Here's a couple of bugs if you want them :)

### Overview
This adds instrumentation for annotated spring methods whose class is not annotated. Same idea as the jax-rs instrumentation. I used the controller class name for cases where there is no root or method path specified. I also did nothing about earlier versions of Spring which may or may not support a similar pattern.

Screenshot from 2021-10-03 13-16-48

### Related Github Issue
#410

### Testing
I included some instrumentation tests.

### Checks

[ y] Are your contributions backwards compatible with relevant frameworks and APIs?
[ n] Does your code contain any breaking changes? Please describe. 
[ n] Does your code introduce any new dependencies? Please describe.

:wave:


twcrone:
Thanks @tspring !  We have discussed this and decided not to merge at this time.  Instead we are advising a workaround that will give them a partial transaction name and prevent a failure that cascades into other transaction naming.

If you have advice on how we could easily include the `rootPath` from the ChildController when the inherited route is called, let us know.  That would be the ideal solution but without lots of work and risk to existing users, the team has decided not to pursue at this time.

kford-newrelic:
@tspring thanks again for this PR!

When we explained the current agent design limitation to the developers that reported this, they didn't object to modifying their class design so this was no longer an issue for them.

PR #481, opened 2021-10-12
Fix test failures on Java 11+ and Scala 2.13

jasonjkeller:
We’re seeing failures with some of our Scala related instrumentation tests (e.g. `com.nr.instrumentation.akka22.test.AkkaTest`) that appear to be happening only on Java 11+, which I think are likely due to this bug with Scala 2.13: https://github.com/scala/bug/issues/12340

We’re seeing the same failure:

```
java.lang.IllegalAccessError: Update to static final field akka.pattern.package$.MODULE$ attempted from a different method (<init>) than the initializer method <clinit> 
```

This can be repro’d locally against main by running the following:

```
./gradlew -Ptest11 :instrumentation:akka-2.2:test --tests "com.nr.instrumentation.akka22.test.AkkaTest" --stacktrace
```

This PR prevents those tests from running on Java 11+


apiwoni:
I seem to have issues on Java 11 with versions v7.2.0 or greater which added Scala Library Intrumentation. I run Scala 2.11 and not Scala 2.13.

PR #525, opened 2021-10-28
Provide support for Datastax/Cassandra WrappedStatments.

stringaling:
A wrapped statement is a class that the driver supplies to support custom Retry and LoadBalancing policies.

If developers implement and extend [com.datastax.driver.core.StatementWrapper](https://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/core/StatementWrapper.html) as part of a custom Retry or LoadBalancing policy, the statements were being rejected as ```unsupported statement type```.

### Overview
Changes include:
* checking if the statement is a WrappedStatement
* if it is, get the statement that is being wrapped and use it as the Query to use for metrics. Otherwise use the original statement.
* Unit tests to verify Wrapped statements work
* Applied to version 3.0.0 and 3.8.0 of the Datastax driver. (This class looks like it may have been present in some version of 2.0 - but not in the dependency included in this 2.0.0 project).

### Related Github Issue
Since it may be called a bug ? I didn't create a github issue for it.

### Testing

### Checks

[Yes] Are your contributions backwards compatible with relevant frameworks and APIs?
[No] Does your code contain any breaking changes? Please describe. 
[NO] Does your code introduce any new dependencies? Please describe.

### DIFF
--- a/instrumentation/cassandra-datastax-3.0.0/src/main/java/com/datastax/driver/core/NewRelicChainedResultSetFuture.java
+++ b/instrumentation/cassandra-datastax-3.0.0/src/main/java/com/datastax/driver/core/NewRelicChainedResultSetFuture.java
@@ -60,11 +60,18 @@ public class NewRelicChainedResultSetFuture extends ChainedResultSetFuture {
         }
         String host = queriedHost != null ? queriedHost.getHostName() : null;
         Integer port = queriedHost != null ? queriedHost.getPort() : null;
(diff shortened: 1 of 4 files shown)


jasonjkeller:
Thanks for the contribution @stringaling ! We're reviewing.

jasonjkeller:
Branch rebased against `main` for running internal tests: https://github.com/newrelic/newrelic-java-agent/tree/pr-525

jasonjkeller (approved):

PR #535, opened 2021-11-04
Run Java tests against 17 instead of 16

twcrone:
### Overview
Change GHA workflows for java to use 17 instead of 16


XiXiaPdx (approved):
LGTM.

PR #536, opened 2021-11-04
Add reusable workflow for Java instrumentation tests

twcrone:
### Overview
Trying out reusable workflow for Java instrumentation tests

### DIFF
--- a/.github/workflows/X-Reusable-Test.yml
+++ b/.github/workflows/X-Reusable-Test.yml
@@ -1,34 +1,27 @@
-## Matrix strategy
-name: GHA-Instrumentation-Tests
+name: X - Reusable Test Workflow 
 
 on:
-  pull_request:
-    branches: [ disabled ] ## branches TBD
-    types: [assigned, opened, synchronize, reopened, ready_for_review] ## Types TBD
-  push:
-    # branches: ['**']
-    branches: [disabled]
-  # Allows you to run this workflow manually from the Actions tab
-  workflow_dispatch:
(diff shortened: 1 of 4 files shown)


XiXiaPdx (approved):
LGTM!

.github/workflows/X-Reusable-Test.yml
```diff
@@ -0,0 +1,134 @@
...
+      # Check ENV variables
+      - name: Check environmental variables
+        run: printenv | sort -f
+
+      ## AWS jars - plan to cache
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.aws-access-key-id }}
+          aws-secret-access-key: ${{ secrets.aws-secret-access-key }}
```
meiao, line 82: Can't the reusable workflow access the repo secrets directly?
That would save a few lines in the concrete workflows.

twcrone: It failed when accessing directly from reusable.

.github/workflows/X-Reusable-Test.yml
```diff
@@ -0,0 +1,134 @@
+name: X - Reusable Test Workflow
```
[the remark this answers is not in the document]
twcrone: If we keep it, we can rename to something appropriate, but we may need only one reusable workflow and perhaps not one specific to instrumentation.  We shall see.

PR #538, opened 2021-11-06
Spring security

### LINKED ISSUE #353: Instrumentation needed for Spring Security Config Starting with bundles for Spring Boot 2.2.2 (Spike)
Starting in Spring Boot bundle 2.2.2 there is a change in how SecurityConfig is used.  
A new thread gets started, causing our current instrumentation to lose the Transaction context and name the Transaction by the dispatcher. The controller instrumentation doesn't get included on the Transaction so the naming is wrong and part of the Transaction is missing. 

## Feature Description

We need to add support for the new SecurityConfig.

## Additional context

Internal GTSE-9940

## Priority

Please help us better understand this feature request by choosing a priority from the following options: Really Want

Aha! Link: https://newrelic.aha.io/features/JAVA-333

jasonjkeller:
Fixes #353 

Properly links `Tokens` across contexts when Spring Security is used (and other potential use cases).

[UFI tests](https://javaagent-build.pdx.vm.datanerd.us/view/Java%20Agent/job/Unit_Functional_Instrumentation_Tests_Configurable/52/): 🟢 
[AIT](https://javaagent-build.pdx.vm.datanerd.us/job/AIT_Configurable/202/): 🟢


instrumentation/netty-reactor-0.8.0/README.md
```diff
@@ -0,0 +1,36 @@
...
+Instrumentation for Reactor Netty server and also some widely used Reactor Core library code.
+
+This module is largely responsible for instrumenting the Reactor Core library to facilitate the passing, retrieval,
+and linking of `Tokens` across contexts to tie asynchronous threads together for individual `Transactions`.
+
+This instrumentation is dependent on other instrumentation modules to start a `Transaction`.
+Typically, the `netty-n.n` modules work with this instrumentation and will start a `Transaction` (see `NettyDispatcher#channelRead`).
+
+Most commonly this instrumentation comes into play with SpringBoot usage, in which case the `spring` and `spring-webflux` 
+instrumentation modules also apply and should result in `Transactions` being renamed after the Spring controller.
```
XiXiaPdx, line 12: What do you think about calling out here in the README the Tokens mentioned earlier are created by ServerWebExchange in Spring WebFlux instrumentation? 

Or, there is already a comment in the code for Netty Reactor TokenLinkingSubscriber.  I think that comment should mention ServerWebExchange so the next person doesn't have to wonder about where that token comes from.

instrumentation/netty-reactor-0.8.0/src/main/java/reactor/core/scheduler/Schedulers_Instrumentation.java
```diff
@@ -0,0 +1,45 @@
...
+        final Scheduler cached;
+        final String key;
+
+        CachedScheduler(String key, Scheduler cached) {
+            /*
+             * Add tokenLift hook if it hasn't already been added. This allows for tokens to be retrieved from
+             * the current context and linked across threads at various points of the Flux/Mono lifecycle.
+             *
+             * When using Netty Reactor with SpringBoot this hook will be added by the HttpTrafficHandler_Instrumentation
+             * but when using other embedded web servers (e.g. Tomcat, Jetty, Undertow) the HttpTrafficHandler class
```
XiXiaPdx, line 31: I feel like there is another piece of relevant piece of work here (for the future, after this PR). 

If the question comes up and people wonder if this instrumentation works for a non-netty server, I've seen it does from your tests.  It could be good to create an issue to confirm this behavior for a 1-2 most popular non-netty servers and create AITS for them.

XiXiaPdx (approved):
This is great, ❤️‍🔥  all the documentation makes it so much easier to dive back into this code in the future!!!

instrumentation/netty-reactor-0.8.0/src/main/java/reactor/core/scheduler/Schedulers_Instrumentation.java
jasonjkeller: Agreed! I'm hoping to look into that before wrapping up this work.

instrumentation/netty-reactor-0.8.0/README.md
jasonjkeller: Yea, I think the readme definitely has room for improvement, especially around the behavior of how the interactions between instrumentation modules work. I need to dive into that a bit deeper.

jasonjkeller: Updated the comment in `TokenLinkingSubscriber` to specifically mention the token being created in `ServerWebExchange`.

instrumentation/netty-reactor-0.8.0/src/main/java/reactor/core/scheduler/Schedulers_Instrumentation.java
jasonjkeller: @XiXiaPdx This should do it.

This [PR](https://source.datanerd.us/java-agent/agent-integration-tests/pull/65) updates the AITs to test Spring WebFlux with two different SpringBoot apps.
- SpringBoot configured with Spring Security and Reactor Netty as the embedded web server
- SpringBoot configured with Spring Security and Tomcat as the embedded web server

PR #559, opened 2021-11-16
Obfuscate sql for external

XiXiaPdx:
WIP 
AIT 🟢  https://javaagent-build.pdx.vm.datanerd.us/job/AIT_Configurable/203/

Sample Test App - https://github.com/XiXiaPdx/mysql-agent-testapp

significant changes:

1. Obfuscate early in DefaultSqlTracer 
2. removing an agent attribute being set in DefaultSlowQueryListener.  
3. Documentation

Needs tests and one todo to resolve

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/sql/DefaultSlowQueryListener.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/sql/DefaultSlowQueryListener.java
@@ -43,20 +43,20 @@ public class DefaultSlowQueryListener implements SlowQueryListener {
     @Override
     public <T> void noticeTracer(Tracer tracer, SlowQueryDatastoreParameters<T> slowQueryDatastoreParameters) {
         if (tracer.getDurationInMilliseconds() > thresholdInMillis) {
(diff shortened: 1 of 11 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/sql/DefaultSlowQueryListener.java
```diff
@@ -43,7 +43,7 @@ public DefaultSlowQueryListener(String appName, double thresholdInMillis) {
     @Override
     public <T> void noticeTracer(Tracer tracer, SlowQueryDatastoreParameters<T> slowQueryDatastoreParameters) {
         if (tracer.getDurationInMilliseconds() > thresholdInMillis) {
-            String query = (String) slowQueryDatastoreParameters.getQuery();
+            String query = slowQueryDatastoreParameters.getQuery().toString();
```
meiao, line 46: This fixed a ClassCastException, but it is an oversimplification that may not work for all query objects from different databases.
Ideally here it should use the queryConverter in the SlowQueryDatastoreParameters instead of assuming the toString call will return a query.

PR #561, opened 2021-11-17
Fix scala version to nothing when not specified

twcrone:
Fix scala version to nothing when not specified

### Checks

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -46,9 +46,12 @@ gradleEnterprise {
         } else {
             tag "Local"
         }
-        tag System.getenv("LANGUAGE") ?: "polyglot"
         tag System.getenv("JRE") ?: "jre-not-specified"
-        tag System.getenv("SCALA_VERSION") ?: "no-scala"
+        tag System.getenv("LANGUAGE") ?: "polyglot"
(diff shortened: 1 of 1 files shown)


settings.gradle
```diff
@@ -46,9 +46,12 @@ gradleEnterprise {
...
             tag "Local"
         }
-        tag System.getenv("LANGUAGE") ?: "polyglot"
         tag System.getenv("JRE") ?: "jre-not-specified"
-        tag System.getenv("SCALA_VERSION") ?: "no-scala"
+        tag System.getenv("LANGUAGE") ?: "polyglot"
+        String scalaVersion = System.getenv("SCALA_VERSION")
+        if (scalaVersion) {
+            tag scalaVersion
+        }
```
jasonjkeller, line 54: What is this even used for? And why would the behavior be different for Scala? What previously would happen if `"no-scala"` was returned as the value and what happens now if nothing is returned?

twcrone: Its just a tag on the build.  So you can search on particular things.  "no-scala" was showing up for Scala Functional Test runs because they are not version specific.  If there is nothing for a tag, it just doesn't have that tag.

jasonjkeller (approved):

Issue #376, opened 2021-08-19
expected_status_codes is not honored when there is an exception in the transaction


meiao:
## Description
Setting a `expected_status_code` will not mark an error as expected when there is an exception in the transaction.

## Expected Behavior
If a transaction returns with a status code that is in the `expected_status_code` an error from this transaction should be marked as expected regardless of the existence of an exception.

This is how `ignored_status_code` works.

meiao:
This is likely due to the difference between ThrowableError and HttpTracedError.

meiao:
The repository below has a simple app to reproduce the problem.
https://github.com/meiao/repro-nr-376

sdmcintyre:
Any idea when this might get looked at? This causes many false alarms in our monitoring due to things like input validation errors (HTTP bad request).

tbradellis:
I think the problem starts here with `markExpected:`
https://github.com/newrelic/newrelic-java-agent/blob/bf6cb39760c3d5e66d8614cd698a95bcb381c347/newrelic-agent/src/main/java/com/newrelic/agent/errors/ErrorServiceImpl.java#L441

We only check whether it's expected on the Throwable. As @meiao hinted on `ThrowableError` and `HttpTracedError`.  When we get to `HttpTracedError` we don't consult the status code to see if we need to set markedExpected differently based on status.  I have a local change and will run some tests with a change for this.
Essentially, I think we should just check if the status code is expected and reset `markedExpected` here:
https://github.com/newrelic/newrelic-java-agent/blob/bf6cb39760c3d5e66d8614cd698a95bcb381c347/newrelic-agent/src/main/java/com/newrelic/agent/errors/ErrorServiceImpl.java#L467

tbradellis:
Correction...we also need to check this for ThrowableError.

tbradellis:
https://github.com/newrelic/newrelic-java-agent/blob/bellis/expected-error-response-status/newrelic-agent/src/main/java/com/newrelic/agent/errors/ErrorServiceImpl.java#:~:text=boolean%20responseStatusExpected%20%3D%20errorCollectorConfig,markedExpected%20%3D%20responseStatusExpected%20%7C%7C%20throwableExpected%3B

sdmcintyre:
Do we need to update our New Relic Agent to apply the fix?

meiao:
Yes. Agent 7.5.0 has the fix.

PR #565, opened 2021-11-18
check response and throwable

tbradellis:
This address the issue described here:
https://github.com/newrelic/newrelic-java-agent/issues/376

Setting a expected_status_code will not mark an error as expected when there is an exception in the transaction.

The problem starts here with `markedExpected`:
[newrelic-java-agent/newrelic-agent/src/main/java/com/newrelic/agent/errors/ErrorServiceImpl.java
](https://github.com/newrelic/newrelic-java-agent/blob/3a39a70c96ae3f339ff33cbcb46556c0beeb8ac8/newrelic-agent/src/main/java/com/newrelic/agent/errors/ErrorServiceImpl.java#L441)

We only check whether it's expected on the Throwable and do not consult the expected status codes....As @meiao hinted on `ThrowableError` and `HttpTracedError`. 

Essentially, we should check if the throwable OR a status code is expected and mark it as such.  This change does that.

### Overview
checks both response status and throwable for if they were configured to be expected.  Previously, we were not checking the response status in this code path, so there are cases where a response status can be configured expected, but the TracedError will not get created as expected.

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/376

### Checks

[ NA] Are your contributions backwards compatible with relevant frameworks and APIs?
[ no] Does your code contain any breaking changes? Please describe. 
[ no] Does your code introduce any new dependencies? Please describe.


newrelic-agent/src/main/java/com/newrelic/agent/errors/ErrorServiceImpl.java
```diff
@@ -438,7 +456,9 @@ private void noticeTransaction(TransactionData td, TransactionStats transactionS
     private TracedError createTracedError(final String theAppName, TransactionData td, Throwable throwable, int responseStatus, String statusMessage) {
         TracedError error;
         // noticeError(expected = true)?
-        boolean markedExpected = td.getThrowable() == null ? false : td.getThrowable().expected;
+        boolean responseStatusExpected = errorCollectorConfig.getExpectedStatusCodes().contains(responseStatus);
```
meiao, line 459: In ErrorAnalyzerImpl#isExpectedError we also do some verification on both the exception and status code.
But if the exception is not expected, it assumes the error is not expected.
Maybe we have that other check for a different report, but I believe it should follow the same rules.

tbradellis: That's interesting...We seem to return early here if the status code is expected:
https://github.com/newrelic/newrelic-java-agent/blob/e25fa98a0967879e0da0f0deaec2ade3ca7fa167/newrelic-agent/src/main/java/com/newrelic/agent/errors/ErrorAnalyzerImpl.java#L102

But, the confusing part is here where we loop through `throwable != null` and do some additional checking, but only on the throwable and not for expected status code.
https://github.com/newrelic/newrelic-java-agent/blob/e25fa98a0967879e0da0f0deaec2ade3ca7fa167/newrelic-agent/src/main/java/com/newrelic/agent/errors/ErrorAnalyzerImpl.java#L107
There is only one TransactionThrowable, so I don't fully understand these lines.  I will step through it and see.

tbradellis: @meiao After stepping through this, I believe this line here:
https://github.com/newrelic/newrelic-java-agent/blob/e25fa98a0967879e0da0f0deaec2ade3ca7fa167/newrelic-agent/src/main/java/com/newrelic/agent/errors/ErrorAnalyzerImpl.java#L102

Is enough to always confirm that the status has been checked in the ErrorAnalyzer.  The status code  parameter for the method is the same as what would be on the Throwable (from what I'm seeing).  Unless, you see some way of breaking it that I'm not understanding.

meiao: Nevermind, the fail fast threw me off. I read it the previous block as if the exception is not expected the error is not expected.

Memory must have played some trick on me, because I thought there was some two step verification in someplace from when I first investigated this issue.

meiao (approved):

PR #611, opened 2021-12-14
Reduce logging granularity

jeffalder:
This message is logged by default with customer input. Push this down to a non-default log level.


tbradellis (approved):
lgtm

PR #618, opened 2021-12-20
Adds logback-classic-1.2 instrumentation

jsubirat:
### Overview
The following PR introduces instrumentation for the Logback logging framework. In particular, this instrumentation counts the amount of log lines emitted in total, as well as faceted by each of the log levels (DEBUG, INFO, etc.).

Note that this instrumentation doesn't simply count how many `logger.info()` calls have been performed, but how many log records have actually been _accepted_ by the framework (that is, logs above the configured log level, and having been accepted by the turbofilters): we instrument the `buildLoggingEventAndAppend` method, which gets called when a log record has been accepted.

### Testing
- Unit tests have been provided that test the new functionality.
- Internal performance tests have been performed using several services, which demonstrate no significant CPU/memory impact. Please get in touch with me to discuss about the results.

### Checks

[x] Does your code contain any breaking changes? Please describe. 


instrumentation/logback-classic-1.2/src/main/java/com/nr/agent/instrumentation/logbackclassic12/Logger_Instrumentation.java
```diff
@@ -0,0 +1,18 @@
...
+import com.newrelic.api.agent.weaver.Weaver;
+import org.slf4j.Marker;
+
+@Weave(originalName = "ch.qos.logback.classic.Logger", type = MatchType.Interface)
+public abstract class Logger_Instrumentation {
+    private void buildLoggingEventAndAppend(final String localFQCN, final Marker marker, final Level level, final String msg, final Object[] params,
+            final Throwable t) {
+        Weaver.callOriginal();
+        NewRelic.incrementCounter(Constants.LOG_LINE.getVal());
+        NewRelic.incrementCounter(String.format(Constants.LOG_LINE_LEVEL.getVal(), level));
```
jasonrclark, line 16: If we're concerned about overhead, is there a strong reason to separate the constants out so we have to use an additional method call to get them? I don't see that done elsewhere in the instrumentation commonly.

instrumentation/logback-classic-1.2/build.gradle
```diff
@@ -0,0 +1,18 @@
+jar {
+    manifest { attributes 'Implementation-Title': 'com.newrelic.instrumentation.logback-classic-1.2' }
+}
+
+dependencies {
+    implementation(project(":agent-bridge"))
+    implementation("ch.qos.logback:logback-core:1.2.6")
```
jasonrclark, line 7: Do we actually rely on anything in `logback-core`? I don't see mention of it in the instrumented classes.

jasonrclark:
Question for the agent folks -- is there already a shared way built in if we needed to disable this instrumentation, or do we need to add a specific configuration of some sort?

instrumentation/logback-classic-1.2/build.gradle
jsubirat: Indeed, we don't! I removed it.

instrumentation/logback-classic-1.2/src/main/java/com/nr/agent/instrumentation/logbackclassic12/Logger_Instrumentation.java
jsubirat: Good point, I'm going to revert this.

dylanmei:
Why shade another logging dependency instead of relying on simple [java.util.logging](https://docs.oracle.com/javase/7/docs/api/java/util/logging/package-summary.html)?

jsubirat:
> Why shade another logging dependency instead of relying on simple [java.util.logging](https://docs.oracle.com/javase/7/docs/api/java/util/logging/package-summary.html)?

We focused on the Logback library since it is one of the most used, way above `java.util.logging`.

jasonrclark:
To clarify @dylanmei this is adding _instrumentation_ for clients who use logback -- not using the library in the agent itself.

dylanmei:
With that argument, we'd stick with Log4j. It came as quite a surprise to many folks last week that New Relic was shading a complex logging library for an APM agent running directly within our JVM processes. This is just trading one for another.

instrumentation/logback-classic-1.2/src/main/java/com/nr/agent/instrumentation/logbackclassic12/Logger_Instrumentation.java
```diff
@@ -0,0 +1,19 @@
...
+import org.slf4j.Marker;
+
+@Weave(originalName = "ch.qos.logback.classic.Logger", type = MatchType.Interface)
+public abstract class Logger_Instrumentation {
+
+    private void buildLoggingEventAndAppend(final String localFQCN, final Marker marker, final Level level, final String msg, final Object[] params,
+            final Throwable t) {
+        Weaver.callOriginal();
+        NewRelic.incrementCounter("Logging/lines");
+        NewRelic.incrementCounter(String.format("Logging/lines/%s", level));
```
jasonjkeller, line 17: Typically we want to do New Relic specific stuff first and then return execution to the original code, which means that the `Weaver.callOriginal()` should come last. I assume that the value of `level` isn't dependent on the original code executing first and that it's merely the value that was passed in to this method as an argument? If that's true, then let's move the NewRelic API calls ahead of `Weaver.callOriginal()`.

instrumentation/logback-classic-1.2/build.gradle
```diff
@@ -0,0 +1,18 @@
...
+}
+
+dependencies {
+    implementation(project(":agent-bridge"))
+    implementation("ch.qos.logback:logback-classic:1.2.6")
+}
+
+verifyInstrumentation {
+    passes("ch.qos.logback:logback-classic:[1.0,)")
+    excludeRegex 'ch.qos.logback:logback-classic:1.3.*'
```
jasonjkeller, line 12: This regex exclusion is too broad as it will prevent the instrumentation from applying to any legitimate 1.3.x versions that are released.

Does the instrumentation actually fail to verify against the `1.3.0-alphaX` artifact versions? Or are you simply excluding them because they're `alpha` releases? If it's the latter, then the regex should be updated to only capture artifacts with `alpha` in their names. If this instrumentation does begin failing against 1.3.X versions, then the excludeRegex should be removed and the `passes` range should be changed to `passesOnly("ch.qos.logback:logback-classic:[1.0.0,1.3.0-alpha0)")`. This will ensure that all artifacts from 1.0.0 up to, but not including 1.3.0-alpha0, pass verification while verifying that all other artifacts outside the range fail against the instrumentation.

jasonjkeller:
> Question for the agent folks -- is there already a shared way built in if we needed to disable this instrumentation, or do we need to add a specific configuration of some sort?

Nothing special needs to be done. The `Implementation-Title` defined in the `build.gradle` can be used to disable the instrumentation as follows:

```
  class_transformer:
    com.newrelic.instrumentation.logback-classic-1.2:
      enabled: false
```

Is this instrumentation that we want enabled by default?

instrumentation/logback-classic-1.2/build.gradle
jasonjkeller: Ok, I ran this against the instrumentation verifier and the following succeeds. That block should be changed to the following as the instrumentation applies to versions 0.9.3+ (though 0.9.6 fails for some reason, hence the exclude):

```
verifyInstrumentation {
    passesOnly("ch.qos.logback:logback-classic:[0.9.3,)")
    excludeRegex '.*(alpha|groovyless).*'
    excludeRegex 'ch.qos.logback:logback-classic:0.9.6'
}
```

instrumentation/logback-classic-1.2/src/main/java/com/nr/agent/instrumentation/logbackclassic12/Logger_Instrumentation.java
jasonjkeller, line 17: It's probably marginal but I think it will actually be more performant to use concatenation:

```
NewRelic.incrementCounter("Logging/lines/" + level);
```

Instead of `String.format()`:

```
NewRelic.incrementCounter(String.format("Logging/lines/%s", level));
```

jasonrclark:
Great feedback @jasonjkeller! I definitely learned some new stuff.

> Is this instrumentation that we want enabled by default?

Conversations internally have been that we will. Let's chat about it this week and we can loop back. There was also some internal conversation around naming that we should probably settle before we land anything here.

jasonjkeller:
FYI I've created a [logging-initiative](https://github.com/newrelic/newrelic-java-agent/tree/logging-initiative) feature branch to merge PRs like this into and I've updated this PR to use that base instead of `main`. We'll have all PRs related to the logging initiative target the `logging-initiative` feature branch so that nothing makes it into an official agent release until absolutely ready.

danybmx:
Hi @jasonjkeller!

Comments from the review addressed on 5c73287, thanks for the review and the feedback on this!

instrumentation/logback-classic-1.2/src/main/java/com/nr/agent/instrumentation/logbackclassic12/Logger_Instrumentation.java
```diff
@@ -0,0 +1,19 @@
...
+import com.newrelic.api.agent.weaver.Weaver;
+import org.slf4j.Marker;
+
+@Weave(originalName = "ch.qos.logback.classic.Logger", type = MatchType.Interface)
+public abstract class Logger_Instrumentation {
+
+    private void buildLoggingEventAndAppend(final String localFQCN, final Marker marker, final Level level, final String msg, final Object[] params,
+            final Throwable t) {
+        NewRelic.incrementCounter("Logging/lines");
+        NewRelic.incrementCounter("Logging/lines" + level);
```
jasonjkeller, line 16:
```suggestion
        NewRelic.incrementCounter("Logging/lines/" + level);
```

jasonjkeller (changes requested):

instrumentation/logback-classic-1.2/src/main/java/com/nr/agent/instrumentation/logbackclassic12/Logger_Instrumentation.java
danybmx: Ops 🤦🏻‍♂️, nice catch @jasonjkeller

jasonjkeller (approved):

Issue #623, opened 2021-12-27
OOM errors and memory leak caused by caffeine cache and java8


T00fy:
We recently upgraded the agent in our java8 application from 7.1.1 to 7.4.2 due to the log4shell exploit. We've noticed since then that very randomly and intermittently a single container can encounter a memory leak that eventually causes an OOM exception. Initial troubleshooting lead me to this issue which looks very similar to the issue I am encountering https://github.com/newrelic/newrelic-java-agent/issues/558

## Description
We deploy multiple docker containers, and randomly after a set amount of time one of these will eventually hit some condition which causes the GC to not clean up unused resources. We have been able to capture a heap dump at the time of this memory leak and the suspected issue using MAT is com.newrelic.agent.deps.com.github.benmanes.caffeine.cache.WS. For comparison, in containers not experiencing a memory leak or running on an older version of NR agent we do not see this in the report.

## [Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results
I have a large heap dump file which for obvious reasons I am not going to post in here, but if you could point me to any relevant details to provide which may be of help I would be happy to oblige.

So far I have just been using MAT to explore the dump file and there are a LOT of objects being instantiated in the caffeine cache with a WeakKeyReference

<img width="628" alt="image" src="https://user-images.githubusercontent.com/12580207/147425161-07276911-2453-4250-9d0c-dbd1bf9ee585.png">

<img width="773" alt="image" src="https://user-images.githubusercontent.com/12580207/147425204-3856e7e4-97e4-4081-8ff7-75c9d4a0ff25.png">

<img width="692" alt="image" src="https://user-images.githubusercontent.com/12580207/147425260-5e9f8c0c-f9cd-4dd3-9249-de9afba5b4ad.png">

<img width="795" alt="image" src="https://user-images.githubusercontent.com/12580207/147425534-d08ad250-2ae6-472b-811d-fdfac581c886.png">

## Your Environment
Newrelic agent 7.4.2
openJDK 1.8.0_292-8u292

## Additional context
Looks very similar to https://github.com/newrelic/newrelic-java-agent/issues/558 but I am running java8 and not upgrading to java 17 or 11. There was a lot of discussion in that issue about a bug in the JDK but I'm not convinced that is the issue unless it is also affecting java8 somehow.

[...]

PR #626, opened 2021-12-29
Log4shell fix for 7.2.0

T00fy:
**I am unable to create a branch referencing tag v7.2.0, if this change is accepted then the base branch this merges into needs to change.**

### Overview
Later versions of NR agent which include changes with a caffeine cache have a memory leak issue. Version 7.2.0 was the last version before this change was included. In the interim until this issue is fixed, a user should be able to decide if they want to use a patch of 7.2.0 in order to not have their application vulnerable to log4shell.

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/623

### Testing

### Checks

[x] Does your code contain any breaking changes? Please describe. 
**The PR removes a reference to anorm 2.0, I cannot pull this as bintray has been shutdown and I cannot find it from other sources on the internet.**
**log4j 2.17.1**

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 # The agent version.
-agentVersion=7.2.0
+agentVersion=7.2.1
 
 newrelicDebug=false
 org.gradle.jvmargs=-Xmx2048m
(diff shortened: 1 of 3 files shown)


meiao:
We will not be releasing new patch versions at this moment.
https://docs.newrelic.com/docs/security/new-relic-security/security-bulletins/security-bulletin-nr21-03/

If you must use 7.2.0 you can disable the agent logs and it should be safe.

PR #627, opened 2021-12-30
Fix cacheability for JavaCompile, ScalaCompile and Test tasks

ghale:
### Overview
These changes fix the cacheability of of JavaCompile, ScalaCompile and Test tasks.  There are a few changes:

- Use toolchains instead of executable for java compilation
- Set "test.jdk" input property to avoid getting false cache hits across the crossversion jdk test tasks using the same JDK version
- Unset executable for ScalaCompile to workaround Gradle defect
- Refactor the play-2.6, play-2.6.13 and play-2.7 instrumentation projects to make their Scala test compile tasks cacheable
- Add cacheability to the `writeCachedWeaveAttributes` tasks

A couple of assumptions/caveats:
- The crossversion jdk tests tasks should really use toolchains instead of properties to select different JDKs.  This would allow the build to be specific about what the expectations are for each testing scenario.  However, the selection criteria in its current form only allows the specification of version, vendor and implementation (see https://github.com/gradle/gradle/blob/c53d50f1e81500f2f84f98d2e2e3f65739c15c66/subprojects/platform-jvm/src/main/java/org/gradle/jvm/toolchain/JavaToolchainSpec.java).  This is a problem for this use case because some of the JDKs that are being tested against use the same vendor/version/implementation. 
 https://github.com/gradle/gradle/issues/18896 addresses adding further criteria (such as `java.vm.name`).  Once that is available, this could be converted to using toolchains as well and eliminate the need for the JDK properties.
- The newrelic-agent test task was setting the source/target compatibility, but was always running tests with whatever JDK the build was running with.  I've converted that to use toolchains and keep the same behavior, but one of the tests _can't_ run with a JDK < 9 (`JfrServiceTest`).  This should really target a specific JDK, rather than relying on the build process JDK.
- The routes compilation for the play-2.6, play-2.6.13 and play-2.7 instrumentation projects was generating sources into the source directory of the test compile task.  This meant that the task was _never_ up-to-date or cacheable since every time it would run, it would change its inputs.  I've changed it so that it generates the sources outside of the source directories and set these up as local state of the task (neither an input or an output) and added the routes file as an input.  This should make these tasks up-to-date and cacheable now.  Since it was generating the Routes source files outside of the build directory, it would not clean these up with the "clean" task and you could get duplicate class files.  To fix this, I added `src/test/scala/routes` to the clean task, but this is only necessary to clean up what it was generating before.  There are other options for this if this approach seems too heavy-handed.
- The newrelic-api Javadoc task had a conditional set of arguments if the build was run with JDK11.  Rather than try to react to an unknown condition (i.e. the JDK the build process is executing with) it seemed better to just specifically use JDK11.  This can obviously be handled differently if there is some reason why we shouldn't be more specific here.

### Related Github Issue
N/A

### Testing

### Checks

There should be no impacts on backwards compatibility.
[x ] Does your code contain any breaking changes? Please describe. 
There should not be any, beyond the minor assumptions stated above.
No.

### DIFF
--- a/agent-interfaces/build.gradle
+++ b/agent-interfaces/build.gradle
@@ -7,11 +7,3 @@ dependencies {
     api(project(":newrelic-api"))
     implementation(project(":agent-model"))
 }
-
-sourceCompatibility = JavaVersion.VERSION_1_8
-targetCompatibility = JavaVersion.VERSION_1_8
-
-compileTestJava {
-    sourceCompatibility = JavaVersion.VERSION_1_8
-    targetCompatibility = JavaVersion.VERSION_1_8
-}
--- a/agent-model/build.gradle
+++ b/agent-model/build.gradle
@@ -5,11 +5,3 @@ plugins {
 dependencies {
     api 'com.googlecode.json-simple:json-simple:1.1'
 }
-
-sourceCompatibility = JavaVersion.VERSION_1_8
-targetCompatibility = JavaVersion.VERSION_1_8
-
-compileTestJava {
-    sourceCompatibility = JavaVersion.VERSION_1_8
-    targetCompatibility = JavaVersion.VERSION_1_8
-}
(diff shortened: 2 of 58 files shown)


twcrone (approved):
Looks good to me and AITs run green too!

Issue #628, opened 2022-01-02
Make using the agent with future java versions configurable as an experimental feature.


tbradellis:
It should be easier to attempt to run an agent on `ea` or other future versions of the JDK.
We currently check in `JavaVersionUtils` to see if the agent has been loaded in a supported Java version. We fail and exit if we detect the agent has been loaded in an unsupported version (both old and new jdk versions).

I propose we create a configurable bypass to allow for experimental usage of an agent build in any forward JDK.  This would allow for some simple smoke testing or other usages without updating all of the test suites or add hoc changing of JavaVersionUtils.

## Describe Alternatives
continue as we do.  Create a special branch and change what's needed to build and run on a future jdk.

Given the faster cadence of jdk releases, this would help us be more nimble in getting ready for future releases and staying on top of breaking changes.  Similarly, it would make it easier to do ad hoc tests of the agent on experimental JDK branches.

## Priority
Nice to have

tbradellis:
https://github.com/newrelic/newrelic-java-agent/pull/629

PR #629, opened 2022-01-02
add bypass for agent version chck

tbradellis:
Feature https://github.com/newrelic/newrelic-java-agent/issues/628
This adds a configuration option (must use system property or environment variable) that bypasses the agents check on supported version. 
We currently check in JavaVersionUtils to see if the agent has been loaded in a supported Java version. We fail and exit if we detect the agent has been loaded in an unsupported version (both old and new jdk versions).

This bypass is to allow for experimental usage of an agent build in any forward JDK. This would allow for some simple smoke testing or other usages without having to do a build with changes  to JavaVersionUtils.

Describe Alternatives
continue as we do. Create a special branch and change what's needed to build and run on a future java runtime.

Given the faster cadence of jdk releases, this would help us be more nimble in getting ready for future releases and staying on top of breaking changes. Similarly, it would make it easier to do ad hoc tests of the agent on experimental JDK branches.

to use, build and configure either:
system property

`-Dnewrelic.config.experimental_runtime=true
`

ENVVAR

`NEW_RELIC_AGENT_EXPERIMENTAL_RUNTIME=true
`
### Checks

[ yes ] Are your contributions backwards compatible with relevant frameworks and APIs?
[ no ] Does your code contain any breaking changes? Please describe. 
[ no  ] Does your code introduce any new dependencies? Please describe.

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/Agent.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/Agent.java
@@ -9,11 +9,7 @@ package com.newrelic.agent;
 
 import com.google.common.collect.ImmutableMap;
 import com.newrelic.agent.bridge.AgentBridge;
-import com.newrelic.agent.config.AgentJarHelper;
-import com.newrelic.agent.config.ConfigService;
-import com.newrelic.agent.config.ConfigServiceFactory;
-import com.newrelic.agent.config.JarResource;
-import com.newrelic.agent.config.JavaVersionUtils;
+import com.newrelic.agent.config.*;
 import com.newrelic.agent.core.CoreService;
 import com.newrelic.agent.core.CoreServiceImpl;
 import com.newrelic.agent.logging.AgentLogManager;
@@ -35,16 +31,8 @@ import java.io.IOException;
 import java.lang.instrument.Instrumentation;
 import java.net.URL;
 import java.text.MessageFormat;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.ResourceBundle;
-import java.util.jar.Attributes;
-import java.util.jar.JarEntry;
-import java.util.jar.JarFile;
-import java.util.jar.JarOutputStream;
-import java.util.jar.Manifest;
+import java.util.*;
+import java.util.jar.*;
 import java.util.logging.Level;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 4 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/Agent.java
```diff
@@ -294,13 +282,24 @@ private static boolean isLicenseKeyEmpty(String licenseKey) {
 
     public static void main(String[] args) {
         String javaSpecVersion = JavaVersionUtils.getJavaSpecificationVersion();
-        if (!JavaVersionUtils.isAgentSupportedJavaSpecVersion(javaSpecVersion)) {
+        String sysExperimentalRuntime = System.getProperty("newrelic.config.experimental_runtime");
+        String envExperimentalRuntime = System.getenv("NEW_RELIC_AGENT_EXPERIMENTAL_RUNTIME");
```
jasonjkeller, line 286: I believe that `System.getProperty` and `System.getenv` can potentially cause Exceptions that need to be guarded against.

https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#getProperty-java.lang.String-
https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#getenv-java.lang.String-

Also we shouldn't be hardcoding the the same strings in multiple places throughout the codebase (e.g. `"newrelic.config.experimental_runtime"` and `"NEW_RELIC_AGENT_EXPERIMENTAL_RUNTIME"`). It's better if they can be centralized into constants with the proper access level.

newrelic-agent/src/main/java/com/newrelic/bootstrap/BootstrapAgent.java
```diff
@@ -88,7 +93,19 @@ static String decodeAndDecompressAgentArguments(String agentArgs) throws IOExcep
      */
     public static void premain(String agentArgs, Instrumentation inst) {
         String javaSpecVersion = JavaVersionUtils.getJavaSpecificationVersion();
-        if (!JavaVersionUtils.isAgentSupportedJavaSpecVersion(javaSpecVersion)) {
+        String sysExperimentalRuntime = System.getProperty("newrelic.config.experimental_runtime");
+        String envExperimentalRuntime = System.getenv("NEW_RELIC_AGENT_EXPERIMENTAL_RUNTIME");
```
jasonjkeller, line 97: Same comment as the previous regarding `System.getProperty` and `System.getenv`.

twcrone (approved):

newrelic-agent/src/main/java/com/newrelic/agent/Agent.java
tbradellis: Both  System.getProperty() and System.getenv() are effectively guarded by Boolean.parseBoolean()..because if parseBoolean(null) takes a null parameter then it will return false.  

` 
   public static boolean parseBoolean(String s) {
        return ((s != null) && s.equalsIgnoreCase("true"));
    }
    
`

newrelic-agent/src/main/java/com/newrelic/bootstrap/BootstrapAgent.java
tbradellis: Both  System.getProperty() and System.getenv() are effectively guarded by Boolean.parseBoolean()..because if parseBoolean(null) takes a null parameter then it will return false.  

` 
   public static boolean parseBoolean(String s) {
        return ((s != null) && s.equalsIgnoreCase("true"));
    }
    
`

newrelic-agent/src/main/java/com/newrelic/agent/Agent.java
tbradellis: After discussing with @twcrone on a review we decided to keep the string values for now. This pattern was chosen to go with how `agent_enabled` is done for the same classes.  The options needs to be set at a similar time and way as agent_enabled.

https://github.com/newrelic/newrelic-java-agent/blob/2cc6d9643e9060732940351950176e07172e824c/newrelic-agent/src/main/java/com/newrelic/bootstrap/BootstrapAgent.java#L98

If it is preferable to use a global static constant we decided that we should do that for each case in a separate PR to "clean-up" the usage in the same way for consistency.

Issue #568, opened 2021-11-18
SqlObfuscator does not obfuscate certain html documents 


XiXiaPdx:
It was reported and confirmed that the agent does not obfuscate certain html documents that are parameters in queries. https://newrelic.atlassian.net/browse/GTSE-12323

In that linked issue, the document-content.json is the parameter that fails to be fully obfuscated. 

On this branch, I added a test that uses the resource and is failing to obfuscate. 
https://github.com/newrelic/newrelic-java-agent/tree/fail-html-obfuscate

Expected Behavior: 

The html document is obfuscated into `?`

To Resolve:

What  exactly is slipping through the SqlObfuscator?

Does the agent spec for sql obfuscation cover this use case? If it doesn't, this is a feature request and would require cross agent spec changes and discussion.

meiao:
This is due the way that queries incoming from PreparedStatements are treated.
There is also a related issue that a query that escapes single quotes using `\'` will have any content after that hidden.
[Internal link](https://newrelic.slack.com/archives/G96BPDE4Q/p1640819017373000) with more context.

PR #632, opened 2022-01-04
SqlObfuscator does not obfuscate certain html documents

meiao:
### Overview
When recreating a "raw" query from a PreparedStatement, the values from the params were added to the "raw" query without escaping any quotes inside of it.
This caused some problems later on when the obfuscator would try to match the quotes to remove any value in the query.

Fixed by escaping single quotes (unless it was already escaped using a `\`) prior to adding values to the "raw" query.

### Related Github Issue
#568 

### Testing
Tests were added to verify the quotes were being properly escaped.

### Checks


newrelic-agent/src/test/java/com/newrelic/agent/tracers/DefaultSqlTracerTest.java
```diff
@@ -543,14 +543,67 @@ public void allParametersTrailingSpaces() throws Exception {
     public void missingParameters() throws Exception {
         String sql = "SELECT * FROM ROGER WHERE COL1 = ? AND COL2 = ? AND COL3 = ? AND COL4 = ?";
         String expectedSql = "SELECT * FROM ROGER WHERE COL1 = 'String1' AND COL2 = 1.0 AND COL3 = 1 AND COL4 = ?";
-        Map<Integer, Object> parameters = new HashMap<>();
-        parameters.put(1, "String1");
-        parameters.put(2, 1.0f);
-        parameters.put(3, 1);
+
```
meiao, line 546: This map was never used.

twcrone (approved):

PR #634, opened 2022-01-04
issue #344 completable future tokens not expired correctly

### LINKED ISSUE #344: CompletableFutures instrumentation potentially causes token timeouts
## Description
CompletableFutures instrumentation currently creates tokens whenever a CompletableFuture is created. We assume that CF will be completed, whether success or error. When completed, we catch all the finish points to expire our token.

But the common usage pattern for CompletableFutures do not guarantee every CF be completed. It is typical to create a CF for success and for error. The instrumentation would then create a token for both but potentially only one will be completed.  This results in an unexpired token that will not be removed from the token cache until it times out (default is 180 seconds). 

The consequence of this is a Transaction will not finish until the token times out. So a transaction will not be reported until 3-4 minutes after it occurred. From early testing, the transaction duration is not affected (3 minutes is not added to the transaction duration). There is just a delay in reporting and also long lived tokens.

## Expected Behavior
CompletableFutures instrumentation would not create unexpired tokens, causing token timeouts. 

A potential next step  - https://newrelic.slack.com/archives/G96BPDE4Q/p1575831177351600

## Additional context
This behavior can be observed with the WIP JDK11 HttpClient instrumentation and also the AWS sdk dynamo V2  instrumentation.

richard-gibson:

### Overview
CompletableFutures instrumentation currently creates tokens whenever a CompletableFuture is created. We assume that CF will be completed, whether success or error. When completed, we catch all the finish points to expire our token.

New implementation only creates Tokens on execution of CompletionStage rather than creation of the Structure.
Token is expired on completion of Execution.

### Related Github Issue
Resolves #344 

### Testing

### Checks


functional_test/build.gradle
```diff
@@ -2,8 +2,6 @@ import com.nr.builder.JarUtil
 
 evaluationDependsOn(":newrelic-agent") // This is important because we need newrelic-agent to be configured before functional_test so the correct ("unversioned") jar name gets used
 
-apply plugin: 'scala'
```
richard-gibson, line 5: scala plugin not needed for functional tests

jasonjkeller (approved):

Issue #560, opened 2021-11-16
MongoDB instrumentation won't apply to the sync client if async/reactivestreams client is also present


jasonjkeller:
MongoDB instrumentation won't apply to the sync client if async/reactivestreams client is also present.

This was an unexpected consequence of changes that were released in agent 7.4.0 in an effort to prevent the MongoDB instrumentation from applying to the async/reactivestreams client because it was causing token timeouts and incorrect query response times.

For context: https://github.com/newrelic/newrelic-java-agent/issues/198#issuecomment-951395245

If only the sync client is present then the instrumentation should still apply. 

One workaround is to roll back to agent 7.3.0 until this can be addressed in a new release.

Related to: https://github.com/newrelic/newrelic-java-agent/issues/198

pmthortful:
Hi, 

Just wanted to flag that we are experiencing this issue, and due to not wanting to be on an old version of the agent due to the recent log4j vulnerability, we currently have 0 visibility on our mongo database operations

kford-newrelic:
Hi @pmthortful , while not ideal, if this issue is significantly impacting you, you could rollback to the `7.3.0` agent and **disable** agent logging (to compensate for the log4j vulnerability). This isn't ideal because if you have any static code scans going, the `7.3.0` release will show up but it's relatively straightforward to confirm that agent logging is turned `OFF`.

jasonjkeller:
@pmthortful I've submitted a PR that should restore visibility into MongoDB queries that is planned to be in our next release coming out shortly (which will also have the latest Log4j patches included).

PR #635, opened 2022-01-06
Revert changes that prevented mongodb-3.1 from applying when the asyn…

jasonjkeller:
…c/reactive client was present.

This essentially reverts changes from https://github.com/newrelic/newrelic-java-agent/pull/476 that resulted in the `mongodb-3.1` instrumentation no longer applying if the async/reactivestreams clients were also being used.

After further investigation it doesn't appear that the `mongodb-3.1` instrumentation is causing the same token timeout issues that the `mongodb-3.7` instrumentation was.

This will result in instrumentation applying for uses of the mongodb sync client version 3.1.0 and above once again.

Resolves: https://github.com/newrelic/newrelic-java-agent/issues/560


twcrone (approved):

twcrone:
Looks straightforward

PR #644, opened 2022-01-11
Updating version some instrumentation modules apply to

meiao:
### Overview
Some instrumentation modules had the versions that they apply to outdated due to new releases of the libraries.
This PR fixes that and also improves some of the checks to certify that a module is not applied to a version of the library it should not apply to.

### DIFF
--- a/instrumentation/cats-effect-2/build.gradle
+++ b/instrumentation/cats-effect-2/build.gradle
@@ -16,9 +16,14 @@ jar {
 }
 
 verifyInstrumentation {
-    passes 'org.typelevel:cats-effect_2.13:[2.1,)'
-    passes 'org.typelevel:cats-effect_2.12:[2.1,)'
-    excludeRegex 'org.typelevel:cats-effect_2.1(2|3):3.*'
+    // scala 12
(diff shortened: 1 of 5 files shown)


instrumentation/cats-effect-2/build.gradle
```diff
@@ -16,9 +16,14 @@ jar {
...
-    passes 'org.typelevel:cats-effect_2.13:[2.1,)'
-    passes 'org.typelevel:cats-effect_2.12:[2.1,)'
-    excludeRegex 'org.typelevel:cats-effect_2.1(2|3):3.*'
+    // scala 12
+    passes 'org.typelevel:cats-effect_2.12:[2.1,3.0)'
+    fails 'org.typelevel:cats-effect_2.12:[3.0,)'
+
+    // scala 13
+    passes 'org.typelevel:cats-effect_2.13:[2.1,3.0)'
+    fails 'org.typelevel:cats-effect_2.13:[3.0,)'
```
meiao, line 25: Tried to use passesOnly, but it does apply to some versions prior to 2.1.
Didn't want to spend too much time fixing it.

Tickets were created for each module that the instrumentation no longer applies.

tbradellis (approved):
lgtm

PR #678, opened 2022-01-25
Fix Java 17 Instrumentation tests

twcrone:
We need to move reusable workflow to a different repo

### DIFF
--- a/.github/workflows/X-Reusable-Test.yml
+++ b/.github/workflows/X-Reusable-Test.yml
@@ -68,6 +68,12 @@ jobs:
           distribution: 'temurin'
           java-version: 17
 
+      # Save new JDK variable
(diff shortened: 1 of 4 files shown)


.github/workflows/X-Reusable-Test.yml
```diff
@@ -68,6 +68,12 @@ jobs:
           distribution: 'temurin'
           java-version: 17
 
+      # Save new JDK variable
+      - name: Save JAVA_HOME as JDK11 for later usage
```
meiao, line 72: oops, copy pasta 11

twcrone: Thanks!

meiao (approved):

PR #684, opened 2022-01-27
Adding attribute Automatic-Module-Name for JDK9/11

sgtjim:

### Overview
Our platform is updating from Java 8 to Java 11. We're building module-info.java files and we're currently experiencing an error because the `requires newrelic.api` module is not found.  My understanding is that this identifier needs to be in the MANIFEST.MF file.

### Related Github Issue
No linked issue available

### Checks

[X ] Are your contributions backwards compatible with relevant frameworks and APIs?

### DIFF
--- a/newrelic-api/build.gradle
+++ b/newrelic-api/build.gradle
@@ -8,7 +8,7 @@ plugins {
 jar {
     from("$rootDir/LICENSE")
     manifest {
-        attributes 'Implementation-Title': 'New Relic Agent API', 'Implementation-Version': project.version
+        attributes 'Implementation-Title': 'New Relic Agent API', 'Implementation-Version': project.version, 'Automatic-Module-Name': 'newrelic.api'
     }
 }
 


tbradellis:
I'm curious why that's failing without the `Automatic-Module-Name`.  I don't think it should be based on the javadocs for `ModuleFinder`

From what I can tell the naming should resolve the same as proposed.  The `newrelic-api.jar` should be auto-named to newrelic.api as described here:

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/module/ModuleFinder.html

The [version](https://docs.oracle.com/javase/9/docs/api/java/lang/module/ModuleDescriptor.html#version--), and the module name when the attribute "Automatic-Module-Name" is not present, are derived from the file name of the JAR file as follows:

The ".jar" suffix is removed.

If the name matches the regular expression "-(\\d+(\\.|$))" then the module name will be derived from the subsequence preceding the hyphen of the first occurrence. The subsequence after the hyphen is parsed as a [Version](https://docs.oracle.com/javase/9/docs/api/java/lang/module/ModuleDescriptor.Version.html) and ignored if it cannot be parsed as a Version.

All non-alphanumeric characters ([^A-Za-z0-9]) in the module name are replaced with a dot ("."), all repeating dots are replaced with one dot, and all leading and trailing dots are removed.

As an example, **a JAR file named "foo-bar.jar"** will derive a module name "foo.bar" and no version. A JAR file named "foo-bar-1.2.3-SNAPSHOT.jar" **will derive a module name "foo.bar"** and "1.2.3-SNAPSHOT" as the version.

I'm not against updating this and getting this pr merged, but making sure there's no additional impact first.

tbradellis:
@sgtjim Out of curiosity, did you change the newrelic-api.jar name?  Or let me know if you have any comment on my comments above.

sgtjim:
Hello, I must've missed the notification about the first reply so I'm sorry about the delay.
To answer the second question;  No I didn't change the newrelic-api.jar name.  All this change is doing is adding an attribute to the Manifest.MF that's generated when the code is compiled into a jar.

So my platform may be several years behind, so we're just now rolling into the module system that's been implemented in JDK9.  From what I've gathered, as we're trying to break our own platform into different modules, we're creating `module-info.java` files at the root of each..well.. module.  This is talking to the java compiler and saying what resources the module will need access to, and similarly what resources will be exported from the module.

The module-info.java file is pulling these resource/dependency names from each dependencies Manifest.MF file and more specifically the "Automatic-Module-Name" attribute.
For example, if we wanted to use Apache Commons Lang3 for StringUtils, etc, we would add a line item:
`requires org.apache.commons.lang3`,  which is generated here:
https://github.com/apache/commons-lang/blob/master/pom.xml#L792

And if you're able to, you can open your local `org.apache.commons:commons-lang3:3.<whatever>` and look at the Manifest.MF file to see what I mean (or pretty much any other relatively-updated jar).

So, we're trying to continue to use the newrelic-api in our platform, but we're being blocked because the compiled jar doesn't have that attribute.  So this is simply adding it for modularity.

Sorry about the novel and I hope that answered your question :) Thanks!

tbradellis:
Sounds good. Thanks @sgtjim . We'll get this merged. If you want to rebase this branch it should pass our tests now and I'll approve and merge, or for expediency I can merge from here which will keep you as the author:
https://github.com/newrelic/newrelic-java-agent/pull/708
Either way works for us.

sgtjim:
If you can merge it from there that would be great, thanks.  I'll watch that to see if anyone has concerns that I can help with

Issue #641, opened 2022-01-10
Add new instrumentation for Jedis driver 4.0.x+


meiao:
Jedis released version 4.0.0 and our current instrumentation no longer applies.

tbradellis:
https://github.com/newrelic/newrelic-java-agent/pull/698

PR #698, opened 2022-02-08
Jedis 4.0.0

tbradellis:
This adds instrumentation for Jedis 4.0.0+)
Maintains the same instrumentation points as previous versions. JedisPubSub and Connection instrumentation.

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/641
https://github.com/redis/jedis/tree/4.0
### Testing
Uses verifyInstrumentation in the usual way.
Additional integration test with local apps to ensure correct counts and no loss of instrumentation features.

### Checks

[ NA] Are your contributions backwards compatible with relevant frameworks and APIs?


tbradellis:
This is ready for review. I have one question concerning a new method.   In this new method there is a `ProtocolCommand` and a `Keyword`
https://github.com/redis/jedis/blob/ea18faa55728de4697ecf7f504aa826491a752d6/src/main/java/redis/clients/jedis/Connection.java#L139

Example:
https://github.com/redis/jedis/blob/ea18faa55728de4697ecf7f504aa826491a752d6/src/main/java/redis/clients/jedis/Jedis.java#L3377
This code already reports the command given. I suppose we could include the keyword as an attribute, but I don't know if it is necessary.    I'm currently not tracing this method, but if we aren't going to report the Keyword, then tracing is not needed because it is handed in `sendCommand(CommandArguments args)`
 What do you think?

instrumentation/jedis-3.0.0/build.gradle
```diff
@@ -13,7 +13,7 @@ dependencies {
 verifyInstrumentation {
     passes 'redis.clients:jedis:[3.0.0,4.0.0)'
     fails 'redis.clients:jedis:[1.4.0,3.0.0)'
-    fails 'redis.clients:jedis:[4.0.0,)'
+//    fails 'redis.clients:jedis:[4.0.0,)'
```
tbradellis, line 16: Oops! I'll clean that up.

instrumentation/jedis-4.0.0/src/main/java/redis/clients/jedis/JedisPubSub_Instrumentation.java
```diff
@@ -0,0 +1,90 @@
...
+@Weave(type = MatchType.BaseClass, originalName = "redis.clients.jedis.JedisPubSub")
+public class JedisPubSub_Instrumentation {
+
+    @NewField
+    private String host;
+
+    @NewField
+    private int port;
+
+    @Trace
```
meiao, line 29: Any reason to trace this new method?
It may cause confusion when people upgrade Jedis and see a new span in their traces.

instrumentation/jedis-4.0.0/build.gradle
```diff
@@ -0,0 +1,25 @@
...
+    implementation(project(":agent-bridge-datastore"))
+    implementation(project(":newrelic-api"))
+    implementation(project(":newrelic-weaver-api"))
+    implementation("redis.clients:jedis:4.0.0")
+}
+
+verifyInstrumentation {
+    passes 'redis.clients:jedis:[4.0.0,)'
+    fails 'redis.clients:jedis:[1.4.0,3.8.0]'
+//    fails 'redis.clients:jedis:[4.0.0,)'
```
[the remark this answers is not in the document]
tbradellis: done

tbradellis:
@aonuki I noticed that. maybe there is an opportunity in this JedisPubSub_Instrumentation class to also drop some @NewField  usage and save a little memory (like in the Connection_Instrumentation).
https://github.com/newrelic/newrelic-java-agent/blob/4d9df067a0d761eb0b5250912d8645111dcb4b04/instrumentation/jedis-4.0.0/src/main/java/redis/clients/jedis/JedisPubSub_Instrumentation.java#L35

The only thing I see that might make a difference is the number of calls I’d need to make to get the host and port in each of the methods. Do you think this  would be a good change here or would leaving it in this case make sense for any reason? Readability or anything else?

meiao:
In JedisPubSub, you could move the call to inside the `reportMethodAsExternal` method and remove the host and port parameters.

tbradellis:
Oh, derp...yes that could be done the same too. Thanks!!

PR #718, opened 2022-02-23
Logging initiative

jasonjkeller:
Logs in context based on agent instrumentation


tbradellis (approved):
lgtm

instrumentation-test/src/main/java/com/newrelic/agent/introspec/internal/IntrospectorLogSenderService.java
```diff
@@ -0,0 +1,140 @@
...
+ *
+ */
+
+package com.newrelic.agent.introspec.internal;
+
+import com.newrelic.agent.Agent;
+import com.newrelic.agent.config.AgentConfig;
+import com.newrelic.agent.deps.com.google.common.collect.LinkedListMultimap;
+import com.newrelic.agent.deps.com.google.common.collect.ListMultimap;
+import com.newrelic.agent.deps.com.google.common.collect.Maps;
```
twcrone, line 14: For my education, I want to understand this usage.

instrumentation/apache-log4j-1/src/main/java/com/nr/agent/instrumentation/log4j1/AgentUtil.java
```diff
@@ -0,0 +1,98 @@
...
+    private static final String SPAN_ID = "span.id";
+    // Enabled defaults
+    private static final boolean APP_LOGGING_DEFAULT_ENABLED = true;
+    private static final boolean APP_LOGGING_METRICS_DEFAULT_ENABLED = true;
+    private static final boolean APP_LOGGING_FORWARDING_DEFAULT_ENABLED = false;
+    private static final boolean APP_LOGGING_LOCAL_DECORATING_DEFAULT_ENABLED = false;
+
+    /**
+     * Gets a String representing the agent linking metadata in blob format:
+     * NR-LINKING|entity.guid|hostname|trace.id|span.id|
```
twcrone, line 37: Is entity.name supposed to be here now?

instrumentation/apache-log4j-1/src/test/java/com/nr/agent/instrumentation/log4j1/Category_InstrumentationTest.java
```diff
@@ -0,0 +1,67 @@
...
+@RunWith(InstrumentationTestRunner.class)
+@InstrumentationTestConfig(includePrefixes = { "org.apache.log4j" }, configName = "application_logging_enabled.yml")
+public class Category_InstrumentationTest {
+
+    private static final String CAPTURED = "This log message should be captured";
+    private static final String NOT_CAPTURED = "This message should NOT be captured";
+
+    @Test
+    public void shouldIncrementEmittedLogsCountersIndependentlyIfLogLevelEnabled() {
+        // Given
```
twcrone, line 21: I really like the "Given, When, Then" format.

instrumentation/apache-log4j-2/src/main/java/com/nr/agent/instrumentation/log4j2/AgentUtil.java
```diff
@@ -0,0 +1,136 @@
...
+import com.newrelic.api.agent.NewRelic;
+import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.core.LogEvent;
+import org.apache.logging.log4j.message.Message;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class AgentUtil {
+    public static final int DEFAULT_NUM_OF_LOG_EVENT_ATTRIBUTES = 3;
```
twcrone, line 20: Are these constants going to be the same across most logging instrumentation?  Does it make any sense to put in core rather than instrumentation modules?

instrumentation/apache-log4j-2/src/main/java/com/nr/agent/instrumentation/log4j2/AgentUtil.java
```diff
@@ -0,0 +1,136 @@
...
+
+                    AgentBridge.getAgent().getLogSender().recordLogEvent(logEventMap);
+                }
+            }
+        }
+    }
+
+    /**
+     * Gets a String representing the agent linking metadata in blob format:
+     * NR-LINKING|entity.guid|hostname|trace.id|span.id|
```
twcrone, line 75: entity.name

instrumentation/apache-log4j-2/src/main/java/com/nr/agent/instrumentation/log4j2/AgentUtil.java
twcrone: I'm not suggesting doing it now even if it makes sense.  Once we add more instrumentation, perhaps it will make sense to centralize.

instrumentation/java.logging-jdk8/src/main/java/java/util/logging/Logger_Instrumentation.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public boolean isLoggable(Level level) {
+        return Boolean.TRUE.equals(Weaver.callOriginal());
+    }
+
+    public void log(LogRecord record) {
+        // Do nothing if application_logging.enabled: false
+        if (isApplicationLoggingEnabled()) {
+            if (isApplicationLoggingMetricsEnabled()) {
+                if (isLoggable(record.getLevel()) && getFilter() == null || getFilter().isLoggable(record)) {
```
twcrone, line 32: Can record ever be null at this point?

instrumentation/logback-classic-1.2/src/main/java/ch/qos/logback/classic/Logger_Instrumentation.java
```diff
@@ -0,0 +1,46 @@
...
+import com.newrelic.api.agent.weaver.Weaver;
+import com.nr.agent.instrumentation.logbackclassic12.AgentUtil;
+import org.slf4j.Marker;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+@Weave(originalName = "ch.qos.logback.classic.Logger", type = MatchType.ExactClass)
+public abstract class Logger_Instrumentation {
+    @NewField
+    public static AtomicBoolean instrumented = new AtomicBoolean(false);
```
twcrone, line 24: Can this be final?

instrumentation/logback-classic-1.2/src/main/java/com/nr/agent/instrumentation/logbackclassic12/AgentUtil.java
```diff
@@ -0,0 +1,127 @@
...
+                logEventMap.put(LOG_LEVEL, level);
+            }
+
+            AgentBridge.getAgent().getLogSender().recordLogEvent(logEventMap);
+        }
+    }
+
+    /**
+     * Gets a String representing the agent linking metadata in blob format:
+     * NR-LINKING|entity.guid|hostname|trace.id|span.id|
```
twcrone, line 66: entity.name

newrelic-agent/src/main/java/com/newrelic/agent/config/ApplicationLoggingForwardingConfig.java
```diff
@@ -0,0 +1,50 @@
...
+    private final int maxSamplesStored;
+
+    public ApplicationLoggingForwardingConfig(Map<String, Object> props, String parentRoot, boolean highSecurity) {
+        super(props, parentRoot + ROOT + ".");
+        maxSamplesStored = initMaxSamplesStored();
+        boolean storedMoreThan0 = maxSamplesStored > 0;
+        enabled = storedMoreThan0 && !highSecurity && getProperty(ENABLED, DEFAULT_ENABLED);
+    }
+
+    private int initMaxSamplesStored() {
```
twcrone, line 33: Who wrote this crap?  ;-)

newrelic-agent/src/main/java/com/newrelic/agent/service/ServiceFactory.java
```diff
@@ -78,6 +79,10 @@ public static ConfigService getConfigService() {
         return SERVICE_MANAGER.getConfigService();
```
twcrone, line 79: What is THIS ServiceManager?

newrelic-agent/src/main/java/com/newrelic/agent/AgentLinkingMetadata.java
```diff
@@ -0,0 +1,141 @@
...
+     * Get a map of all agent linking metadata.
+     *
+     * @param traceMetadata TraceMetadataImpl instance to get spanId and traceId
+     * @param configService ConfigService to get hostName and entityName
+     * @param rpmService    IRPMService to get entityGuid
+     * @return Map of all agent linking metadata
+     */
+    public static Map<String, String> getLinkingMetadata(TraceMetadata traceMetadata, ConfigService configService, IRPMService rpmService) {
+        AgentConfig agentConfig = configService.getDefaultAgentConfig();
+        Map<String, String> linkingMetadata = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
twcrone: I would expect it to just use the fastest Map impl locally, then return an immutable copy, but for agent perf etc. pernhaps we create a Concurrent one for concurrent mutation by the caller?

newrelic-agent/src/main/java/com/newrelic/agent/service/logging/LogSenderServiceImpl.java
```diff
@@ -0,0 +1,585 @@
...
+            // Even if the transaction is canceled we still want to send up any events that were held in it
+            TransactionLogs data = (TransactionLogs) transaction.getLogEventData();
+            storeEvents(transaction.getApplicationName(), transaction.getPriority(), data.events);
+        }
+    };
+
+    /**
+     * Listener to detect changes to the agent config
+     */
+    protected final AgentConfigListener configListener = new AgentConfigListener() {
```
twcrone, line 95: My Education...So why do we use anonymous classes like this throughout agent?

newrelic-agent/src/main/java/com/newrelic/agent/service/logging/LogSenderServiceImpl.java
```diff
@@ -0,0 +1,585 @@
...
+    /**
+     * Get the LogEvent reservoir
+     *
+     * @param appName app name
+     * @return Queue of LogEvent instances
+     */
+    @VisibleForTesting
+    public DistributedSamplingPriorityQueue<LogEvent> getReservoir(String appName) {
+        DistributedSamplingPriorityQueue<LogEvent> result = reservoirForApp.get(appName);
+        while (result == null) {
```
twcrone, line 343: So is there any chance this will never be non-null?  I'm guessing not but at a glance it scares me a bit.

instrumentation-test/src/main/java/com/newrelic/agent/introspec/internal/IntrospectorLogSenderService.java
twcrone: `com.newrelic.agent.deps.com.google.common.collect.Maps` for creating maps

newrelic-agent/src/main/java/com/newrelic/agent/service/ServiceFactory.java
twcrone: I will look myself and ask questions as needed.  This is more of a reminder.

instrumentation/apache-log4j-1/src/main/java/com/nr/agent/instrumentation/log4j1/AgentUtil.java
jasonjkeller: Yep, good catch. I'll update it.

instrumentation/apache-log4j-2/src/main/java/com/nr/agent/instrumentation/log4j2/AgentUtil.java
jasonjkeller: Yea, they'll be the same for all log libraries. There's no straightforward way to access agent core code in instrumentation modules other than via APIs (either public NewRelic or private AgentBridge). It doesn't make much sense to me to have an API that just exposes a constant like this.

instrumentation/apache-log4j-2/src/main/java/org/apache/logging/log4j/core/layout/StringBuilderEncoder_Instrumentation.java
```diff
@@ -0,0 +1,40 @@
...
+
+@Weave(originalName = "org.apache.logging.log4j.core.layout.StringBuilderEncoder", type = MatchType.BaseClass)
+public class StringBuilderEncoder_Instrumentation {
+
+    public void encode(final StringBuilder source, final ByteBufferDestination destination) {
+        // Do nothing if application_logging.enabled: false
+        if (isApplicationLoggingEnabled()) {
+            if (isApplicationLoggingLocalDecoratingEnabled()) {
+                // Append New Relic linking metadata from agent to log message
+                appendAgentMetadata(source);
```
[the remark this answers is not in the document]
jasonjkeller: It looks like it would fall into this catch block on the library method that we're weaving this into. 

```
    @Override
    public void encode(final StringBuilder source, final ByteBufferDestination destination) {
        try {
            final Object[] threadLocalState = getThreadLocalState();
            final CharsetEncoder charsetEncoder = (CharsetEncoder) threadLocalState[0];
            final CharBuffer charBuffer = (CharBuffer) threadLocalState[1];
            final ByteBuffer byteBuffer = (ByteBuffer) threadLocalState[2];
            TextEncoderHelper.encodeText(charsetEncoder, charBuffer, byteBuffer, source, destination);
        } catch (final Exception ex) {
            logEncodeTextException(ex, source, destination);
            TextEncoderHelper.encodeTextFallBack(charset, source, destination);
        }
    }
```

instrumentation/java.logging-jdk8/src/main/java/java/util/logging/Logger_Instrumentation.java
jasonjkeller: They don't seem to be doing null checks on `record` in the library method that we're weaving this into so I think we're probably fine using it as is.

```
    /**
     * Log a LogRecord.
     * <p>
     * All the other logging methods in this class call through
     * this method to actually perform any logging.  Subclasses can
     * override this single method to capture all log activity.
     *
     * @param record the LogRecord to be published
     */
    public void log(LogRecord record) {
        if (!isLoggable(record.getLevel())) {
            return;
        }
        Filter theFilter = filter;
        if (theFilter != null && !theFilter.isLoggable(record)) {
            return;
        }

        // Post the LogRecord to all our Handlers, and then to
        // our parents' handlers, all the way up the tree.

        Logger logger = this;
        while (logger != null) {
            final Handler[] loggerHandlers = isSystemLogger
                ? logger.accessCheckedHandlers()
                : logger.getHandlers();

            for (Handler handler : loggerHandlers) {
                handler.publish(record);
            }

            final boolean useParentHdls = isSystemLogger
                ? logger.useParentHandlers
                : logger.getUseParentHandlers();

            if (!useParentHdls) {
                break;
            }

            logger = isSystemLogger ? logger.parent : logger.getParent();
        }
    }
```

instrumentation/logback-classic-1.2/src/main/java/ch/qos/logback/classic/Logger_Instrumentation.java
jasonjkeller: Probably, but I don't think it matters much in this case.

newrelic-agent/src/main/java/com/newrelic/agent/AgentLinkingMetadata.java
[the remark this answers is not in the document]
jasonjkeller: I'm not exactly sure that it needs to use `ConcurrentHashMap` but this is just maintaining the way that this API has been implemented for the past 3-ish years so I'd prefer to keep it that way unless we know for certain that there is a more optimal solution. This API isn't new, it used to live in `AgentImpl` but was refactored here.

newrelic-agent/src/main/java/com/newrelic/agent/config/ApplicationLoggingForwardingConfig.java
jasonjkeller: Disgusting. :P

newrelic-agent/src/main/java/com/newrelic/agent/service/logging/LogSenderServiceImpl.java
jasonjkeller: We use them for various purposes. This one listens for config changes and notifies the `LogSenderServiceImpl` so that it can apply config changes dynamically.

newrelic-agent/src/main/java/com/newrelic/agent/service/logging/LogSenderServiceImpl.java
jasonjkeller: I suppose so but it shouldn't matter should it as we just return the queue in the case that it is non-null?

instrumentation/apache-log4j-1/src/main/java/com/nr/agent/instrumentation/log4j1/AgentUtil.java
jasonjkeller: Fixed

instrumentation/apache-log4j-2/src/main/java/com/nr/agent/instrumentation/log4j2/AgentUtil.java
jasonjkeller: Fixed

instrumentation/logback-classic-1.2/src/main/java/com/nr/agent/instrumentation/logbackclassic12/AgentUtil.java
jasonjkeller: Fixed

newrelic-agent/src/main/java/com/newrelic/agent/service/logging/LogSenderServiceImpl.java
twcrone: sorry...I mistated...I was wondering if we got into the loop and then possibly never got out

jasonjkeller: I don't think that should be possible. If it gets into the while loop then it adds an empty queue associated with the `appName` argument and assigns that to `result`. At that point the `result` queue would no longer be `null` and it should exit the while loop.

We have the same logic in other services and it's never been an issue. https://github.com/newrelic/newrelic-java-agent/blob/3b934f3ccfbd15437a2f85ef705979148ab5a293/newrelic-agent/src/main/java/com/newrelic/agent/service/analytics/InsightsServiceImpl.java#L252-L261

tbradellis (approved):
lg

Issue #709, opened 2022-02-17
Negative Values Reported Deriving from StatsImpl and RepsponseTimeStatsImpl


tbradellis:
A pernicious issue that has alluded root cause discovery for some time due to an inability to consistently reproduce it and/or obtain live logging data has finally been reproduced.
Internally, we have information here:
https://newrelic.atlassian.net/browse/GTSE-13353

It looks like in StatsImpl and RepsonseTimeStatsImpl we need to synchronize on a few metric aggregation points. 

https://github.com/newrelic/newrelic-java-agent/blob/6be021f3f3dc8e8f992ca8eba4ba070c03e1dd3c/newrelic-agent/src/main/java/com/newrelic/agent/stats/ResponseTimeStatsImpl.java#L67

https://github.com/newrelic/newrelic-java-agent/blob/6be021f3f3dc8e8f992ca8eba4ba070c03e1dd3c/newrelic-agent/src/main/java/com/newrelic/agent/stats/ResponseTimeStatsImpl.java#L128

https://github.com/newrelic/newrelic-java-agent/blob/6be021f3f3dc8e8f992ca8eba4ba070c03e1dd3c/newrelic-agent/src/main/java/com/newrelic/agent/stats/StatsImpl.java#L57

https://github.com/newrelic/newrelic-java-agent/blob/6be021f3f3dc8e8f992ca8eba4ba070c03e1dd3c/newrelic-agent/src/main/java/com/newrelic/agent/stats/StatsImpl.java#L123

The links here should not be considered an exhaustive description of where all the problem exists.  Addition metric debug information is provided in the logs from the reporting customers application, so we should double check that there aren't other places that show the Supportability metric for negative values.  We should check to see if the DefaultTracer had any hits in those logs.

## Additional context
[[TIP]:  # ( Add any other context about the problem here. For example, relevant community posts or support tickets. )](https://newrelic.atlassian.net/browse/GTSE-13353
)

kford-newrelic:
Bumped from the `7.6.0` release; will target for the next release.

kford-newrelic:
Not able to determine root cause after extensive exploration. Deferring this issue as a potential future roadmap consideration. Moving back to `Triage`.

kford-newrelic:
Was not able to include in FY23Q1's agent release. Not selected for next quarter's agent roadmap. Will consider for a future agent release.

kford-newrelic:
[...]

PR #719, opened 2022-02-24
Tcrone/add thread safety to stats

twcrone:
### Overview
Fix thread safety issues in stats collections

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/709

### Testing
TBD

### Checks


newrelic-agent/src/main/java/com/newrelic/agent/stats/ResponseTimeStatsImpl.java
```diff
@@ -67,94 +71,116 @@ public void recordResponseTimeInNanos(long responseTime) {
...
         }
-
     }
 
     @Override
     public boolean hasData() {
-        return count > 0 || total > 0 || totalExclusive > 0;
+        boolean hasData;
+        synchronized (lock) {
+            hasData = count > 0 || total > 0 || totalExclusive > 0;
```
miniharryc, line 99: possibly fewer branches: `(count | total | totalExclusive) != 0`.  _Probably_ violates your code style, I'd expect.  

this is also cacheable. in all your mutators you could flip a stored `hasData` field and since booleans are atomic, you could mark `volatile` and avoid a lock.  Micro-optimization and I'd only consider if this were in critical piece of code

newrelic-agent/src/main/java/com/newrelic/agent/stats/StatsImpl.java
```diff
@@ -58,84 +62,100 @@ public void recordDataPoint(float value) {
...
+            total = minValue = maxValue = 0;
+            sumOfSquares = 0;
+        }
     }
 
     @Override
     public float getTotal() {
-        return total;
+        synchronized (lock) {
+            return total;
```
miniharryc, line 109: reading a `float` is  atomic.  if you mark `volatile`, you could avoid the locking here.

miniharryc (commented):
So overall, the `synchronized` + lock object is correct & minimal.  Adheres to "Principle of Least Surprise" and easiest to read and adheres to "Easy to Change" from PragProg 20th AE.

There are some possible micro-optimizations leveraging `volatile` and primitives the JVM guarantees to be atomic.  *However* these are typically extra cognitive load and only warranted if the code's really performance critical.

Unrelated to concurrency: This class doesn't seem to handle overflow conditions.

newrelic-agent/src/main/java/com/newrelic/agent/stats/ResponseTimeStatsImpl.java
twcrone: cool...yeah this is existing logic so I was avoiding changing it but might be worth it...I was just adding `synchronize` around things that were being changed

newrelic-agent/src/main/java/com/newrelic/agent/stats/StatsImpl.java
twcrone: ok, cool...played with volatile but wasn't sure.

twcrone:
> So overall, the `synchronized` + lock object is correct & minimal. Adheres to "Principle of Least Surprise" and easiest to read and adheres to "Easy to Change" from PragProg 20th AE.
> 
> There are some possible micro-optimizations leveraging `volatile` and primitives the JVM guarantees to be atomic. _However_ these are typically extra cognitive load and only warranted if the code's really performance critical.
> 
> Unrelated to concurrency: This class doesn't seem to handle overflow conditions.

Yeah, we actually are seeing that some as well I think but thought it was in part because of all the updates to state that wasn't thread-safe.  Where is a good design/pattern for handling overflow conditions?

meiao (approved):

twcrone:
This did not fix the issue and we think thread safety is handled at a higher level.

PR #779, opened 2022-03-10
Implement LogSenderServiceImplTest

twcrone:


newrelic-agent/src/test/java/com/newrelic/agent/service/logging/LogSenderServiceImplTest.java
```diff
@@ -1,7 +1,234 @@
...
+        LogSenderServiceImpl.TransactionLogs logs = new LogSenderServiceImpl.TransactionLogs(
+                AgentConfigImpl.createAgentConfig(Collections.emptyMap()));
+        when(transaction.getLogEventData()).thenReturn(logs);
+        when(transaction.getApplicationName()).thenReturn(appName);
+        when(transaction.isInProgress()).thenReturn(true);
+
+        logSenderService.recordLogEvent(ImmutableMap.of("field", "value"));
+        logSenderService.recordLogEvent(ImmutableMap.of("field2", "value2"));
+        logSenderService.recordLogEvent(ImmutableMap.of("field3", "value3"));
+        // TODO is this valid?  should an event with empty map of attributes be included?
```
twcrone, line 151: How do I create invalid log event that will be filtered out?

jasonjkeller: Validation logic is here: https://github.com/newrelic/newrelic-java-agent/blob/logging-initiative/newrelic-agent/src/main/java/com/newrelic/agent/service/logging/LogSenderServiceImpl.java#L468-L510

It's mostly validating the attributes on the LogEvent. Look for things that should be filtered out like like `null` for attribute key/value, unsupported attribute type (`Each key should be a String and each value should be a String, Number, or Boolean`), etc.

twcrone: Ok, I'll try this.  I tried an empty map of attributes and it took it.  I'll try some other combos.

twcrone: Ok, so it looks like at the point where it gets called from `AgentUtil` the attribute map will always have at least two entries.  So no way an event can currently be logged with null or empty map of attributes anyways.

newrelic-agent/src/main/java/com/newrelic/agent/service/logging/LogSenderServiceImpl.java
```diff
@@ -184,7 +184,7 @@ private void removeHarvestables() {
      */
     @Override
     public void recordLogEvent(Map<String, ?> attributes) {
-        if (logEventsDisabled()) {
+        if (logEventsDisabled() || attributes == null || attributes.isEmpty()) {
```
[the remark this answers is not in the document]
jasonjkeller: I think this is fine. You can definitely cause an NPE by calling `AgentBridge.getAgent().getLogSender().recordLogEvent(null)` but given that this is an internal API that would be on us. Regardless, it doesn't hurt to just guard against that.

I think the `attributes.isEmpty()` check is less necessary but again we can just leave it in to be extra cautious.

newrelic-agent/src/test/java/com/newrelic/agent/service/logging/LogSenderServiceImplTest.java
jasonjkeller: What about cases where key or value is invalid? I don't see that covered in the tests.

```
logSenderService.recordLogEvent(ImmutableMap.of("field", null));
logSenderService.recordLogEvent(ImmutableMap.of(null, "value"));
```

twcrone: I've done them, but it does not cause the entire event to get removed, just those attributes.

newrelic-agent/src/test/java/com/newrelic/agent/service/logging/LogSenderServiceImplTest.java
```diff
@@ -1,7 +1,247 @@
...
+
+        logSenderService.recordLogEvent(ImmutableMap.of("field", "value"));
+        logSenderService.recordLogEvent(ImmutableMap.of("field2", "value2"));
+        logSenderService.recordLogEvent(ImmutableMap.of("field3", "value3"));
+
+        // these should be filtered out
+        logSenderService.recordLogEvent(null);
+        logSenderService.recordLogEvent(Collections.emptyMap());
+
+        // these might get filtered out but don't cause the event to get filtered out
```
twcrone, line 185: These can't be created via Google ImmutableMap, but are still creatable and don't cause the even to be filtered out best I can tell.

.github/workflows/GHA-Unit-Tests.yaml
```diff
@@ -5,7 +5,7 @@ on:
     branches: [ main ] ## branches TBD
     types: [opened, synchronize, reopened, ready_for_review] ## Types TBD
   push:
-    branches: [main]
+    branches: [tcrone/implement-log-sender-service-impl-test]
```
twcrone, line 8: Ignore these...after a change, wanted to make sure ran bunch of the tests even though this is not a merge to main.

newrelic-agent/src/test/java/com/newrelic/agent/service/logging/LogSenderServiceImplTest.java
jasonjkeller: > I've done them, but it does not cause the entire event to get removed, just those attributes.

That's the expected behavior. Assertions should be on only the valid attributes being present.

twcrone: Ok, let's chat.  Worked with @meiao and @tbradellis to add these tests and fix the code.  The issue was that although the invalid entries were filtered, it could lead to logging an event with empty map of attributes which based on other code is not valid.

newrelic-agent/src/test/java/com/newrelic/agent/service/logging/LogSenderServiceImplTest.java
```diff
@@ -1,7 +1,247 @@
...
+        when(transactionData.getApplicationName()).thenReturn(appName);
+        when(transactionData.getLogEventData()).thenReturn(logs);
+
+        logSenderService.transactionListener.dispatcherTransactionFinished(transactionData, null);
+        logSenderService.harvestHarvestables();
+
+        logSenderService.harvestHarvestables();
+
+        // this fails with value of 5 cause nullValue and nullKey attribute maps are included
+        assertEquals(3, analyticsData.getEvents().size());
```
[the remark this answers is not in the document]
jasonjkeller: The LogEvents are supposed to be created but the null attributes should have been filtered out.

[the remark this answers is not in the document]
jasonjkeller: For each `LogEvent` you should be able to call `event.getMutableUserAttributes()` and make assertions that only the valid attributes are present.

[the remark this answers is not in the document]
twcrone: Ok, I think we are making sure that nulls cannot get into LogEvent now AND won't allow a LogEvent to be created with an empty map (after nulls are removed).

newrelic-agent/src/main/java/com/newrelic/agent/service/logging/LogSenderServiceImpl.java
```diff
@@ -176,6 +177,22 @@ private void removeHarvestables() {
...
+        return true;
+    }
+
+    private static Map<String, ?> filterNulls(Map<String, ?> attributes) {
+        return attributes.entrySet().stream()
+                .filter(LogSenderServiceImpl::isValidMapEntry)
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+    }
+
     /**
```
jasonjkeller, line 196: We don't need all of this extra filtering logic, it is already done in the `createValidatedEvent` method. In actual usage it will not be possible to ever have an empty attribute map because at a minimum it will always contain some agent metadata attributes. 

I don't think we need to add all of this additional code complexity and processing overhead just for a case that can only exist in a contrived unit test scenario. Using the streaming API is an extremely heavy operation when you're operating on a small map and this code needs to be as efficient as possible since it is happening for every single message that gets logged. Can we please revert back to what we had previously?

twcrone: I left it in a single commit in case it was not desired so it could be easily removed.

twcrone: Done.

PR #785, opened 2022-03-15
Refactor adding of agent linking metadata

jasonjkeller:
Moves the adding of agent linking metadata to `LogEvent`s out of individual instrumentation module and into the `LogSenderServiceImpl`.

Potential benefits:
- Might be slightly more performant as it no longer gets the full map of agent attributes and filters them down to just the ones used by `LogEvent`s. In reality this is pretty minor.
- If we ever wanted a public API like `NewRelic.getAgent().getLogSender().recordLogEvent(attributesMap)` then this is probably a better design as the linking metadata attributes (`span.id`, `entity.guid`, `hostname`, `trace.id`) are automatically added in core agent code when calling the API and the `attributesMap` only needs to contain the log attributes (`message`, `timestamp`, `log.level`). In the current design the `attributesMap` needs to have both linking metadata attributes and log attributes added before calling the API.
- Simplifies all log library instrumentation modules slightly

Cons:
- ~~Changes to `AgentImpl` are kind of gross (but can probably be cleaned up a bit)~~ This has since been refactored to address any issues.

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/AgentImpl.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/AgentImpl.java
@@ -12,8 +12,6 @@ import com.newrelic.agent.bridge.NoOpTracedMethod;
 import com.newrelic.agent.bridge.NoOpTransaction;
 import com.newrelic.agent.bridge.TracedMethod;
 import com.newrelic.agent.bridge.Transaction;
-import com.newrelic.agent.config.AgentConfig;
-import com.newrelic.agent.config.Hostname;
 import com.newrelic.agent.service.ServiceFactory;
 import com.newrelic.agent.tracers.Tracer;
 import com.newrelic.api.agent.Insights;
@@ -23,19 +21,11 @@ import com.newrelic.api.agent.MetricAggregator;
 import com.newrelic.api.agent.TraceMetadata;
 
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Level;
 
 public class AgentImpl implements com.newrelic.agent.bridge.Agent {
 
     private final Logger logger;
-
-    private static final String TRACE_ID = "trace.id";
-    private static final String SPAN_ID = "span.id";
-    private static final String HOSTNAME = "hostname";
-    private static final String ENTITY_GUID = "entity.guid";
-    private static final String ENTITY_NAME = "entity.name";
-    private static final String ENTITY_TYPE = "entity.type";
 
     public AgentImpl(Logger logger) {
         this.logger = logger;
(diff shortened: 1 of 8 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/service/logging/LogSenderServiceImpl.java
```diff
@@ -471,7 +472,9 @@ private static String mapInternString(String value) {
      * @return LogEvent instance
      */
     private static LogEvent createValidatedEvent(Map<String, ?> attributes) {
-        Map<String, Object> userAttributes = new HashMap<>(attributes.size());
+        // Initialize new userAttributes map with agent linking metadata required for LogEvents
+        Map<String, Object> userAttributes = new HashMap<>(AgentImpl.getLogEventLinkingMetadata());
```
jasonjkeller, line 476: Is there any chance that there could be a thread switch from the log library instrumentation where we call `AgentBridge.getAgent().getLogSender().recordLogEvent(logEventMap)` to this point here where we actually process that API call and retrieve the linking metadata (`AgentImpl.getLogEventLinkingMetadata()`)? 

I don't think so, but if there was that would likely cause us to add the wrong linking metadata to the `LogEvent`. That would be my main concern with this refactoring.

newrelic-agent/src/main/java/com/newrelic/agent/AgentImpl.java
```diff
@@ -157,19 +164,58 @@ public TraceMetadata getTraceMetadata() {
...
+        Map<String, String> linkingMetadata = new ConcurrentHashMap<>();
+
+        TraceMetadata traceMetadata = TraceMetadataImpl.INSTANCE;
+        String traceId = traceMetadata.getTraceId();
+        if (!traceId.isEmpty()) {
+            linkingMetadata.put(TRACE_ID, traceId);
+        }
+
+        String spanId = traceMetadata.getSpanId();
+        if (!spanId.isEmpty()) {
```
twcrone, line 206: If there is any chance of null here and you have access to a `StringUtils` in apache commons or guava perhaps, you can do `StringUtils.isBlank()` instead.

newrelic-agent/src/main/java/com/newrelic/agent/service/logging/LogSenderServiceImpl.java
twcrone: I'll admit I don't completely understand this question but I'd love to educated on this situation.  I would prefer the `logEventLinkingMetadata` be a parameter or pulled from an injected `Agent` rather than pulled from `AgentImpl` but I realize this is naive on my part and there is likely an issue with this.

twcrone (approved):

newrelic-agent/src/main/java/com/newrelic/agent/AgentImpl.java
```diff
@@ -179,7 +225,7 @@ public TraceMetadata getTraceMetadata() {
         return linkingMetadata;
     }
 
-    private String getLinkingMetaHostname(AgentConfig agentConfig) {
+    private static String getLinkingMetaHostname(AgentConfig agentConfig) {
```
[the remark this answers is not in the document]
twcrone: to be clear, I love private static utility methods that take the parameters they need for their work...

newrelic-agent/src/main/java/com/newrelic/agent/AgentImpl.java
jasonjkeller: These won't ever be `null`, the `TraceMetadata` API guarantees that an actual value or an empty `String` will always be returned.

https://github.com/newrelic/newrelic-java-agent/blob/3b934f3ccfbd15437a2f85ef705979148ab5a293/newrelic-api/src/main/java/com/newrelic/api/agent/TraceMetadata.java#L15-L35

newrelic-agent/src/main/java/com/newrelic/agent/AgentImpl.java
```diff
@@ -157,19 +164,58 @@ public TraceMetadata getTraceMetadata() {
...
+     * Gets a map of agent linking metadata minus entity.type,
+     * entity.name, and any attributes with an empty value.
+     * This subset of linking metadata is added to LogEvents.
+     *
+     * @return Filtered map of agent linking metadata
+     */
+    public static Map<String, String> getLogEventLinkingMetadata() {
+        Map<String, String> linkingMetadata = new ConcurrentHashMap<>();
+
+        TraceMetadata traceMetadata = TraceMetadataImpl.INSTANCE;
```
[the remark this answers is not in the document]
jasonjkeller: I agree, this is a pattern that makes testing extremely difficult. I've refactored and pushed commits that use dependency injection instead so unit tests should be a lot easier.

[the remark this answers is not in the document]
jasonjkeller: I've also added a new test class to prove that it can easily be tested!

Issue #711, opened 2022-02-18
Gap in Kafka metrics instrumentation


meiao:
## Description
The instrumentation module `com.newrelic.instrumentation.kafka-clients-metrics-2.0.0` instruments `KafkaConsumer#poll(Duration)`. This method is overloaded (`KafkaConsumer#poll(long)`) and both call a private method.

This means that if a customer is to call the overloaded method, no metrics will be recorded. So we should instrument the private method instead.

Also, in newer versions of Kafka (3?) the private method has changed its signature, so we will probably have to create a new instrumentation module.

## Expected Behavior
Metrics are recorded regardless of `poll` method called.

## Steps to Reproduce
Call `KafkaConsumer#poll(long)`.

averkhovtsev-xm:
The title of the ticket is a bit misleading. It is not a gap, none of the producer/consumer metrics are being sent to Newrelic.
the kafka-clients-metrics-0.10.0.0 module is not applied to 3+ version of kafka clients, the kafka-clients-metrics-2.0.0 module instruments only 1 method in KafkaConsumer which is not metrics related. 
The reason I'm bringing this up is to increase the priority (and hopefully ETA) of this ticket.

tbradellis:
@meiao The  private signature looks like it changed in 2.5.0 to use Timer instead of long.

https://github.com/apache/kafka/blob/ff4dff044a22ef0f3095c6f11c38a1a718ad7d13/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L1217

Instead of creating a new module it looks like I could just instrument both public poll methods.
`ConsumerRecords<K, V> poll(Duration timeout)`
and
`ConsumerRecords<K, V> poll(long timeoutMs)`

verifyInstrumentation passes open ended for this from 2.0.0.
  `passesOnly 'org.apache.kafka:kafka-clients:[2.0.0,)'`
  
  What do you think of this approach? It would save having to create a new module I believe.
  
  https://github.com/newrelic/newrelic-java-agent/pull/792

meiao:
It is not as simple as this.
We weren't aware that this module is supposed to be used in addition to the 0.10.0.0 instrumentation module.
[...]

PR #792, opened 2022-03-18
Kafka client metrics add second path for poll

tbradellis:
The instrumentation module com.newrelic.instrumentation.kafka-clients-metrics-2.0.0 instruments KafkaConsumer#poll(Duration). This method is overloaded (KafkaConsumer#poll(long)) and both call a private method.

@meiao The  private signature looks like it changed in 2.5.0 to use Timer instead of long.

https://github.com/apache/kafka/blob/ff4dff044a22ef0f3095c6f11c38a1a718ad7d13/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L1217

Instead of creating a new module it looks like I could just instrument both public poll methods.
`ConsumerRecords<K, V> poll(Duration timeout)`
and
`ConsumerRecords<K, V> poll(long timeoutMs)`

verifyInstrumentation passes open ended for this from 2.0.0.
  `passesOnly 'org.apache.kafka:kafka-clients:[2.0.0,)'`

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/711

### Testing
verifyInstrumentation passes open ended
  `passesOnly 'org.apache.kafka:kafka-clients:[2.0.0,)'`

### Alternatives
Instrument the private method,  But will have to split into separate modules at 2.5.0 to cover the private method signature change.

### DIFF
--- a/instrumentation/kafka-clients-metrics-2.0.0/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer_Instrumentation.java
+++ b/instrumentation/kafka-clients-metrics-2.0.0/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer_Instrumentation.java
@@ -22,6 +22,23 @@ import java.time.Duration;
 @Weave(originalName = "org.apache.kafka.clients.consumer.KafkaConsumer")
 public class KafkaConsumer_Instrumentation<K, V> {
 
+    public ConsumerRecords<K, V> poll(long timeoutMs) {
+        final ConsumerRecords<K, V> records;
+        try {
+            records = Weaver.callOriginal();
+        } catch (Exception e) {
+            // Specifically ignore WakeupExceptions because they are common in non-error use cases
+            if (!(e instanceof WakeupException)) {
+                NewRelic.noticeError(e);
+            }
+            throw e;
+        }
+
+        reportAsExternal(records);
+
+        return records;
+    }
+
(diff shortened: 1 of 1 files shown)


meiao (changes requested):
We should not instrument this method as it is already instrumented in the 0.10.0.0 module.

tbradellis:
@meiao I'm confused about what is needed here then, because if ConsumerRecords<K, V> poll(long timeoutMs)
 is instrumented in 0.10 and ConsumerRecords<K, V> poll(long timeoutMs) is instrumented in 2.0.0 there should be coverage and no need to instrument the `private ConsumerRecords<K, V> poll(final long timeoutMs, final boolean includeMetadataInTimeout)`.   
 
 I'm curious if the maybe the modules should all be refactored because of the weird overlap.

meiao:
@tbradellis yes, we are all confused.
The first description of the ticket was made on the assumption that the 2.0.0 module was a standalone module, which turned out to not be true.

Also at that point, it was also believed that the 2.0.0 module would instrument Kafka 3.0.0, since the verifyInstrumentation was open ended. Even though 2.0.0 does apply to 3.0.0, it is not enough by itself, since it is not a standalone module.

So we should revisit the internal ticket that led to open this ticket and see if their problem was in 2.x or 3.x.
If it was 3.x, there is likely no gap in 2.x, but we do need to fully instrument 3.x.

PR #798, opened 2022-03-21
Add entity.name to LogEvents. Fix tests.

jasonjkeller:
Adds `entity.name` as a `LogEvent` attribute as well as including it in the metadata that is appended to log lines. Updates related tests.

Resolves https://github.com/newrelic/newrelic-java-agent/issues/796


tbradellis (approved):
I approve this message...and this PR.

PR #815, opened 2022-04-07
Remove some repetitive fetches and null checks

twcrone:


newrelic-agent/src/main/java/com/newrelic/agent/config/UtilizationDataConfig.java
```diff
@@ -34,21 +34,24 @@ public class UtilizationDataConfig extends BaseConfig {
...
-            totalRamMibConfig = parseConfigVal(TOTAL_RAM_MIB) == null ? null : new Long(parseConfigVal(TOTAL_RAM_MIB));
+            String configValue = parseConfigVal(TOTAL_RAM_MIB);
+            totalRamMibConfig =  configValue == null ? null : new Long(configValue);
         } catch (NumberFormatException e) {
             Agent.LOG.log(Level.WARNING, "Number expected total_ram_mib configuration {0}", e);
         }
         try {
-            logicalProcessorsConfig = parseConfigVal(LOGICAL_PROCESSORS) == null ? null : new Integer(parseConfigVal(LOGICAL_PROCESSORS));
+            String configValue = parseConfigVal(LOGICAL_PROCESSORS);
+            logicalProcessorsConfig = configValue == null ? null : new Integer(configValue);
```
meiao, line 54: Since we are changing this, it is probably worth changing the `new Integer(configValue)` to `Integer.valueOf(configValue)` as the second version has built in cache for some values of integers.

There will likely not be much gain on the Long above since the cache is restricted and will likely not be ever used there.

Issue #830, opened 2022-04-26
Move Verify Instrumentation tests to GHA


meiao:
Verify Instrumentation is the last job we have running in our internal CI.

It has had some errors lately, so migrating might improve that.

PR #846, opened 2022-05-11
Running verify instrumentation tests with Java 11

meiao:
### Overview
Some libraries are built with Java 11. This cause errors when running the tests with Java 8.

It was not straightforward to use toolchain to run a single module with 11. So changing all tests to run with 11 allows all tests to pass.

### Related Github Issue
#830


twcrone (approved):
GREAT IDEA!!!  Where did you get your inspiration?

meiao:
I blame @twcrone

PR #849, opened 2022-05-13
feature(logback-classic): add ability to include stack traces when forwarding logs

fieldju:
### Overview
This PR adds the ability to for logback classic log forwarding to include stack traces.

I complied this and included it in one of my apps.

With out the feature flag, the logs have no stack trace just like it is today:
image

With the feature flag, the logs now include the stack trace:
image
image

### Related Github Issue

### Testing

### Checks

### DIFF
--- a/instrumentation/logback-classic-1.2/src/main/java/com/nr/agent/instrumentation/logbackclassic12/AgentUtil.java
+++ b/instrumentation/logback-classic-1.2/src/main/java/com/nr/agent/instrumentation/logbackclassic12/AgentUtil.java
@@ -11,6 +11,8 @@ import ch.qos.logback.classic.Level;
 import com.newrelic.agent.bridge.AgentBridge;
 import com.newrelic.api.agent.NewRelic;
 
+import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
@@ -21,6 +23,7 @@ public class AgentUtil {
     public static final int DEFAULT_NUM_OF_LOG_EVENT_ATTRIBUTES = 3;
     // Log message attributes
     public static final String MESSAGE = "message";
+    public static final String STACKTRACE = "stacktrace";
     public static final String TIMESTAMP = "timestamp";
     public static final String LOG_LEVEL = "log.level";
     public static final String UNKNOWN = "UNKNOWN";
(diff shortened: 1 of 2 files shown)


instrumentation/logback-classic-1.2/src/main/java/com/nr/agent/instrumentation/logbackclassic12/AgentUtil.java
```diff
@@ -58,6 +62,13 @@ public static void recordNewRelicLogEvent(String message, long timeStampMillis,
                 logEventMap.put(LOG_LEVEL, level);
             }
 
+            if (t != null && isApplicationForwardStackTracesEnabled()) {
```
fieldju, line 65: My co-worker and expected this to be the default behavior, we'd be in-favor of this not even being behind a flag.
I assume this was omitted to save data egress / ingress.

kford-newrelic:
@fieldju thanks for the comprehensive PR! We'll add this to our backlog for review - thanks again!

fieldju:
@kford-newrelic I saw that you added the `Represents a FY Q2 sustainability candidate` label to #866.

I'm not entirely sure what that label means but I strongly feel that this PR and #866 are needed ASAP for the agent to be a feature complete logging solution. 

I am currently forced to run a fork of this agent until these 2 PRs get merged/released.

kford-newrelic:
Not selected for next quarter's agent roadmap. Will consider for a future agent release.

@fieldju thank you for the feedback; that label essentially means we were considering this for our next quarter's (Jul-Sep) agent roadmap but unfortunately, we determined that we would need more engineering time to implement than we had available.

This is a good feature to have but we decided that we would want a consistent experience across other loggers and that we would want to communicate with other language agents, so this could potentially be implemented for those, again for consistency.

We will have to consider this for a future quarter/agent release.

fieldju:
@kford-newrelic this really unfortunate because JVM logs without stack-traces are effectively useless, which means using the logging features of the agent is not a production ready solution.

I am really confused why you are pushing the Agent as a production ready logging solution when you strip away arguably the most important peace of the logs when it comes to debugging production issues in the JVM. 

It seems like you shouldn't be turning on log-forwarding by default and discouraging functional solutions like fluentd until you have this figured out.

fieldju:
Closing this PR.

We talked about it internally and decided that it'd be best to use [Sleuth](https://spring.io/projects/spring-cloud-sleuth), [Logback](https://logback.qos.ch/) and [Micrometer/Prometheus](https://docs.spring.io/spring-metrics/docs/current/public/prometheus) in combination with the [NRI prometheus scrapper](https://github.com/newrelic/nri-prometheus) and [Fluentd](https://www.fluentd.org/) w/ the [NR Plugin](https://docs.newrelic.com/docs/logs/forward-logs/fluentd-plugin-log-forwarding/). 

This means we get the features we need and avoid vendor lock in or heavy weight agents.

We also mulled over the merits of writing a Logback appender that could send logs straight to New Relic.
You can find a couple on google if you want, but it seems like fluentd will enable us to avoid vendor lock in and have a single pattern that is x-language/framework.

fieldju:
Closing the loop on this.

We ended up making a custom logback layout and using the [New Relic Fluent Bit helm chart](https://github.com/newrelic/helm-charts/tree/f7a851795e71ee8228ed33e04a293aa21fe40448/charts/newrelic-logging)

You can see/use/copy that config [here in this gist](https://gist.github.com/fieldju/c4af82bd8663569476be2d5f575d19f8)

:point_up: this will take forward any k,v pair that is in the json blob including any MDC kv pair.

image

Issue #620, opened 2021-12-22
Exception in thread "main" java.lang.IllegalAccessError: Update to non-static final field


twcrone:
Details in https://newrelic.atlassian.net/browse/GTSE-12785

lluppani:
Hi! Can I have more details about this issue? Or access to that Jira ticket :-). Thanks!

meiao:
@lluppani we cannot give access to the Jira ticket since it may contain sensitive information.

This issue has been observed by some customers that use Scala and have recently upgraded to 7.4.3 from 7.2.0 or older releases.

lluppani:
Thats why Im asking. We are facing this issue. Can you provide a workaround?

meiao:
For most customers, downgrading to version 6.5.3 worked.
Version 6.5.3 is not vulnerable to the recent Log4J vulnerabilities.

lluppani:
Yes, we were about to test that. Thanks! How would I know when this issue is fixed in 7.4.x versions?

meiao:
We hope to have it fixed in 7.5.0.
To be notified, you can subscribe to this issue or to the releases.

lluppani:
Great! Thanks!

coding-bunny:
Still not working with the 7.5.0 release:

```
Exception in thread "main" java.lang.IllegalAccessError: Update to static final field akka.pattern.package$.MODULE$ attempted from a different method (<init>) than the initializer method <clinit>
	at akka.pattern.package$.<init>(package.scala:43)
	at akka.pattern.package$.<clinit>(package.scala)
	at akka.event.LoggingBus$class.akka$event$LoggingBus$$addLogger(Logging.scala:179)
	at akka.event.LoggingBus$$anonfun$4$$anonfun$apply$4.apply(Logging.scala:114)
	at akka.event.LoggingBus$$anonfun$4$$anonfun$apply$4.apply(Logging.scala:113)
	at scala.util.Success$$anonfun$map$1.apply(Try.scala:237)
	at scala.util.Try$.apply(Try.scala:192)
	at scala.util.Success.map(Try.scala:237)
	at akka.event.LoggingBus$$anonfun$4.apply(Logging.scala:113)
	at akka.event.LoggingBus$$anonfun$4.apply(Logging.scala:110)
	at scala.collection.TraversableLike$WithFilter$$anonfun$map$2.apply(TraversableLike.scala:683)
[...]

PR #876, opened 2022-06-07
issue 620 scala illegal access error

richard-gibson:

### Overview
This PR fixes issue #620 where an Illegal Access Exception was being thrown from apps using NR agent with scala 2.12 and Java 11. The issue is related to  https://github.com/scala/scala-dev/issues/408 where a scala trait has a `val` field and synthetic setter methods in the trait are called by implementing child class breaking a constraint set in the JMM in JDK 9+. 
The workarround for this is similar to  scala 2.13 where the trait field is marked as not final in the bytecode. 

### Testing

### Checks

[ Y ] Are your contributions backwards compatible with relevant frameworks and APIs?
[ N ] Does your code contain any breaking changes? Please describe. 
[ N ] Does your code introduce any new dependencies? Please describe.

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/instrumentation/context/InstrumentationContextManager.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/instrumentation/context/InstrumentationContextManager.java
@@ -16,6 +16,7 @@ import com.newrelic.agent.instrumentation.ClassNameFilter;
 import com.newrelic.agent.instrumentation.api.ApiImplementationUpdate;
 import com.newrelic.agent.instrumentation.classmatchers.ScalaTraitMatcher;
 import com.newrelic.agent.instrumentation.classmatchers.TraceLambdaVisitor;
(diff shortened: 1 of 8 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/instrumentation/context/InstrumentationContextManager.java
```diff
@@ -69,7 +68,7 @@ public InstrumentationContextManager(Instrumentation instrumentation) {
         this.classWeaverService = new ClassWeaverService(instrumentation);
 
         // these matchers only modify the InstrumentationContext, they don't actually transform classes
-        matchVisitors.put(new ScalaTraitMatcher(), new ScalaTraitFinalFieldTransformer());
```
richard-gibson: ScalaTraitFinalFieldTransformer moved from matchVisitors as it was marking all scala classes with final fields as modified for instrumentation.

meiao (approved):

Issue #387, opened 2021-08-26
HttpURLConnection instrumentation does not mark trace as external


meiao:
## Description
When a transaction has an HTTP call using `HttpURLConnection`, the distributed traces for the HTTP call do show as being in the same transaction and not as a remote call.

## Expected Behavior
These HTTP calls show as remotes, like calls that use Apache's HTTP client.

## Steps to Reproduce
Have 2 servers with DT enabled.
From one server, make an HTTP call using `HttpURLConnection` to the other.
Check DT.

## Your Environment
Java 16
Agent 7.2.0-SNAPSHOT
But most likely all other versions are affected.

jasonjkeller:
It looks like this is because the `httpurlconnection` instrumentation doesn't utilize the newer `NewRelic.getAgent().getTracedMethod().reportAsExternal` API. The `httpurlconnection` instrumentation is ~7 years and predates the `reportAsExternal` API and doesn't look like it was ever updated when the API was introduced.

I think refactoring the `httpurlconnection` instrumentation similar to the following commit so that it takes advantage of the newer `reportAsExternal` API should resolve the issue.

Note: There are problems with this commit that need to be fixed and tested still.
https://github.com/newrelic/newrelic-java-agent/commit/3acebd3ac5330c7fda00b213986ba8f82d3e6446

Incorrect external with existing instrumentation:
<img width="1236" alt="incorrect-external" src="https://user-images.githubusercontent.com/3496648/168929993-15afd341-0348-4cda-92ed-130f981dfe15.png">

Correct external with fixed instrumentation:
<img width="1233" alt="correct-external" src="https://user-images.githubusercontent.com/3496648/168930064-c6044b2b-2b42-47ae-adb2-6f9fc85ecfa0.png">

kford-newrelic:
Found that additional code refactoring and testing is required, will not be able to make the `7.8.0` release with a fix. Targeting the next available release instead.

contentfree:
It seems like this is now broken in a different way. 

In this screenshot, you can see that the external service is correctly recognized as such (small green sliver) but the time _inside_ the external service call is _not_ included in the external service span. (It reports as something like 40 microseconds though the request actually took nearly a minute): 
<img width="1384" alt="image" src="https://github.com/newrelic/newrelic-java-agent/assets/10958/7bb5354f-8300-4be0-b16b-7a23876940de">

Could this be related to this change?

meiao:
[...]

PR #885, opened 2022-06-16
Fix httpurlconnection inst

jasonjkeller:
Updates the `httpurlconnection` instrumentation to use the newer `reportAsExternal` and distributed tracing APIs. The external call span now correctly is marked as external in distributed traces and contains the expected `http.*` attributes.

Fixes https://github.com/newrelic/newrelic-java-agent/issues/387

The `external` AIT will pass once this PR is merged: https://github.com/newrelic/java-agent-integration-tests/pull/55

distributed-trace


jasonjkeller:
Here's an example of one instrumented service service making an external call to another instrumented service using the `HttpURLConnection` client. The second service then makes an external call to an uninstrumented domain, also using the `HttpURLConnection` client.

Screen Shot 2022-07-05 at 2 56 27 PM

PR #889, opened 2022-06-22
Fix Create-Custom-Jar.yml

meiao:
### Overview
The first commit for the Create Custom Jar workflow had a syntax error.


tbradellis (approved):
lgtm

PR #592, opened 2021-12-03
optimization: change String.replaceAll in favor of Pattern.compile

brunolellis:
### Overview
Small optimization reducing regex compilation (`String.replaceAll`) on every method call.

### Checks

[x] Are your contributions backwards compatible with relevant frameworks and APIs?
[ ] Does your code contain any breaking changes? Please describe. 
[ ] Does your code introduce any new dependencies? Please describe.

PR #933, opened 2022-07-14
Replacing regex replace with iterative char replace

zowens:

### Overview
This change replaces a regex-based replaceAll for `[.-]` with `_` to use an explicitly written method. 

Prior to this change, any call to lookup configuration values within `AgentConfigImpl` would incur the allocation cost of a `java.util.regex.Matcher` object in addition to its internal objects (notably a `int[]`). This was improved in #592 but still requires the use of regex to do the replace. The problem here is that this function can be called in the hot path of stat collection.

This PR adds an explicit function to replace the regex utilization in this path. In addition to removing any allocations when a config value does not have a dot or hyphen, this code also benchmarks to be quicker:

```
Benchmark                                        Mode  Cnt        Score       Error  Units
StringReplaceBenchmark.benchmarkUsingIteration  thrpt    5  2674595.746 ± 63511.791  ops/s
StringReplaceBenchmark.benchmarkUsingRegex      thrpt    5   749875.339 ±  3537.357  ops/s
```

<details>
<summary>Benchmark Code</summary>

```java
@State(Scope.Benchmark)
@Fork(1)
@Timeout(timeUnit = TimeUnit.SECONDS, time=20)
@Warmup(time = 10, timeUnit = TimeUnit.SECONDS, iterations = 2)
public class StringReplaceBenchmark {
    private static final List<String> testAttributeNames = List.of(
            "someconfigvalue",
            "SomeConfigurationValue",
            "provider.configuration.value",
            "newrelic.config.value-with-hyphens.and-dots",
            "index",
            "licenseKey",
            "Custom/Name/Metric.Foo",
            "Custom/Name/Metric-With-Really-Long-Char-String",
            "Custom/Name/Metric-With_Really_Long_Char_String",
            "Custom/Name/Metric-With-Really-Long-Char-String._"
    );

    private static final Pattern DOT_HYPHEN_PATTERN = Pattern.compile("[.-]");

    @Benchmark
    public void benchmarkUsingRegex(Blackhole blackhole) {
        for (final String attributeName : testAttributeNames) {
            blackhole.consume(DOT_HYPHEN_PATTERN.matcher(attributeName).replaceAll("_"));
        }
    }

    @Benchmark
    public void benchmarkUsingIteration(Blackhole blackhole) {
        for (final String attributeName : testAttributeNames) {
            blackhole.consume(replaceDotHyphenWithUnderscore1(attributeName));
        }
    }

    public static String replaceDotHyphenWithUnderscore1(String string) {
        // optimize the fast case: no dot or hyphen
        int matchingChar = findDotOrHyphen(string, 0);
        if (matchingChar == -1) {
            return string;
        }

        final StringBuilder updated = new StringBuilder(string.length());
        int start = 0;
        while (matchingChar != -1) {
            updated.append(string, start, matchingChar).append('_');
            start = matchingChar + 1;
            matchingChar = findDotOrHyphen(string, start);
        }
        // append tail
        updated.append(string, start, string.length());
        return updated.toString();
    }

    private static int findDotOrHyphen(String string, int start) {
        if (string == null) {
            return -1;
        }
        for (int i = start; i < string.length(); i++) {
            final char ch = string.charAt(i);
            if (ch == '.' || ch == '-') {
                return i;
            }
        }
        return -1;
    }
}

```
</details>

### Related Github Issue

### Testing

### Checks

[X] Are your contributions backwards compatible with relevant frameworks and APIs?

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/util/Strings.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/util/Strings.java
@@ -101,4 +101,45 @@ public class Strings {
         return className;
     }
 
+
+    /**
+     * Replaces occurrences of {@code '.'} and {@code '-'} within the string with the underscore character ({@code '_'}).
+     *
+     * This optimizes the fast path of config reading.
+     *
+     * @param string The string to replace
+     * @return String value with dots and hyphens replaced with underscore character.
+     */
+    public static String replaceDotHyphenWithUnderscore(String string) {
+        // optimize the fast case: no dot or hyphen
+        int matchingChar = findDotOrHyphen(string, 0);
+        if (matchingChar == -1) {
+            return string;
+        }
+
+        final StringBuilder updated = new StringBuilder(string.length());
+        int start = 0;
+        while (matchingChar != -1) {
+            updated.append(string, start, matchingChar).append('_');
+            start = matchingChar + 1;
+            matchingChar = findDotOrHyphen(string, start);
+        }
+        // append tail
+        updated.append(string, start, string.length());
+        return updated.toString();
+    }
+
+    private static int findDotOrHyphen(String string, int start) {
+        if (string == null) {
+            return -1;
+        }
+        for (int i = start; i < string.length(); i++) {
+            final char ch = string.charAt(i);
+            if (ch == '.' || ch == '-') {
+                return i;
+            }
+        }
+        return -1;
+    }
+
 }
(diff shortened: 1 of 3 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/util/Strings.java
```diff
@@ -101,4 +101,45 @@ public static String fixInternalClassName(String className) {
...
+        // append tail
+        updated.append(string, start, string.length());
+        return updated.toString();
+    }
+
+    private static int findDotOrHyphen(String string, int start) {
+        if (string == null) {
+            return -1;
+        }
+        for (int i = start; i < string.length(); i++) {
```
twcrone, line 136: should we pull that `string.length()` out in front of the for loop into a variable to eliminate the method call?  I know its like Java optimizes but does it optimize in Java 8?

twcrone (approved):

PR #945, opened 2022-07-25
Code level metrics

meiao:
### Overview
Adds Code Level Metrics attributes to spans created by @Trace annotations and XML, indicating the class/method that span relates to.
These attributes are code.namespace and code.function.


newrelic-agent/src/test/resources/com/newrelic/agent/config/span_events_clm.yml
```diff
@@ -0,0 +1,155 @@
+#
+# This file configures the NewRelic Agent.  Place a copy of this file
+# in RAILS_ROOT/config and edit the settings as appropriate.
```
meiao, line 3: didn't notice this.

newrelic-agent/src/test/resources/com/newrelic/agent/config/span_events_clm.yml
```diff
@@ -0,0 +1,155 @@
...
+  app_name: Unit Test
+  enable_auto_app_naming: false
+  app_renaming:
+    App1: App12
+    App2: App12
+    App3: App4
+  trim_stats: false
+
+  code_level_metrics:
+    enabled: true
```
meiao, line 20: This file was copied from span_events.yml and only these 2 lines were added.

newrelic-agent/src/main/java/com/newrelic/agent/config/AgentConfig.java
```diff
@@ -179,6 +179,11 @@ public interface AgentConfig extends com.newrelic.api.agent.Config, DataSenderCo
      */
     ApplicationLoggingConfig getApplicationLoggingConfig();
 
+    /**
+     * Get the Code Level Metrics config.
+     */
+    CodeLevelMetricsConfig getCodeLevelMetricsConfig();
```
tbradellis, line 185: I noticed sometimes we have added new config to the interface and some have only been added to the AgentConfigImpl.  I'm just curious for my own knowledge what is the determining factor on whether to include it here.  Also there is a confusing comment on the class which I've never fully understood why it was there:
```
* Adding new methods to this is no longer preferred. Please access new config using the {@link com.newrelic.api.agent.Config#getValue(String)}
 * method with the keyPath being '.' delimited hierarchy of keys, similar to the command line setting override.<br/>. 
```
Does it apply here or is this comment itself outdated?

newrelic-agent/src/test/java/com/newrelic/agent/config/CodeLevelMetricsConfigImplTest.java
```diff
@@ -0,0 +1,72 @@
...
+    public void enabled() {
+        localProps.put("code_level_metrics.enabled", true);
+        CodeLevelMetricsConfigImpl config = new CodeLevelMetricsConfigImpl(localProps);
+        assertFalse(config.isEnabled());
+    }
+
+    @Test
+    public void canConfigureViaSystemProperty() {
+        Properties properties = new Properties();
+        //default application_logging is true
```
tbradellis, line 44: I think missed a comment change here.

newrelic-agent/src/test/java/com/newrelic/agent/config/CodeLevelMetricsConfigImplTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+        CodeLevelMetricsConfigImpl config = new CodeLevelMetricsConfigImpl(Collections.emptyMap());
+        assertTrue(config.isEnabled());
+
+    }
+
+    @Test
+    public void canConfigureViaEnvironmentVariables() {
+
+        //default forwarding is false
```
tbradellis, line 61: same. I think it's a comment from logging.

tbradellis (approved):
Just saw some comments that could be cleaned up. Had one question but it was for my own understanding.

newrelic-agent/src/test/java/com/newrelic/agent/config/CodeLevelMetricsConfigImplTest.java
meiao: Right. Thanks.

newrelic-agent/src/test/java/com/newrelic/agent/config/CodeLevelMetricsConfigImplTest.java
meiao: Right. Thanks.

newrelic-agent/src/main/java/com/newrelic/agent/config/AgentConfig.java
meiao: It should be added to the interface so we don't rely on the implementation details. This is specifically useful for testing, where we could have a AgentConfigStub that implements AgentConfig. Then if all the code only uses AgentConfig, then we can easily switch to the stub for testing.

I keep forgetting about that comment.
I followed the Logging configuration, which was the last big piece of configuration we added.
I don't really like that comment, to tell you the truth. I like the fact that you call something like `agentConfig.getLoggingConfig().isEnabled`. Or maybe it is just that I really don't like passing Strings to methods (when enums would be a better choice).

PR #983, opened 2022-08-19
initial changes for jakarta webservices

tbradellis:
Jakarta webservices visitor
and javax webservices changes

### Testing
Lots of test updates.
Refactored packaging

### Checks


newrelic-agent/src/main/java/com/newrelic/agent/instrumentation/webservices/JakartaWebServiceVisitor.java
```diff
@@ -0,0 +1,153 @@
+/*
```
tbradellis, line 1: Should I consider changing the package structure here as well and put these two classes in a `javax` and `jakarta` namespace?

Also, just noticed the copyright is not right here. Probably in a few other places as well. I'll dbl check that.

functional_test/build.gradle
```diff
@@ -77,6 +77,13 @@ dependencies {
     }
     testImplementation("com.ning:async-http-client:1.9.40")
     testImplementation("javax.ws.rs:javax.ws.rs-api:2.1")
+
+    //Webservices updates for jakarta ee 9
+    implementation('jakarta.jws:jakarta.jws-api:3.0.0')
```
twcrone, line 82: I would prefer the 'implementation' dependencies all be together at top but that could be a nit.

newrelic-agent/src/main/java/com/newrelic/agent/instrumentation/webservices/JakartaWebServiceVisitor.java
twcrone: Perhaps if that is the pattern you are following else where

twcrone: but honestly if its just one class and the other class is prefixed "Javax...", may be overkill for new packages

twcrone (approved):

PR #986, opened 2022-08-22
Comment on the fix for DT headers not added to HttpUrlConnection

meiao:
### Overview
Adds a comment on a bug fix, so it is clear why the code got into that state.


tbradellis (approved):
lgtm

PR #990, opened 2022-08-27
test platform classloader path for all supported versions

tbradellis:
DO NOT MERGE!!!!!
Create pull request for ease of running tests.

Hypothesis:
            When we go through the System Classloader code path, we never add the AgentBridge...datastore jar to the class path. 
            This causes the RecordSql NoClassDefFound exception
             I believe this was done because of some conflict with java.sql classloader on certain versions(but not fully confirmed).  Regardless, because of current JDK classloaders, I believe this is a completely obsolete workaround. JDK 9+ should be fine to always use platform classloader to add our agent and agent-bridge-datastore classes to the classpath.
JDK will use the bootstrap loader, which is fine. <8  jdks are obsolete anyway.
 
    Do not go through IBM workaround and always go through the platform classloader path in order to test against current JVM test suites.
   The system classloader path was done in order to workaround some bugs. Those bugs were on IBM jdk versions 6 and 7. The agent no longer supports
    6 and 7 obsolete code and we can open up suppport for Openj9 and Semeru


tbradellis:
Consensus achieved. Will submit a pr that addresses this issue.

Issue #841, opened 2022-05-06
NR agent needlessly fails on IBM Semeru Runtimes


mstoodle:
## Is your feature request related to a problem? Please describe.

Ref discussion here: https://discuss.newrelic.com/t/feature-idea-java-new-relic-agent-doesnt-work-with-openj9-17/170671

When [IBM Semeru Runtimes](https://www.ibm.com/semeru-runtimes) Java 11 recently changed its `java.vendor` string to "IBM Corporation", the NR Java agent seems to have stopped working and apparently never supported Java 17 or any other recent Java release. When I asked for more details about why the vendor string mattered, I was pointed at this repo to investigate for myself.

I discovered that the `AttachLifecycleObserver` is disabled for any JVM that sets `java.vendor = "IBM Corporation"` which includes all Semeru Runtimes (Open and Certified Edition) as well as IBM's older IBM SDK for Java 7, 7.1, and 8. There isn't any reasoning for it in the code, that I could find. I'm suspicious that the reason this change was made is no longer relevant.

Since the NR agent worked just fine for Java 11 before we changed the java.vendor string and the Eclipse OpenJ9 JVM built into IBM Semeru Runtimes uses the same source code across all Java releases (8,11,17,18), I think the NR agent should just work. That probably includes the IBM SDK for Java 8 (which also uses the Eclipse OpenJ9 JVM).

## Feature Description
I believe all that's needed is to remove isAgentSafe() from the AttachLifecycleObserver class. I have a pull request ready to go if the project would like me to submit it.

## Describe Alternatives
A clear and concise description of any alternative solutions or features you've considered. Are there examples you could link us to?

## Additional context
IBM Semeru Runtimes are the same binaries as used to be built by the AdoptOpenJDK community as "OpenJDK with Eclipse OpenJ9". When AdoptOpenJDK moved to the Eclipse Foundation, they were prevented (by contract with Oracle) from continuing to produce the OpenJDK with OpenJ9 binaries. IBM stepped forward at the time and has produced the same binaries under the name "IBM Semeru Runtimes": built with open source components and the binaries are available at no cost for everyone to use. As chief architect for IBM Java and Semeru Runtimes, I would love to see New Relic agents working smoothly with Semeru Runtimes for all our Java releases.

## Priority
Really Want. We have customers who have been happily using NR with Semeru Runtimes 11 who are now broken and, while we have a workaround that involves an agent that mashes back the older java.vendor setting, it's a super hack.

[...]

PR #993, opened 2022-08-30
NEWRELIC-3836 remove obsolete ibm workaround - support Semeru/OpenJ9 versions

tbradellis:
Issue explained in these comments:
We never add the agent-bridge-datastore.jar in the ibmWorkaround path.  However, even with that, there is still a bug in the IBM Workaround path where we fail to use the correct branch of logic for versions 8 and above.

```
         // Check for the IBM workaround
            boolean ibmWorkaround = IBMUtils.getIbmWorkaroundDefault();
            if (System.getProperty("ibm_iv25688_workaround") != null) {
                ibmWorkaround = Boolean.parseBoolean(System.getProperty("ibm_iv25688_workaround"));
            }

            ClassLoader classLoader;
            if (ibmWorkaround) {
                // For the IBM workaround lets just use the System ClassLoader
                //if you come down this path, we never end up
                //performing addBridgeJarToClassPath(inst, AGENT_BRIDGE_DATASTORE_JAR_NAME); hence NoClassDefFound for Record.Sql
                classLoader = ClassLoader.getSystemClassLoader();
            } else {
                ClassLoader agentClassLoaderParent = getPlatformClassLoaderOrNull();

                // Create a new URLClassLoader instance for the agent to use instead of relying on the System ClassLoader
                URL[] codeSource;
                if (isJavaSqlLoadedOnPlatformClassLoader(javaVersion)) {
                    //Semeru and other OpenJ9s for JDK 9+ are not getting to this path, and they should be.
                    //the logic in getIbmWorkaroundDefault() is faulty and causes anything above JDK 7 to not return false
                    URL url = BootstrapLoader.getDatastoreJarURL();
                    codeSource = new URL[] { getAgentJarUrl(), url };
                } else {
                    //If you get here, then the BootstrapLoader will have already
                    // performed addBridgeJarToClassPath(inst, AGENT_BRIDGE_DATASTORE_JAR_NAME); so, no problem (i.e. older JDKs did this).  
                    //IBM JDK 8 versions should have ended up here too, but did not because of the bug in  getIbmWorkaroundDefault() is faulty 
                  
                    codeSource = new URL[] { getAgentJarUrl() };
                }

                classLoader = new JVMAgentClassLoader(codeSource, agentClassLoaderParent);

                redefineJavaBaseModule(inst, classLoader);
                addReadUnnamedModuleToHttpModule(inst, agentClassLoaderParent);
            }
            
            
```

You can see in the if condition here that we only really check 1.7, so IBM JVMs 8 and above will always be true.  8 is above srNum 4.  Also, 9+ uses platform classloader, so this is all just uneccesary.  Had this check been correct, this issue probably wouldn't have come up. Still, best to rip this stuff out because it just is pointless complexity in the BootstrapAgent.

```
    public static boolean getIbmWorkaroundDefault() {
        try {
            if (isIbmJVM()) {
                String jvmVersion = System.getProperty("java.specification.version", "");
                int srNum = getIbmSRNumber();
                //We only set to false if 1.7 and above srNum
                //that means anything greater than jdk 7 will go through the true branch.
                if ("1.7".equals(jvmVersion) && srNum >= 4) {
                    // IBM Ticket says 7.SR3 is fixed, but no jvm can be found to test against
                    return false;
                }
                return true;
            }
            return false;
        } catch (Exception e) {
            return true;
        }
    }
```

This entire logic is obsolete anyway because the agent no longer supports Java 6 and 7. 

**Tests:**
We need to add OpenJ9 and/or Semeru runtimes to our AITs

### DIFF
--- a/functional_test/src/test/java/com/newrelic/agent/instrumentation/ClassLoaderTest.java
+++ b/functional_test/src/test/java/com/newrelic/agent/instrumentation/ClassLoaderTest.java
@@ -10,6 +10,7 @@ package com.newrelic.agent.instrumentation;
 import com.newrelic.agent.Transaction;
 import com.newrelic.api.agent.NewRelic;
 import com.newrelic.api.agent.Trace;
+import com.newrelic.test.marker.IBMJ9IncompatibleTest;
 import com.newrelic.test.marker.Java17IncompatibleTest;
 import com.newrelic.test.marker.Java19IncompatibleTest;
 import org.junit.Assert;
@@ -67,7 +68,7 @@ public class ClassLoaderTest {
 
     // Java 12 no longer lets us access the declared field
     @Test
-    @Category({ Java17IncompatibleTest.class, Java19IncompatibleTest.class })
+    @Category({ IBMJ9IncompatibleTest.class, Java17IncompatibleTest.class, Java19IncompatibleTest.class })
     public void testSetSystemClassLoader() throws Exception {
 
         final ClassLoader systemClassLoader = ClassLoader.getSystemClassLoader();
--- a/functional_test/src/test/java/test/newrelic/test/agent/spans/SpanErrorsTest.java
+++ b/functional_test/src/test/java/test/newrelic/test/agent/spans/SpanErrorsTest.java
@@ -11,9 +11,12 @@ import com.google.common.collect.ImmutableMap;
 import com.newrelic.agent.interfaces.SamplingPriorityQueue;
 import com.newrelic.agent.model.SpanEvent;
 import com.newrelic.agent.service.ServiceFactory;
+import com.newrelic.test.marker.IBMJ9IncompatibleTest;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.experimental.categories.Categories;
+import org.junit.experimental.categories.Category;
 import test.newrelic.EnvironmentHolderSettingsGenerator;
 import test.newrelic.test.agent.EnvironmentHolder;
 
@@ -98,6 +101,7 @@ public class SpanErrorsTest {
         ));
     }
 
+    @Category(IBMJ9IncompatibleTest.class)
     @Test
     public void testNoticedErrorOverridesThrownError() {
         runTransaction(new SpanErrorFlow.NoticeAndThrow());
--- a/instrumentation/cassandra-datastax-4.0.0/src/test/java/com/nr/agent/instrumentation/cassandra/CassandraInstrumented.java
+++ b/instrumentation/cassandra-datastax-4.0.0/src/test/java/com/nr/agent/instrumentation/cassandra/CassandraInstrumented.java
@@ -14,6 +14,7 @@ import com.newrelic.agent.introspec.InstrumentationTestRunner;
 import com.newrelic.agent.introspec.Introspector;
 import com.newrelic.agent.introspec.TraceSegment;
 import com.newrelic.agent.introspec.TransactionTrace;
+import com.newrelic.test.marker.IBMJ9IncompatibleTest;
 import com.newrelic.test.marker.Java11IncompatibleTest;
 import com.newrelic.test.marker.Java17IncompatibleTest;
 import com.newrelic.test.marker.Java19IncompatibleTest;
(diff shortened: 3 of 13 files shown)


tbradellis:
relates to:
https://github.com/newrelic/newrelic-java-agent/issues/841

tbradellis:
Related to ait work needed for these runtimes:
https://github.com/actions/setup-java/issues/279
https://github.com/actions/setup-java/pull/289

functional_test/src/test/java/test/newrelic/test/agent/spans/SpanErrorsTest.java
```diff
@@ -98,6 +101,7 @@ public void testNoticedStringDoesNotBubble() {
         ));
     }
 
+    @Category(IBMJ9IncompatibleTest.class)
     @Test
     public void testNoticedErrorOverridesThrownError() {
         runTransaction(new SpanErrorFlow.NoticeAndThrow());
```
tbradellis, line 107: IBMj9 throws this as `divide by zero` and our tests look for the String `/ by zero`.

functional_test/src/test/java/com/newrelic/agent/instrumentation/ClassLoaderTest.java
```diff
@@ -72,7 +73,7 @@ public void testMissingNewRelicClass() throws ClassNotFoundException {
 
     // Java 12 no longer lets us access the declared field
     @Test
-    @Category({ Java12IncompatibleTest.class, Java13IncompatibleTest.class, Java14IncompatibleTest.class,
+    @Category({ IBMJ9IncompatibleTest.class, Java12IncompatibleTest.class, Java13IncompatibleTest.class, Java14IncompatibleTest.class,
             Java15IncompatibleTest.class, Java16IncompatibleTest.class, Java17IncompatibleTest.class, Java18IncompatibleTest.class })
     public void testSetSystemClassLoader() throws Exception {
```
tbradellis, line 79: The `Classloader` class used in the J9 implementation does not have the field `scl`.

instrumentation/cassandra-datastax-4.0.0/src/test/java/com/nr/agent/instrumentation/cassandra/CassandraInstrumented.java
```diff
@@ -39,7 +40,9 @@
 import static org.junit.Assert.assertNotNull;
 
 // Issue when running cassandra unit on Java 9+ - https://github.com/jsevellec/cassandra-unit/issues/249
-@Category({Java9IncompatibleTest.class, Java10IncompatibleTest.class, Java11IncompatibleTest.class, Java12IncompatibleTest.class, Java13IncompatibleTest.class, Java14IncompatibleTest.class,
+// Many OpenJ9 and Cassandra issues e.g. :https://github.com/jsevellec/cassandra-unit/issues/295
```
tbradellis, line 43: From what I can tell, anyone attempting to use cassandra with OpenJ9 is already having issues.

tbradellis:
lasp_strip_exceptions.py test needs to be reworked specifically for OpenJ9. The return value for ArithmeticException is different than openJDK.  `/ by zero` vs `divide by zero`

This will need to be done when the AIT work for IBM Semeru/OpenJ9 is implemented.

tbradellis:
Our current gradle version is blocking us from testing with Semeru.  Long story, but we had upgrading our gradle version on the road map anyway.
This will get us to 7.5.1.  Gradle will work well with Semeru builds starting in 7.4
https://github.com/newrelic/newrelic-java-agent/pull/997

tbradellis:
errors.py:
```
--Warning: Unexpected Scoped Metrics--
Did not expect <Error ts:1662248600895
 path:Unknown
 message:divide by zero
 exception_type:java.lang.ArithmeticException
 params:{'userAttributes': {'str': 'dude'}, 'intrinsics': {'error.expected': True}, 'agentAttributes': {'request.uri': ''}, 'stack_trace': ['\tcom.nr.test.servlet.NewRelicApiServlet$1.run(NewRelicApiServlet.java:26)', '\tjava.lang.Thread.run(Thread.java:827)']}>
--/Warning: Unexpected Scoped Metrics--
```

This is similar to the SpanErrorsTest failure:
https://github.com/newrelic/newrelic-java-agent/pull/993#discussion_r960278554

It looks like OpenJ9 uses the string `divide by zero` vs `/ by zero` which is the string our tests compare against.

```
        self.assertIterAttributes(
            self.test_env.collector.get_errors(),
            {
                'path'           : 'Unknown',
                'message'        : '/ by zero',
                'exception_type' : 'java.lang.ArithmeticException'
            })


```

Various in in-tracgin.py

```
runtest.sh: running test tests/java/functionality/basic_features/inf_tracing_test.py with args 
/home/runner/work/newrelic-java-agent/newrelic-java-agent/agent-integration-tests/src/newrelic/agenttest/testenv.py:295: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if len(self.exceptions) is not 0:
/home/runner/work/newrelic-java-agent/newrelic-java-agent/agent-integration-tests/src/newrelic/agenttest/testenv.py:323: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if os.environ.get('TEST_FAIL_EXCEPTIONS', False) and len(self.unique_exceptions) is not 0:

```
Others as well:
https://github.com/newrelic/newrelic-java-agent/runs/8172652668?check_suite_focus=true


Unit tests for Java 8 and 11 OpenJ9 do not finish. It is unclear why.
https://github.com/newrelic/newrelic-java-agent/runs/8172302024?check_suite_focus=true

PR #997, opened 2022-09-02
Update gradle version to 7.5.1

tbradellis:
This upgrades us to 7.5.1 so we can do more of the new things. 
Removes several deprecated and removed features.  Replaces them with the new things.
Also, there was a workaround for compileScala jobs that is no longer needed as of Gradle Version 7.1.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id 'com.github.prokod.gradle-crossbuild' version '0.12.1' apply false
+    id 'com.github.prokod.gradle-crossbuild' version '0.14.1' apply false
     id "org.jetbrains.kotlin.jvm" version "1.6.20" apply false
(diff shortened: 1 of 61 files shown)


tbradellis:
Updating zinc to 1.7.1 solved almost all of the scalaCompile issues.
For some reason:
newrelic-scala-monix-api is still getting 
`
## Exception when compiling 2 sources to /Users/bellis/dev/maintainer/newrelic-java-agent/instrumentation/newrelic-scala-monix-api/build/classes/scala/main
java.lang.NoClassDefFoundError: scala/jdk/javaapi/CollectionConverters

`
I'm not sure why. The other similar builds no longer give NoClassDefFoundError for CollectionConverters.

twcrone (approved):

PR #999, opened 2022-09-02
Fix instrumentation verifications

twcrone:
- R2DBC Postgres exclude starting at 0.9.2.RELEASE
- Rename r2dbc-postresql module to ...-0.9.0
- jersey-3 to exclude starting at 3.1.0-M7
- jetty-11 to exclude starting at 12.0.0.alpha


meiao (approved):

jurriaan:
Why is the exclude starting at 0.9.2.RELEASE? We are currently on  r2dbc-postgresql > 0.9.1.RELEASE, and have no R2DBC instrumentation available.

meiao:
There were changes in the library that made it incompatible with the instrumentation we have.
So we need to write a new instrumentation module for version 0.9.2 and newer.

koenpunt:
@meiao is there a timeline on when this new instrumentation module will be written?

meiao:
No timeline. But I added a note in our internal ticket that also has other customers interested.

koenpunt:
Any update on this?

PR #1002, opened 2022-09-07
separate javax.xml - sun httpserver modules

tbradellis:
It's possible that using the verifyInstrumentation with these instrumentation packages that were lumped into one module is causing a problem.

This change separates `com.sun.net.httpserver` and `javax.xml.ws.spi.http` into separate modules.  Previous to the jakarta changes we were not using verifyInstrumentation for this module.  Adding it produces some conflicting test results. Sometimes verification passes, we could not get it to pass on GHA. Hence, this attempt to separate the logic.


meiao:
Running verifyInstrumentation on this branch:
https://github.com/newrelic/newrelic-java-agent/actions/runs/3010624310

meiao:
Verify instrumentation run:
https://github.com/newrelic/newrelic-java-agent/actions/runs/3011110178

PR #1003, opened 2022-09-07
Fix verification to use implementation

twcrone:
This should be merged AFTER the verification plugin is updated an published to 3.3

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -52,6 +52,12 @@ include 'functional_test:weave_test'
 include 'functional_test:weave_test_impl'
 
 // JDK 9+ module system tests
+if (JavaVersion.current().isJava9Compatible()) {
+    include 'module_test_9'
+}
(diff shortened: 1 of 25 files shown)


twcrone:
Waiting on maven repo to get new verify plugin

tbradellis (approved):
lgtm

settings.gradle
```diff
@@ -52,12 +52,6 @@ include 'functional_test:weave_test'
 include 'functional_test:weave_test_impl'
 
 // JDK 9+ module system tests
-if (JavaVersion.current().isJava9Compatible()) {
```
meiao, line 55: These only caused failures locally, right?
I think we should leave these for GHA for now.

twcrone: um...not failures on GHA yet WITH these changes

meiao (approved):

PR #1012, opened 2022-09-19
Add support for blaze server 0.23

magJ:
### Overview
Implements support for blaze server 0.23, which uses cats effect 3.
Additionally modifies the existing cats effect 3 support to work with different constructors, as newer versions are different.

### Checks


kford-newrelic:
@magJ thanks for the suggestion - support for Blaze 0.23 is being worked on now and should be in our next release, targeted for late October

magJ:
> @magJ thanks for the suggestion - support for Blaze 0.23 is being worked on now and should be in our next release, targeted for late October

Great, I'll close this one then.

PR #1014, opened 2022-09-21
NEWRELIC-658 check expected error config for noticeError api

tbradellis:
### Overview
The changes here force the noticeError API to check the expected error configuration when using APIs that do not explicitly set the expected boolean. 
Previously, we just passed false all the way through and the expected boolean only gets set if the caller uses the API that includes the boolean.

### Related Github Issue

### Testing
Tested with local apps and config.
This will add new tests.

### Checks

[ yes] Are your contributions backwards compatible with relevant frameworks and APIs?
[ no] Does your code contain any breaking changes? Please describe. 
[ no] Does your code introduce any new dependencies? Please describe.

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
+++ b/newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
@@ -15,6 +15,7 @@ import com.newrelic.agent.attributes.CustomAttributeSender;
 import com.newrelic.agent.bridge.AgentBridge;
 import com.newrelic.agent.bridge.PublicApi;
 import com.newrelic.agent.config.ConfigConstant;
+import com.newrelic.agent.config.ExpectedErrorConfig;
 import com.newrelic.agent.dispatchers.Dispatcher;
 import com.newrelic.agent.service.ServiceFactory;
 import com.newrelic.agent.transaction.TransactionNamingPolicy;
@@ -24,6 +25,7 @@ import java.text.MessageFormat;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
@@ -56,7 +58,7 @@ public class NewRelicApiImplementation implements PublicApi {
      */
     @Override
     public void noticeError(Throwable throwable, Map<String, ?> params) {
-        noticeError(throwable, params, false);
+        noticeError(throwable, params, isExpectedErrorConfigured(throwable));
     }
 
     /**
@@ -68,14 +70,16 @@ public class NewRelicApiImplementation implements PublicApi {
     @Override
     public void noticeError(Throwable throwable) {
         Map<String, String> params = Collections.emptyMap();
-        noticeError(throwable, params, false);
+        noticeError(throwable, params, isExpectedErrorConfigured(throwable));
+
     }
 
     /**
      * Notice an error and report it to New Relic. If this method is called within a transaction, the error message will
      * be reported with the transaction when it finishes. If it is invoked outside of a transaction, a traced error will
      * be created and reported to New Relic.
-     *
+     * Expecting Errors via configuration is not supported from the APIs that do not accept a Throwable as an attribute.
+     * To mark an Error that accepts a String as expected use the API that accepts a boolean.
      * @param message the error message
      * @param params  Custom parameters to include in the traced error. May be null. Map is copied to avoid side effects.
      */
@@ -86,7 +90,8 @@ public class NewRelicApiImplementation implements PublicApi {
 
     /**
      * Report an error to New Relic.
-     *
+     * Expecting Errors via configuration is not supported from the APIs that do not accept a Throwable as an attribute.
+     * To mark an Error that accepts a String as expected use the API that accepts a boolean.
      * @param message the error message
      * @see #noticeError(String, Map)
      */
@@ -97,11 +102,19 @@ public class NewRelicApiImplementation implements PublicApi {
     }
 
     @Override
+    public void noticeError(Throwable throwable, boolean expected) {
+        Map<String, String> params = Collections.emptyMap();
+        noticeError(throwable, params, expected);
+    }
+
     public void noticeError(Throwable throwable, Map<String, ?> params, boolean expected) {
         try {
             ServiceFactory.getRPMService().getErrorService().reportException(throwable, filterErrorAtts(params, attributeSender), expected);
 
             MetricNames.recordApiSupportabilityMetric(MetricNames.SUPPORTABILITY_API_NOTICE_ERROR);
+            //SUPPORTABILITY_API_EXPECTED_ERROR_API_THROWABLE metric is intended to be recorded independent of whether
+            //the expected error was defined in config or via the actual api.  Read simply as the call came through the noticeError API (SUPPORTABILITY_API_NOTICE_ERROR)
+            //and it was expected.
             if (expected) {
                 MetricNames.recordApiSupportabilityMetric(MetricNames.SUPPORTABILITY_API_EXPECTED_ERROR_API_THROWABLE);
             }
@@ -117,15 +130,10 @@ public class NewRelicApiImplementation implements PublicApi {
     }
 
     @Override
-    public void noticeError(Throwable throwable, boolean expected) {
-        Map<String, String> params = Collections.emptyMap();
-        noticeError(throwable, params, expected);
-    }
-
-    @Override
     public void noticeError(String message, Map<String, ?> params, boolean expected) {
         try {
             ServiceFactory.getRPMService().getErrorService().reportError(message, filterErrorAtts(params, attributeSender), expected);
+
             MetricNames.recordApiSupportabilityMetric(MetricNames.SUPPORTABILITY_API_NOTICE_ERROR);
             if (expected) {
                 MetricNames.recordApiSupportabilityMetric(MetricNames.SUPPORTABILITY_API_EXPECTED_ERROR_API_MESSAGE);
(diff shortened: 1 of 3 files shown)


newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
```diff
@@ -574,6 +579,20 @@ public void setInstanceName(String instanceName) {
         MetricNames.recordApiSupportabilityMetric(MetricNames.SUPPORTABILITY_API_SET_INSTANCE_NAME);
     }
 
+    private static boolean isExpectedErrorConfigured(Throwable t) {
```
tbradellis, line 582: The reason I'm doing this check here is because it is too late to check the config from anywhere in `ErrorServiceImpl`.

The agent spec requires that the `noticeError(String, boolean expected)` and similar APIs take precedence over config (where as the methods without the boolean need to check config).  If a user invokes the API and sets the boolean as false and it is configured to be expected via config, this should still report the error.

So, if we want to adhere to the spec for the `noticeError(String msg)` case, then we have to check config before calling the `noticeError(String msg, boolean expected)` method. Otherwise, there is no way to distinguish where the boolean was set.  OR we would need to do some significant refactoring and send the method signatures without the expected boolean down a different path.  I don't think it's really worth it.




expected-errors-api

newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
```diff
@@ -127,6 +131,7 @@ public void noticeError(String message, Map<String, ?> params, boolean expected)
         try {
             ServiceFactory.getRPMService().getErrorService().reportError(message, filterErrorAtts(params, attributeSender), expected);
             MetricNames.recordApiSupportabilityMetric(MetricNames.SUPPORTABILITY_API_NOTICE_ERROR);
+
             if (expected) {
                 MetricNames.recordApiSupportabilityMetric(MetricNames.SUPPORTABILITY_API_EXPECTED_ERROR_API_MESSAGE);
             }
```
tbradellis, line 137: There also aren't good options for this supportability metric either.  If this was set by config, then this metric isn't really accurate. I don't really see a great way to clean that up.

newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
```diff
@@ -574,6 +579,20 @@ public void setInstanceName(String instanceName) {
...
     }
 
+    private static boolean isExpectedErrorConfigured(Throwable t) {
+        boolean expected = true;
+        String app_name = NewRelic.getAgent().getConfig().getValue("app_name");
+        Set<ExpectedErrorConfig> expectedErrors = ServiceFactory.getConfigService().getErrorCollectorConfig(app_name).getExpectedErrors();
+        for (ExpectedErrorConfig error : expectedErrors) {
+            if (error.getErrorClass().equals(t.getClass().getName()) && error.getErrorMessage() == null ||
+                    error.getErrorClass().equals(t.getClass().getName()) && error.getErrorMessage().equals(t.getMessage())) {
+                return expected;
```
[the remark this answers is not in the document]
tbradellis: Our logic actually already handles this. The Set<ExpectedErrorConfig> has an entry for every key:value configured in the yml (and the case of message: val, val, val, there is a message:val object for each val.  This also means that the conditional logic needs to check for the case of key:null and key:msg separately.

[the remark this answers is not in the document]
tbradellis: image

newrelic-agent/src/test/java/com/newrelic/api/agent/NewRelicApiImplementationTest.java
```diff
@@ -77,6 +78,22 @@ public void noticeErrorShouldAcceptNumberAndBooleanAttributeType() {
...
+        Map<String, Object> expectedValues = new HashMap<>();
+        String msg = "this should not be expected";
+
+        expectedValues.put("java.lang.Throwable", msg);
+        NewRelicApiImplementation target = new NewRelicApiImplementation();
+        target.noticeError(msg);
+        Mockito.verify(errorService).reportError(msg, Collections.EMPTY_MAP, false);
+
+    }
+
```
tbradellis, line 96: Turns out all of this is well tested in the APITest class.  I think I'm going to to end up dropping this.

newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
```diff
@@ -56,7 +62,7 @@ public NewRelicApiImplementation() {
      */
     @Override
     public void noticeError(Throwable throwable, Map<String, ?> params) {
-        noticeError(throwable, params, false);
+        noticeError(throwable, params, isExpectedErrorConfigured(throwable.getClass().getName(), throwable.getMessage()));
```
tbradellis, line 65: I will need to add this check here as well:
`noticeError(throwable, params, throwable != null && isExpectedErrorConfigured(throwable.getClass().getName(), throwable.getMessage()));`

newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
```diff
@@ -36,6 +38,10 @@
 public class NewRelicApiImplementation implements PublicApi {
     private final AttributeSender attributeSender;
 
+    private final String EXPECTED_MESSAGE_CONFIG_PLACEHOLDER = "message";
```
tbradellis, line 41: Would this be better as `static`?

newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
```diff
@@ -36,6 +38,10 @@
...
+
+    private boolean setExpectedByConfig = false;
```
tbradellis, line 43: I think this won't ultimately work. Need a different approach.

newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
```diff
@@ -56,7 +58,9 @@ public NewRelicApiImplementation() {
      */
     @Override
     public void noticeError(Throwable throwable, Map<String, ?> params) {
-        noticeError(throwable, params, false);
+        noticeError(throwable, params, throwable != null &&
+                        isExpectedErrorConfigured(throwable.getClass().getName(), throwable.getMessage()),
```
tbradellis, line 62: Maybe now I can just pass the throwable here and do the messy logic inside `isExpectedErrorConfigured()`
Would that make more sense or is this okay?

newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
```diff
@@ -147,6 +163,21 @@ public void noticeError(String message, boolean expected) {
...
     }
 
+    private boolean isExpectedErrorConfigured(String expectedConfigName, String message) {
+        boolean expected = true;
+        String app_name = NewRelic.getAgent().getConfig().getValue("app_name");
+        Set<ExpectedErrorConfig> expectedErrors = ServiceFactory.getConfigService().getErrorCollectorConfig(app_name).getExpectedErrors();
+        for (ExpectedErrorConfig errorConfig : expectedErrors) {
+            if ((errorConfig.getErrorClass().equals(expectedConfigName) && errorConfig.getErrorMessage() == null) ||
+                    (errorConfig.getErrorMessage() != null) &&
+                            (errorConfig.getErrorClass().equals(expectedConfigName) && errorConfig.getErrorMessage().equals(message))) {
```
jasonjkeller, line 173: You can store the values from `errorConfig.getErrorClass()` and `errorConfig.getErrorMessage()` in variables before evaluating the if statement to avoid calling the getter methods repeatedly.

newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
```diff
@@ -147,6 +163,21 @@ public void noticeError(String message, boolean expected) {
         noticeError(message, params, expected);
     }
 
+    private boolean isExpectedErrorConfigured(String expectedConfigName, String message) {
+        boolean expected = true;
```
jasonjkeller, line 167: There's really no reason for this `expected` variable to exist as the value is never modified. You could just directly return `true` or `false` in place of it.

newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
```diff
@@ -98,11 +108,23 @@ public void noticeError(String message) {
...
+
+    }
+
+    @Override
+    public void noticeError(Throwable throwable, boolean expected) {
+        Map<String, String> params = Collections.emptyMap();
+        noticeError(throwable, params, expected, true);
+    }
+
+    private void noticeError(Throwable throwable, Map<String, ?> params, boolean expected, boolean expectedByApi) {
```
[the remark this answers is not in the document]
tbradellis: I looked at that, but I don't see how we can manage the state between where the error was expected between these api calls. 
The other option seemed to be to add some synchronization, which seems a lot worse.

newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
```diff
@@ -98,11 +108,23 @@ public void noticeError(String message) {
...
         try {
             ServiceFactory.getRPMService().getErrorService().reportException(throwable, filterErrorAtts(params, attributeSender), expected);
 
             MetricNames.recordApiSupportabilityMetric(MetricNames.SUPPORTABILITY_API_NOTICE_ERROR);
-            if (expected) {
+
+            if (expected && expectedByApi) {
```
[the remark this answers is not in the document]
tbradellis: Ah, this should probably just be 
if (expectedByApi)

newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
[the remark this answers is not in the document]
tbradellis: Another option I guess was to completely decouple the code paths for when noticeError(throwable, expected) and noticeError(throwable)is called.  But that would be a lot of code duplication essentially.

newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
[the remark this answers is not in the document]
jasonjkeller: That misses my point, I think. What happens if only `expected` evaluates to `true`? That is a case where a metric should be recorded, no?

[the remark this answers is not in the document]
tbradellis: This metric should only be recorded if it is marked expected by the API. In my mind it is part of the issue that is reported.  Previously we were over reporting expected by API calls and not reporting consulting config (though in fact because of how we report supportability metrics for config it was getting recorded as such.

jasonjkeller (approved):
Thanks for tackling this. LGTM!

PR #866, opened 2022-05-27
NR-23881 - Give users possibility to add MDC to logging

StanlieK:
# Include MDC data into logs forwarded to NR from log4j-2 and logback

### Overview
This PR will add support for including MDC data in logs forwarded to NR from log4j-2 and logback. It maps the MDC map into log events prefixed by `mdc.`. E.g. if you have the MDC key `example`, it will have the key `mdc.example` in New Relic.

The whole feature is **disabled** by default and can be enabled by setting:
* `application_logging.forwarding.include_mdc.enabled` to `true`
* `NEW_RELIC_APPLICATION_LOGGING_FORWARDING_INCLUDE_MDC_ENABLED` to `true`

**Important:** This feature is not available for log4j 1.x as the agent does not support forwarding of the logs at all.

#### Example - Logback with disabled INCLUDE_MDC feature
Screenshot 2022-05-27 at 15 55 04

#### Example - Logback with enabled INCLUDE_MDC feature
Screenshot 2022-05-27 at 15 48 58

#### Example - Log4j with disabled INCLUDE_MDC feature
Screenshot 2022-05-27 at 15 57 54

#### Example - Log4j with enabled INCLUDE_MDC feature
Screenshot 2022-05-27 at 15 58 49

### Related Github Issue
The feature has been requested in [this issue #801](https://github.com/newrelic/newrelic-java-agent/issues/801).

### Testing
The agent includes a suite of tests that should be used to
verify your changes don't break existing functionality. These tests will run with
Github Actions when a pull request is made. More details on running the tests locally can be found
[here](https://github.com/newrelic/newrelic-java-agent/blob/main/CONTRIBUTING.md).

### Checks

[x] Are your contributions backwards compatible with relevant frameworks and APIs?
[ ] Does your code contain any breaking changes? Please describe. 
[ ] Does your code introduce any new dependencies? Please describe.

PR #1016, opened 2022-09-21
Internal PR for testing #866

meiao:
This is an internal PR to run the tests for the external PR #866


instrumentation/apache-log4j-2/src/main/java/com/nr/agent/instrumentation/log4j2/AgentUtil.java
```diff
@@ -59,55 +33,62 @@ public static void recordNewRelicLogEvent(LogEvent event) {
             Throwable throwable = event.getThrown();
 
             if (shouldCreateLogEvent(message, throwable)) {
-                HashMap<String, Object> logEventMap = new HashMap<>(DEFAULT_NUM_OF_LOG_EVENT_ATTRIBUTES);
+                Map<LogAttributeKey, Object> logEventMap = new HashMap<>(DEFAULT_NUM_OF_LOG_EVENT_ATTRIBUTES);
                 if (message != null) {
                     String formattedMessage = message.getFormattedMessage();
                     if (formattedMessage != null && !formattedMessage.isEmpty()) {
-                        logEventMap.put(MESSAGE, formattedMessage);
+                        logEventMap.put(AppLoggingUtils.MESSAGE, formattedMessage);
```
jasonjkeller, line 40: Do you think it's worth adding a static import of `AppLoggingUtils` to avoid having to use the `AppLoggingUtils.` prefix all over?

instrumentation/apache-log4j-2/src/main/java/com/nr/agent/instrumentation/log4j2/AgentUtil.java
```diff
@@ -59,55 +33,62 @@ public static void recordNewRelicLogEvent(LogEvent event) {
...
                     }
                 }
-                logEventMap.put(TIMESTAMP, event.getTimeMillis());
+                logEventMap.put(AppLoggingUtils.TIMESTAMP, event.getTimeMillis());
+
+                if (AppLoggingUtils.isAppLoggingContextDataEnabled() && event.getContextData() != null) {
+                    event.getContextData().forEach((key, value) -> {
```
jasonjkeller, line 46: Is this implementation of `forEach` just a simple for loop under the hood?

If not, isn't this much less performant that using a simple loop? For the extra allocation reasons that you pointed out previously? Also based on some experimentation that I've done I found that the execution time is much longer when using streaming operations on small data sets as compared to a simple loop. Unless you expect that there may be many thousands of context attributes I'm not sure of the benefit of streaming.

instrumentation/spring-aop-2/build.gradle
```diff
@@ -4,9 +4,6 @@ dependencies {
     implementation("org.springframework:spring-aop:3.2.+")
     //     implementation("org.springframework:spring:2.0.3") // First version to have MethodMatchers.matches()
 
-    testImplementation("org.powermock:powermock-module-junit4-rule-agent:1.6.1")
-    testImplementation("org.powermock:powermock-module-junit4:1.6.1")
-    testImplementation("org.powermock:powermock-api-mockito:1.6.1")
```
jasonjkeller, line 9: How is this related to this PR?

newrelic-agent/src/main/java/com/newrelic/agent/attributes/AttributesNode.java
```diff
@@ -79,7 +79,7 @@ protected Boolean applyRules(String key) {
     }
 
     private void logMatch(String key) {
-        if (Agent.LOG.isFinerEnabled()) {
+        if (Agent.LOG.isFinestEnabled()) {
```
jasonjkeller, line 82: doh!

newrelic-agent/src/main/resources/newrelic.yml
```diff
@@ -105,6 +105,12 @@ common: &default_settings
       # Default is 10000. Setting to 0 will disable.
       #max_samples_stored: 10000
 
+      # Whether the log events should include MDC from loggers with support for that.
+      include_context_data:
+
+        # When true, application logs will contain context data.
+        enabled: false
```
jasonjkeller, line 112: I wonder if we should also put the `include`/`exclude` options in the default yaml but keep them commented out like we do elsewhere?

```
      # A comma separated list of attribute keys whose values should be sent to New Relic.
      #include:

      # A comma separated list of attribute keys whose values should not be sent to New Relic.
      #exclude:
```

jasonjkeller (commented):
Overall looks good! Just a couple of questions and potential cleanup things.

I like the cleanup introduced with the `AppLoggingUtils`, that's a nice quality of life enhancement.

instrumentation/apache-log4j-2/src/main/java/com/nr/agent/instrumentation/log4j2/AgentUtil.java
meiao: This is a bit of a different case. `event.getContextData()` does not return a Map. It returns a ReadOnlyStringMap which extends Serializable and implements no interfaces.
This ReadOnlyStringMap provides a toMap method and the forEach that can be used here. There are many implementation of that interface, but the one I got in a unit test was SortedArrayStringMap. This impl, will instantiate a new HashMap when toMap is called. The forEach implementation is done using a for(int i...) loop.
So I'd have to do some profiling to see which one allocates less memory.

instrumentation/spring-aop-2/build.gradle
meiao: I was looking for mocking static methods. Found these and noticed that this module had no tests.

jasonjkeller (approved):
LGTM!!!

PR #1021, opened 2022-09-26
Adding stack trace to when logging CLM data retrieval error

meiao:
### Overview
When an error occurs while retrieving CLM data, a log message is written, but it is missing the stack trace on finer log levels.

This adds the stack trace, as other errors in the same class.


jasonjkeller (approved):

meiao:
Rebased to get the cats and http4s fixes.

PR #1022, opened 2022-09-27
Java 19 support

jasonjkeller:
All tests are 🟢 . Looks good to go.

### DIFF
--- a/instrumentation/cassandra-datastax-4.0.0/src/test/java/com/nr/agent/instrumentation/cassandra/CassandraNoInstrumentation.java
+++ b/instrumentation/cassandra-datastax-4.0.0/src/test/java/com/nr/agent/instrumentation/cassandra/CassandraNoInstrumentation.java
@@ -21,6 +21,7 @@ import com.newrelic.test.marker.Java15IncompatibleTest;
 import com.newrelic.test.marker.Java16IncompatibleTest;
 import com.newrelic.test.marker.Java17IncompatibleTest;
 import com.newrelic.test.marker.Java18IncompatibleTest;
+import com.newrelic.test.marker.Java19IncompatibleTest;
 import com.newrelic.test.marker.Java9IncompatibleTest;
 import org.cassandraunit.CassandraCQLUnit;
 import org.cassandraunit.dataset.cql.ClassPathCQLDataSet;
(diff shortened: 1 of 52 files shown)


instrumentation/play-shaded-async-http-client-1.0.0/build.gradle
```diff
@@ -18,28 +18,17 @@ verifyInstrumentation {
 }
 
 test {
```
tbradellis, line 20: I'm guessing that you'd prefer these be fixed in a separate branch.

instrumentation/cassandra-datastax-4.0.0/src/test/java/com/nr/agent/instrumentation/cassandra/CassandraNoInstrumentation.java
```diff
@@ -32,8 +33,8 @@
 import static org.junit.Assert.assertEquals;
 
 // Issue when running cassandra unit on Java 9+ - https://github.com/jsevellec/cassandra-unit/issues/249
-@Category({Java9IncompatibleTest.class, Java10IncompatibleTest.class, Java11IncompatibleTest.class, Java12IncompatibleTest.class, Java13IncompatibleTest.class, Java14IncompatibleTest.class,
-        Java15IncompatibleTest.class, Java16IncompatibleTest.class, Java17IncompatibleTest.class, Java18IncompatibleTest.class})
+@Category({ Java9IncompatibleTest.class, Java10IncompatibleTest.class, Java11IncompatibleTest.class, Java12IncompatibleTest.class, Java13IncompatibleTest.class, Java14IncompatibleTest.class,
+        Java15IncompatibleTest.class, Java16IncompatibleTest.class, Java17IncompatibleTest.class, Java18IncompatibleTest.class, Java19IncompatibleTest.class })
 @RunWith(InstrumentationTestRunner.class)
```
tbradellis, line 38: I've been wondering about this growing list of annotations similar to what you called out on the long list of `project.hasProperty(...)` checks in some build.gradles.  Not for this PR, but do you think we should consider a long term solution for these annotations as well?

tbradellis (approved):

instrumentation/play-shaded-async-http-client-1.0.0/build.gradle
jasonjkeller: Oh, I'll address those in this PR to see if the commented out code works as expected. If so, it would greatly clean things up.

instrumentation/cassandra-datastax-4.0.0/src/test/java/com/nr/agent/instrumentation/cassandra/CassandraNoInstrumentation.java
jasonjkeller: Honestly, we should just update these to only have annotations for Java versions that we actually test against. 

We should be able to simplify it to just this:

```
@Category({ Java11IncompatibleTest.class, Java17IncompatibleTest.class, Java19IncompatibleTest.class })
```

jasonjkeller: I think I'll leave such a change for a different PR, so we can discuss first.

PR #1029, opened 2022-10-04
NEWRELIC-4024 gradle

jasonjkeller:
Given the current fast paced release of Java it is unsustainable to always run all tests against every supported version of Java. To keep test times reasonable we only test against supported LTS Java releases as well as the latest non-LTS release of Java, which currently include Java 8/11/17/19.

This PR cleans up the following:
- Removes all of the `JavaNNIncompatibleTest` annotations for versions of Java that we no longer test against.
- Configures gradle so that the `-PtestN` gradle property only executes tests for specific versions of Java (i.e. 8/11/17/19). 
- Clean up readme a bit.

If tests are run with a `-PtestN` value for a version of Java that we don't test against (e.g. `./gradlew :instrumentation:vertx-core-3.3.3:test --rerun-tasks -Ptest10`) the tests will be prevented from executing and a message such as the following will be printed as part of the gradle output:

```
> Task :instrumentation:vertx-core-3.3.3:test SKIPPED
The newrelic-java-agent project does not support testing with the Java version configured by the -PtestN gradle property.
```


instrumentation/jax-rs-1.0/build.gradle
```diff
@@ -14,12 +14,6 @@ jar {
     'Implementation-Title-Alias': 'rest_annotations'}
 }
 
-test {
```
tbradellis, line 17: Do we know why these were only being used for 9 & 10?   I guess it doesn't really matter since there are no failing tests on 11 or 17. Just seemed a bit odd.

newrelic-cats-effect3-api/build.gradle.kts
```diff
@@ -58,22 +58,6 @@ tasks {
...
-        val jdk10: String by project
-        val jdk9: String by project
-        val jdk8: String by project
-
-        if (project.hasProperty("test10")) {
-            jvmArgs("--add-modules", "java.xml.bind")
-        } else if (project.hasProperty("test9")) {
-            jvmArgs("--add-modules", "java.xml.bind")
-        }
-
```
tbradellis, line 76: same curiosity here.

tbradellis (approved):
LGTM. just left a couple comments, but I think there not really issues because tests are passing.

meiao (approved):

instrumentation/jax-rs-1.0/build.gradle
jasonjkeller: I'm not entirely sure but I was wondering the same thing. I verified that the tests pass on later versions of Java but I don't know what was causing issues with 9/10 specifically. I also saw a number of tests that had `JavaNNIncompatibleTest` annotations on them up to Java 16/17 but then no longer had them for Java 18/19 but they mysteriously worked on these new Java versions, which I wouldn't have expected.

PR #1039, opened 2022-10-15
NEWRELIC-3834 add embedded tomcat jmx instrumentation

tbradellis:
Add instrumentation to cover embedded tomcat.
The agent doesn’t correctly query the following Tomcat JMX metrics when using embedded Tomcat with SpringBoot:

type=DataSource
type=ThreadPool
type=Manager

We add Tomcat JMX values for the above categories if the server was started via one of the following two methods:

```
org.apache.catalina.startup.HostConfig.start()
org.apache.catalina.core.StandardServer.startInternal()

```

It appears that when using embedded Tomcat with Spring Boot StandardServer does get instrumented properly:

```
Supportability/WeaveInstrumentation/WeaveClass/com.newrelic.instrumentation.tomcat-jmx/org/apache/catalina/core/StandardServer
Supportability/WeaveInstrumentation/WeaveClass/com.newrelic.instrumentation.tomcat-8.5.2/org/apache/catalina/core/StandardServer

```

By default, embedded Tomcat JMX is disabled by Spring Boot. In application.properties Tomcat mbeans need to be enabled by adding: server.tomcat.mbeanregistry.enabled=true

The problem though is that the Mbean group namespaces are different, standalone Tomcat uses the prefix Catalina for queries whereas embedded Tomcat uses the prefix Tomcat. Our tomcat-jmx instrumentation only queries MBeans with the Catalina prefix thus queries initiated by embedded Tomcat fail due to an incorrect group name.

Standalone Tomcat JMX:

Embedded Tomcat JMX

For now, the recommended workaround would be to utilize Custom JMX instrumentation by YAML and create a dashboard to view the metrics.

### Overview
Describe the changes present in the pull request

### Related Github Issue

### Testing
run local tests against embedded tomcat to confirm metrics are collected

### DIFF
--- a/instrumentation/tomcat-jmx/src/main/java/com/nr/agent/instrumentation/tomcat/TomcatUtils.java
+++ b/instrumentation/tomcat-jmx/src/main/java/com/nr/agent/instrumentation/tomcat/TomcatUtils.java
@@ -7,22 +7,50 @@
 
 package com.nr.agent.instrumentation.tomcat;
 
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.logging.Level;
-
 import com.newrelic.agent.bridge.AgentBridge;
 import com.newrelic.api.agent.NewRelic;
+
+import javax.management.MBeanServer;
+import javax.management.MalformedObjectNameException;
+import javax.management.ObjectName;
+import java.lang.management.ManagementFactory;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.logging.Level;
 
 public class TomcatUtils {
 
     private static final String JMX_PREFIX = "Catalina";
+    private static final String JMX_EMBEDDED_PREFIX = "Tomcat";
+    private static final String JMX_EMBEDDED_DATASOURCE_PREFIX = "org.apache.tomcat.jdbc.pool.jmx";
+
     private static final AtomicBoolean addedJmx = new AtomicBoolean(false);
 
     public static void addJmx() {
         if (System.getProperty("com.sun.aas.installRoot") == null) {
             if (!addedJmx.getAndSet(true)) {
-                AgentBridge.jmxApi.addJmxMBeanGroup(JMX_PREFIX);
-                NewRelic.getAgent().getLogger().log(Level.FINER, "Added JMX for Tomcat");
+                MBeanServer server = ManagementFactory.getPlatformMBeanServer();
+                //The ObjectName is different if embedded tomcat is used. Checking that this object has been registered
+                //will let us know that we are in an embedded tomcat. The Set returned by queryNames will return size 0 if the ObjectName is not
+                //found
+                try {
+                    if (server.queryNames(new ObjectName("Tomcat:type=Server"), null).size() == 1) {
+                        AgentBridge.jmxApi.addJmxMBeanGroup(JMX_EMBEDDED_PREFIX);
+                        NewRelic.getAgent().getLogger().log(Level.FINER, "Added JMX for Tomcat");
+
+                    }
+                    if(server.queryNames(new ObjectName("org.apache.tomcat.jdbc.pool.jmx:name=dataSourceMBean,type=ConnectionPool"), null)
+                            .size() == 1){
+                        AgentBridge.jmxApi.addJmxMBeanGroup(JMX_EMBEDDED_DATASOURCE_PREFIX);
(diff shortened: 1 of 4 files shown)


tbradellis:
Do not merge. There's still an issue with the DataSource metrics. They are not reported as expected, though Manager and ThreadPool are the same as the standard tomcat instrumetnation.

newrelic-agent/src/main/java/com/newrelic/agent/jmx/values/EmbeddedTomcatJmxValues.java
```diff
@@ -0,0 +1,111 @@
...
+        /*
+         * Provides information about the thread pool. The current thread count and the current number of threads which
+         * are busy.
+         */
+        metrics.add(new BaseJmxValue(name + ":type=ThreadPool,name=*", MetricNames.JMX_THREAD_POOL + "{name}/",
+                new JmxMetric[] { CURRENT_ACTIVE_COUNT, CURRENT_IDLE_COUNT, CURRENT_MAX_COUNT }));
+
+        /*
+         * Provides information about the data source by finding the number of active and idle connections and the max connections.
+         * In Tomcat 7, the number of max connections is represented by the maxActive attribute. In tomcat 8
```
tbradellis, line 85: Comment should be updated to reflect whatever the new mbean ends up being (and why it is different)

tbradellis:
Tested locally with Spring Boot and embedded Tomcat.
Requires:
spring.datasource.tomcat.jmx-enabled=true
server.tomcat.mbeanregistry.enabled=true
spring.datasource.type=org.apache.tomcat.jdbc.pool.DataSource

instrumentation/tomcat-jmx/src/main/java/com/nr/agent/instrumentation/tomcat/TomcatUtils.java
```diff
@@ -7,21 +7,29 @@
...
     public static void addJmx() {
         if (System.getProperty("com.sun.aas.installRoot") == null) {
             if (!addedJmx.getAndSet(true)) {
+                //We need to add all three possible types because embedded tomcat breaks out
+                //the types (which we named PREFIX) differently than standalone tomcat, yet the same Server/HostConfig_Instrumentation is used.
                 AgentBridge.jmxApi.addJmxMBeanGroup(JMX_PREFIX);
+                AgentBridge.jmxApi.addJmxMBeanGroup(JMX_EMBEDDED_PREFIX);
+                AgentBridge.jmxApi.addJmxMBeanGroup(JMX_EMBEDDED_DATASOURCE_PREFIX);
+
                 NewRelic.getAgent().getLogger().log(Level.FINER, "Added JMX for Tomcat");
```
tbradellis, line 33: I don't know if it makes more sense here to try and check wether to add this or just let it run through. 
I suppose we could check for the appropriate System properties and only add this MBeanGroup for those cases.

tbradellis: This is gonna need some rework

instrumentation/tomcat-jmx/src/main/java/com/nr/agent/instrumentation/tomcat/TomcatUtils.java
```diff
@@ -7,22 +7,45 @@
...
             if (!addedJmx.getAndSet(true)) {
-                AgentBridge.jmxApi.addJmxMBeanGroup(JMX_PREFIX);
-                NewRelic.getAgent().getLogger().log(Level.FINER, "Added JMX for Tomcat");
+                MBeanServer server = ManagementFactory.getPlatformMBeanServer();
+                //The ObjectName is different if embedded tomcat is used. Checking that this object has been registered
+                //will let us know that we are in an embedded tomcat. The Set returned by queryNames will return 0 if the ObjectName is not
+                //found
+                try {
+                    if (server.queryNames(new ObjectName("Tomcat:type=Server"), null).size() == 1) {
+                        AgentBridge.jmxApi.addJmxMBeanGroup(JMX_EMBEDDED_PREFIX);
```
tbradellis, line 37: Embedded tomcat will have 
`Tomcat:type=Server`
Non-embedded will have:
`Catalina:type=Server`

jtduffy (approved):
+1

PR #1046, opened 2022-10-20
NR-23881 - change the method used to get the MDC

meiao:
### Overview
The method used in the first PR for this feature in Log4J was introduced in version 2.7.
But the instrumentation module supports version 2.6.

So this is changing the method used to the existing in 2.6.


instrumentation/apache-log4j-2/src/main/java/com/nr/agent/instrumentation/log4j2/AgentUtil.java
```diff
@@ -44,7 +44,7 @@ public static void recordNewRelicLogEvent(LogEvent event) {
             Throwable throwable = event.getThrown();
 
             if (shouldCreateLogEvent(message, throwable)) {
-                ReadOnlyStringMap contextData = event.getContextData();
+                Map<String, String> contextData = event.getContextMap();
```
meiao, line 47: All implementations of this method (in 2.17) simply call `getContextDate().toMap()`.

jtduffy (approved):

PR #1048, opened 2022-10-20
NEWRELIC-3498 fix Vfs.Dir scala dir

tbradellis:
Fixes build issue.
We (I) added the apply scala plugin to the java.gradle script, which meant we were expecting scala classes in each module. Our build task however did not expect that so was appropriately throwing an exception when the directories for scala classes weren't available in modules that did not in fact compile any scala.
### Checks

### DIFF
--- a/gradle/script/java.gradle
+++ b/gradle/script/java.gradle
@@ -1,5 +1,4 @@
 apply plugin: 'java-library'
-apply plugin: 'scala'
 
 configurations {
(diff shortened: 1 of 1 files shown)


jasonjkeller (approved):

tbradellis:
scala tests are good, which is all this could potentially impact. ignoring inf_tracing problematic test.

PR #1054, opened 2022-10-22
Support Java 17-compiled instrumentation modules

jeffalder:
I want to use an instrumentation module I compiled with Java 17, and this is preventing me from doing so.

### Overview
Minor change to support specifications of Java 17 and later

### Related Github Issue
none

### Testing

### Checks

n/a Does your code contain any breaking changes? Please describe. 
n/a Does your code introduce any new dependencies? Please describe.

### DIFF
--- a/newrelic-weaver/src/main/java/com/newrelic/weave/utils/WeaveUtils.java
+++ b/newrelic-weaver/src/main/java/com/newrelic/weave/utils/WeaveUtils.java
@@ -182,7 +182,9 @@ public final class WeaveUtils {
     private static int getRuntimeMaxSupportedClassVersion() {
         try {
             double jvmSpecVersion = Double.valueOf(System.getProperty("java.specification.version"));
-            if (jvmSpecVersion >=11) {
+            if (jvmSpecVersion >= 17) {
+                return 61;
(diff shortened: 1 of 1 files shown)


newrelic-weaver/src/main/java/com/newrelic/weave/utils/WeaveUtils.java
```diff
@@ -182,7 +182,9 @@ public final class WeaveUtils {
     private static int getRuntimeMaxSupportedClassVersion() {
         try {
             double jvmSpecVersion = Double.valueOf(System.getProperty("java.specification.version"));
-            if (jvmSpecVersion >=11) {
+            if (jvmSpecVersion >= 17) {
+                return 61;
+            } else if (jvmSpecVersion >= 11) {
```
tbradellis, line 187: We are going to do something like below in place of the continued branch logic. This way we will also get 19 (and future non-lts and lts versions as well)

```
if version >= 11
  return version + 44;

```
Do you want to make that change? Or I can just pull this into something local and change it.

jeffalder:
@tbradellis Your logic makes total sense to me. You can create a different PR. Just link back to this one and close this. ✅

PR #1066, opened 2022-10-31
Tracer debug logging

jasonjkeller:
Adds some additional logging to help troubleshoot `Inconsistent state! tracer != last tracer` errors. 

Additional logging will appear when the following is set:

```
-Dnewrelic.config.log_level=finest 
-Dnewrelic.debug=true
```

Grep'ing the agent logs for `Tracer Debug:` should produce all of the debug log messages related to this issue, which shows how tracers are being created and processed.

For context, this issue occurs when the agent is handling some async activity and is adding tracers for each method call and popping tracers off the stack as each method call completes. The agent tracks what the first tracer (root tracer) is when the async activity started and by the end we would expect that that last tracer we pop off should be the root, but in this case it isn't, which means that the agent has lost track of the correct call stack into this async code. When the pointer to the tracer object at the top of the stack differs from the actual tracer object that gets popped off then the agent is in an inconsistent state. If this happens the agent discards the async information it was attempting to record and logs the inconsistent state error.

It could potentially be caused by an internal error in some instrumented code that causes the agent to exit early without properly handling it or if the agent loses track of the active transaction.

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/TransactionActivity.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/TransactionActivity.java
@@ -338,7 +338,9 @@ public class TransactionActivity {
         } else {
             if (tracer.getParentTracer() != null) {
                 lastTracer = tracer;
-                Agent.LOG.log(Level.INFO, "Tracer Debug: lastTracer set to {0}", tracer);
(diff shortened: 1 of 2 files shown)


tbradellis (approved):
lgtm

PR #1067, opened 2022-11-03
Adding .snyk suppression files for known critical/high false positives

kford-newrelic:
Adding/updating .snyk suppression files for known false positives. Covering only false critical/high issues.


tbradellis (approved):
lgtm

PR #1070, opened 2022-11-08
NR-26418 - More logs on the scripts that sends data to S3

meiao:
### Overview
Improves the validation of the commands executed in the script that sends data to S3.

### DIFF
--- a/.github/workflows/Release-CopyToS3-prod.yml
+++ b/.github/workflows/Release-CopyToS3-prod.yml
@@ -10,11 +10,12 @@ on:
 
 jobs:
   copy-to-s3:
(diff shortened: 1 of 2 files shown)


.github/workflows/README.md
```diff
@@ -12,20 +12,12 @@ This command creates custom jars that we can share with customers.
 - **description** - a description that will show on the summary, so we know the contents of that custom jar.
 
 
-### Z - Sandbox job
```
meiao, line 15: This sandbox job is not really necessary. There is a "hack" to get workflows to show.

twcrone (approved):

Issue #1000, opened 2022-09-05
Slick + Datasource instrumentation is broken in 7.9.0 with scala 2.13


smiklos:
## Description

We are using Akka http 10.2.+ and Sangria so the only route we monitor is /graphql 

After upgrading to the current latest agent version (7.9.0) the agent seems to loose track of transactions across method calls. 
I can still manually create a transaction and link to it via its token but the transaction propagation between methods and future chaining no longer works at all.  I can only suspect the work related to https://github.com/newrelic/newrelic-java-agent/pull/876 as  cause

## Expected Behavior
When a web transaction starts we used to have segments span across the whole app, measuring db time and external calls. Now all of that is gone. Transactions still start but none of the above mentioned segments/metrics are measured anymore. 

## Your Environment
Tried both openjdk 11 and graalvm  22.1

(Migrate to Jira)

kford-newrelic:
@smiklos does the behavior you mention happen with `Akka HTTP` alone or is the combined use of `Sangria` necessary for the problem to occur? Any chance you have a simple app that can reproduce the problem?

While we can't commit to a date when we can investigate this further, having a small app that repros the problem is helpful!

smiklos:
@kford-newrelic  I tried to reproduce this locally but it does work using akka http and sangria alone. 
However it seems that if I check out the latest main branch of the agent, remove code that was introduced by [this pr](https://github.com/newrelic/newrelic-java-agent/pull/876) ,, build it locally and it actually works totally fine. 

Somehow the changes in that pr nukes jdbc datasource (and maybe other) instrumentation. 
The pr was meant to fix an issue with scala 2.12 so curious how it affected us running on scala 2.13 but it certainly did

smiklos:
I've ran the agent with debug and can see a bunch of instrumentation errors that happen during the transformation phase of scala final fields (as introduced by the pr I linked to) 

It shows that it is unable to transform several classes that are pretty important for a scala app to work with the agent : 

```
2022-11-10T14:44:42,665+0100 [25062 1] com.newrelic DEBUG: Instrumenting scala class scala/concurrent/impl/Promise$Transformation
2022-11-10T14:44:42,666+0100 [25062 1] com.newrelic FINE: Unable to transform class scala/concurrent/impl/Promise$Transformation
java.lang.TypeNotPresentException: Type scala/util/Try not present
[...]

PR #1071, opened 2022-11-09
slick 3.4.0 scala 2.13 instrumentation module

smiklos:
### Overview

There was no support for the latest scala slick lib version 3.4.0. 
This pr implements that with scala 2.13 support.

### Related Github Issue

[I reported a vague issue a month ago](https://github.com/newrelic/newrelic-java-agent/issues/1000)

I somewhat believe this is partially what causes our instrumentation not to work on newer agent versions though I can't tell what exactly is the issue. Making this change will only bring us closer to completeness. 

### Testing
Pretty much copied over the 2.12-3.2.0 module and made it work with scala 2.13 and the 3.4.0 agent version. 

I did run the locally built agent against my app and according to the debug logs the agent was able to instrument the `slick.util.AsyncExecutor` class. (Note however that I could still not see sql traces in my apm view.)

The same tests succeed for this module as the ones for the earlier version though looking at it more closely they are all ignored... wth 
 
### Checks

[ x] Are your contributions backwards compatible with relevant frameworks and APIs?


instrumentation/slick-2.13_3.4.0/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+jar {
+    manifest { attributes 'Implementation-Title': 'com.newrelic.instrumentation.slick-2.13_3.4.0' }
+}
+
+verifyInstrumentation {
+    // scala 11 should be instrumented by another module
+    fails 'com.typesafe.slick:slick_2.12:[3.4.0,)'
+
+    // scala 13
+    passesOnly 'com.typesafe.slick:slick_2.13:[3.4.0,)'
```
smiklos, line 28: Not entirely sure what the upper bound should be here

smiklos:
I would also appreciate if someone explained how could we run the slick module 2.12-3.2.0 when we are on scala 2.13? 
It seems that some byte code manipulation logic got changed in the agent version 7.9 that disables the slick integration. since the 3.2.0 version of slick was built on scala 2.12 it shouldn't really be compatible with 2.13.

kford-newrelic:
@smiklos thanks for the PR! One of our engineers will be taking a look at this soon - standby!

tbradellis:
@smiklos Thanks for pointing this out and the PR!
Forgive me if I get some scala terminology wrong, I'm somewhat new to scala.

The scala version isn't really the determining factor for instrumentation.  The Weaver visits the components of the class file structure for matches, so as long as the class and methods being weaved are the same (regardless of compile version) then the instrumentation would apply.

I noticed the overlapping instrumented versions between 2.12 and 2.13 slick 3.2.0+ versions.  As you saw, it looks like the "breaking change" was simply the previous use of a default parameter in the Slick libraries and the move to no default for that `apply` function.  However, in testing with our `verifyInstrumentation` plugin I can see that the more general case with no default will apply the instrumentation whether there is a default parameter or not. But, weaving the more specific signature will cause the break.
When there's overlap like this, we often have to find some class to weave that will prevent the other module from loading. It looks like in this case we could just change our old instrumentation `slick-2.12_3.2.0` to weave with a non default parameter specified and it will apply across scala and slick versions the way we need. (At least based on how our instrumentation verification tests work). 

I made that change in the branch I pulled in for testing:
https://github.com/newrelic/newrelic-java-agent/pull/1072/files

Can you validate that doing it the way I'm suggesting works with you application?  
And if so, do you want to make that change here? 
Also, happy to just finish out in the branch I pulled in. Don't want to send too much work back your way, totally up to you.

smiklos:
@tbradellis 
Thanks for looking into it! Indeed your prebuilt jar works on my app. It's fine by me if you can merge your branch. 
I'm still a bit hesitant to support the idea that some modules in the agent are compiled with scala 2.12 as in scala these minor versions are not binary compatible. In scala you'd always cross compile the libs to each target version (2.11, 2.12, 2.13 for ex) and so I think the agent should follow a similar approach.   

Anyways after this we only need to solve the breaking change introduced in 7.9 as I can still not see database calls in the UI , see this [issue](https://github.com/newrelic/newrelic-java-agent/issues/1000) for more detail :)

tbradellis:
Closing here because we will merge this work from the below branch where we moved these changes into one module:
https://github.com/newrelic/newrelic-java-agent/pull/1072/commits
Thanks again!

PR #1093, opened 2022-11-28
NR-00000 - Removing "bad" glassfish version from verifier

meiao:
### Overview
The glassfish-6 instrumentation shouldn't apply to Glassfish 7, but it did apply to 7.0.0-M10.

Disabling that version for now.


jtduffy (approved):
+1

meiao:
Rebased the original commit with main so the fix for the AITs is in the branch.

PR #1094, opened 2022-11-28
NR-00000 - Fixing Python version on AITs due to new GHA runner version

meiao:
### Overview
A new GHA runner version provides a newer Python version. This changes the script to use the new version.


jtduffy (approved):

meiao:
The proposed solutions in the PR did not fix the problem. A different PR fixes it.

PR #1098, opened 2022-12-01
NR-00000 - Update checkout and setup-java actions

meiao:
### Overview

Some actions used by the workflows use features being deprecated by Github ([save-state and set-output commands](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) and [Node 12](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/).

This PR updates those actions and updates the only use of set-output in the workflows.


.github/workflows/X-Reusable-VerifyInstrumentation.yml
```diff
@@ -48,20 +48,20 @@ jobs:
             echo " }"
           ) | jq -c .)
           # save the output of the job
-          echo "::set-output name=modules::$modules"
+          echo "modules=$modules" >> $GITHUB_OUTPUT
```
meiao, line 51: This is the only changes that is not updating the actions.

jasonjkeller (approved):

Issue #1091, opened 2022-11-28
External HTTP requests missing /External/{host}/all rollup metric


kford-newrelic:
### Description
Customer upgraded from Weblogic 10 to Weblogic 12, while also upgrading their agent from 3.x to 7.x.. After doing so they are no longer seeing their external calls. It looks like this is likely related to the `HttpUrlConnection` instrumentation which has undergone massive refactoring from the 3.x agent version they upgraded from to the current 7.x version.

The following screenshot shows two spans that got generated by the `HttpUrlConnection` instrumentation but neither has an external call associated with it. This is likely because neither of the instrumented methods in the trace, `connect` and `getOutputStream`, will result in the instrumentation calling `reportAsExternal` which is required to create an "External HTTP" span that will link to an external host.

Also note that calling `reportAsExternal` is what results in the External rollup metrics getting created. The fact that the `External/{HOST}/all` rollup metric is missing further confirms that `reportAsExternal` is never being called by the instrumentation.

<img width="1145" alt="external-call-issue-gh" src="https://user-images.githubusercontent.com/3496648/206064094-635e6ce1-f72f-4b41-b6d9-0b1e03f92a1d.png">

### Acceptance Criteria
Need to investigate the HTTP instrumentation to determine if there may be an issue related to the ordering of `HttpUrlConnection` method calls - which has been an issue in a past issue.

kford-newrelic:
Targeted for the Jan-Mar quarter

PR #1102, opened 2022-12-09
NR-70183 httpurlconnection fix

jasonjkeller:
Refactoring of `httpurlconnection` instrumentation due to several bugs discovered.

Behavior expected when `HttpURLConnection` APIs are called in different combos and orders after the refactoring.
* If only `connect` is called, then NO request is made over the wire and NO external call is reported. The instrumentation starts a `TimerTask` if `connect` is called first, waits for a set period of time (currently configured to be 1 minute) to determine if any further `HttpURLConnection` APIs are called before deciding how to proceed. If no other API is called, then the segment is just ignored and no external is reported. If any other method is called an external call will be recorded.
* Calling any of `getOutputStream`, `getInputStream`, `getResponseCode`, or `getHeaderFields` all result in an external call being recorded.

Fixes: https://github.com/newrelic/newrelic-java-agent/issues/1091

It should also address these related issues:
https://issues.newrelic.com/browse/NEWRELIC-3496
https://issues.newrelic.com/browse/NR-70183

The `externals.py` AIT failures will pass when this PR is merged: https://github.com/newrelic/java-agent-integration-tests/pull/138


instrumentation/httpurlconnection/src/main/java/com/nr/agent/instrumentation/httpurlconnection/MetricState.java
```diff
@@ -10,90 +10,261 @@
...
+                endSegmentForNonNetworkCall(connection);
+            }
+        };
+
+        /*
+         * The following tests do a Thread.sleep to account for this delay. If this value is changed then the tests will also need to be updated.
+         * functional_test/src/test/java/com/newrelic/agent/instrumentation/pointcuts/net/HttpURLConnectionTest
+         * instrumentation/httpurlconnection/src/test/java/com/nr/agent/instrumentation/httpurlconnection/MetricStateConnectTest
+         */
+        long segmentExpirationDelayInMillis = 1000L;
```
jasonjkeller, line 100: Is 1 second a reasonable default for ending the segment timing when only `connect` and/or `getOutputStream` are called? Or does a different value make more sense?

jasonjkeller: Hmmm, the github test runner might have answered the first part of this for me. I don't think 1 second is long enough but I'm not sure what is ideal.

I'm seeing `com.newrelic.agent.instrumentation.pointcuts.net.HttpURLConnectionTest > connectFirstTest` fail regularly when run here on GHA even though it passes consistently when run locally.

This test uses the `HttpURLConnection` API as follows:

```
connection.connect();
connection.getOutputStream();
connection.connect(); // should be no-op
```

This would be a scenario where the `TimerTask` expires, which should result in reporting the external call and ending the segment timing. However, I believe the reason that it is failing is that the 1 second timeout isn't long enough to evaluate `getOutputStream` once it is invoked after the first `connect` call, which means that the segment timing is simply ended and ignored, causing the test assertion to fail because no `External/` metrics are reported.

Given this, I think it means that a 1 second timeout likely won't be long enough in some customer environments either.

jasonjkeller: I upped the timeout to 10 seconds and the tests passed on GHA first time. I guess that confirms the issue.

New question, is 10 seconds a reasonable default?

jasonjkeller: This should no longer be a concern. The code has been refactored so that the `TimerTask` will only ever apply if only `connect` is called, in which case it will timeout in 1 minute and ignore the segment so that it won't ever show up in a trace.

meiao (approved):

PR #1105, opened 2022-12-19
NEWRELIC-4191 Properly link tokens so transaction is available to user code

jtduffy:
Insure transactions are visible to user code.

### DIFF
--- a/instrumentation/grpc-1.40.0/src/main/java/com/nr/agent/instrumentation/grpc/GrpcUtil.java
+++ b/instrumentation/grpc-1.40.0/src/main/java/com/nr/agent/instrumentation/grpc/GrpcUtil.java
@@ -48,16 +48,4 @@ public class GrpcUtil {
             }
         }
     }
-
-    /**
-     * Expire the supplied {@link Token} instance if it's non-null.
-     *
-     * @param token The {@link Token} to expire
-     */
-    public static void expireToken(Token token) {
-        if (token != null) {
-            token.expire();
-            token = null;
-        }
-    }
 }
--- a/instrumentation/grpc-1.40.0/src/main/java/com/nr/agent/instrumentation/grpc/GrpcUtil.java
+++ b/instrumentation/grpc-1.40.0/src/main/java/com/nr/agent/instrumentation/grpc/GrpcUtil.java
@@ -1,0 +1,63 @@
+/*
+ *
+ *  * Copyright 2022 New Relic Corporation. All rights reserved.
+ *  * SPDX-License-Identifier: Apache-2.0
+ *
+ */
+
+package com.nr.agent.instrumentation.grpc;
+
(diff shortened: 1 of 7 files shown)


instrumentation/grpc-1.40.0/src/main/java/com/nr/agent/instrumentation/grpc/GrpcUtil.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Expire the supplied {@link Token} instance if it's non-null.
+     *
+     * @param token The {@link Token} to expire
+     */
+    public static void expireToken(Token token) {
+        if (token != null) {
+            token.expire();
+            token = null;
```
meiao, line 60: I don't think this method is very useful.
The token that gets nulled out here is the reference that is used in this method.
If while debugging, you step out of this method, you'll see that the reference that was passed to this method is still referencing the original.

instrumentation/grpc-1.40.0/src/main/java/io/grpc/internal/ServerStream_Instrumentation.java
```diff
@@ -28,56 +29,24 @@ public abstract class ServerStream_Instrumentation {
...
+        GrpcUtil.finalizeTransaction(token, status, new Metadata());
+        GrpcUtil.setServerStreamResponseStatus(status);
 
         Weaver.callOriginal();
 
-        if (token != null) {
-            token.expire();
-            token = null;
-        }
+        GrpcUtil.expireToken(token);
```
meiao, line 49: The token still needs to be nulled out here.

meiao (changes requested):

instrumentation/grpc-1.40.0/src/main/java/io/grpc/internal/ServerStream_Instrumentation.java
```diff
@@ -28,56 +29,24 @@ public abstract class ServerStream_Instrumentation {
...
+        GrpcUtil.finalizeTransaction(token, status, metadata);
+        GrpcUtil.setServerStreamResponseStatus(status);
 
         Weaver.callOriginal();
 
-        if (token != null) {
-            token.expire();
-            token = null;
-        }
+        GrpcUtil.expireToken(token);
```
meiao, line 38: Still need to expire the token.

instrumentation/grpc-1.40.0/src/main/java/com/nr/agent/instrumentation/grpc/GrpcUtil.java
jtduffy: Good catch ... I'll refactor and remove this method

meiao (approved):

meiao (approved):

PR #1106, opened 2022-12-29
Add ZIO 2 support

duartesaraiva98:
### Overview
Adding instrumentation for ZIO 2 to the agent. Currently there is only instrumentation for ZIO 1 which is not compatible with ZIO 2. The majority of this code is based on ZIO 1 implementation with some slight changes to make it work. 

Background for why this change is needed.

When we migrated to ZIO 2 we noticed some issues in the transactions data.

#### Problem 1 - Segments in Web Transactions do not get child segments

Stack: 

- Scala 2.13
- ZIO 2.0.2
- newrelic-scala-api 7.11.1

What we currently experience is that the web-transactions that call on ZIO methods don’t track child
segments. In the example of the image below, the tree of segments should be ”command-call-1” as the
root segment and two child segments ”test-pg-call-1” and ”test-pg-call-2”. As you can see, only the
root segment (”command-call-1”) is reported back.

zio2-web-example

##### In ZIO 2:
Stack: 

- Scala 2.13
- ZIO 2.0.2
- newrelic-scala-api 7.11.1

What we currently experience is that non-web transactions are started but not linked to the segments
inside the method that starts the transaction. In the example of the image below, the tree of segments
should be ”command-call-1” as the root segment and two child segments ”test-pg-call-1” and ”test-
pg-call-2”. The reality is that none of the segments show up and the transaction is ”completed” in
just a few milliseconds.

zio2-non-web

##### In ZIO 1:
Stack: 

- Scala 2.13
- ZIO 1.0.10
- newrelic-scala-api 7.11.1

For the non-web transactions, we saw the same issue happening in ZIO 1.x. The transaction exists
but completes in just a few milliseconds and is not linked to any segments. Even when using the
newrelic-scala-zio-api.

zio1-non-web

#### Expected Result for both Non-Web and Web Transactions
Stack: 

- Scala 2.13
- ZIO 1.0.10
- newrelic-scala-api 7.11.1

What we expect for both non-web and web transactions can be seen in the image below. This is the
behavior that we have always seen with new relic, even before we used ZIO. In this example, we were
using ZIO 1.x in combination with new relic’s instrumentation for ZIO, which is only compatible with
ZIO 1.x. In this example you can see the root segment ”command-call-1” as well as the two child
segments ”test-pg-call-1” and ”test-pg-call-2”

zio1-web-tx

### Related Github Issue
[Not applicable]

### Checks

[Yes ] Are your contributions backwards compatible with relevant frameworks and APIs?
[ No] Does your code contain any breaking changes? Please describe. 
[ No] Does your code introduce any new dependencies? Please describe.


duartesaraiva98:
@tbradellis Is this something you could take a look at? You added instrumentation for ZIO 1, right?

kford-newrelic:
@duartesaraiva98 thanks for the PR!

We've got a fair bit of work on our board at the moment but we'll review this as soon as possible

kford-newrelic:
@duartesaraiva98 apologies but we had to bump this out to the Apr-Jun quarter - we're over-capacity for the current quarter

instrumentation/zio-2/src/main/java/zio/TokenAwareRunnable.java
tbradellis, line 1: @duartesaraiva98 
Sorry for the delay on getting this reviewed.  I'd like to get it into main this week for our upcoming release. 
It looks like this group of 4 classes in `zio` are duplicates of the correct ones in `zio.internal`.
Do you want to push a change to delete those to this PR? Or, I can make the change where I pulled your commits in here for testing:
https://github.com/newrelic/newrelic-java-agent/pull/1144
Whatever works for you.

tbradellis (changes requested):

newrelic-scala-zio-2-api/src/test/scala/com/newrelic/zio/api/TraceOpsDSLTest.scala
```diff
@@ -0,0 +1,225 @@
...
+  val threadPoolTwo: ExecutionContext = executorService(3)
+  val threadPoolThree: ExecutionContext = executorService(3)
+
+  @After
+  def resetTxn() = {
+    com.newrelic.agent.Transaction.clearTransaction
+  }
+
+  @Test
+  def asyncTraceProducesOneSegment(): Unit = {
```
tbradellis, line 33: All 6 unit tests seem to be failing.
https://github.com/newrelic/newrelic-java-agent/actions/runs/5079780519/jobs/9128030914?pr=1144

tbradellis: I believe I have this resolved. Running tests now. These need to be run in a separate workflow as functional tests.

tbradellis: @duartesaraiva98 I was able to resolve the testing issues.
My changes are here:
https://github.com/newrelic/newrelic-java-agent/pull/1144/files

However, testing the instrumentation locally with restful webservices in:
https://github.com/zio/zio-quickstarts
 I can see the module loads, but we lose the Transaction context and only report `NettyDispatcher` and don't see the underlying traces for the route.
 I saw this on 2.0.13; scala 2.13.8, and Java 11.

tbradellis:
We are closing this for now, because we are not able to get it on the roadmap to investigate the issue with losing Transaction context described in the comment [here](https://github.com/newrelic/newrelic-java-agent/pull/1106#discussion_r1211734129).

Feel free to resubmit with that addressed and we can likely get it merged.  Tests were otherwise passing for the PR.

kford-newrelic:
Reviewing for possible further action in the oct-dec quarter

kford-newrelic:
Not able to address fixing the lost transaction issue in the coming oct-dec quarter. Will consider again for the following quarter.

jasonjkeller:
Here's another repro app, this one from [NR-175811](https://new-relic.atlassian.net/browse/NR-175811) using ZIO 1.x:
[NewrelicTestProject.zip](https://github.com/newrelic/newrelic-java-agent/files/13271941/NewrelicTestProject.zip)

PR #1114, opened 2023-01-03
JdbcHelper urlTofactory Map replacement

jtduffy:
Replaces the existing ConcurrentHashMap used for the `urlToFactory` and `urlToDatabaseName` maps with a simple, custom cache that can expire items in the cache based on the last time an item was accessed.

The ConcurrentHashMap implementations were causing issues when used with MySql replication, where URLs to target replica databases were comprised of unique, generated Strings, causing the maps to grow without bound, which could eventually cause an OOM exception.


agent-bridge-datastore/src/main/java/com/newrelic/agent/bridge/datastore/ExpiringValueCache.java
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * A cache implementation that uses a composition pattern to wrap a {@link ConcurrentHashMap}
+ * populated with {@link TimestampedMapValue} instances. An instance of this class is constructed
+ * with function that contains the logic of when to evict items from the cache.
+ *
+ * @param <K> the class type of the cache key
+ * @param <V> the class type of the value object
+ */
+public class ExpiringValueCache<K, V> {
```
Stephan202, line 21: This agent uses [Caffeie](https://github.com/ben-manes/caffeine/) in [several places](https://github.com/newrelic/newrelic-java-agent/search?q=caffeine); is there a reason not to use it here as well? Using it also avoids the need to start an extra thread.

meiao: Yes, we tried using Caffeine here, but due to the way this module is loaded, we were not able to use it.

jtduffy:
@meiao  Comments addressed: 
- Instant -> long
- static executor instance

tbradellis (approved):
+1 lgtm
I only have one question that is just for my own understanding.  Was there a specific reason to implement this custom cache vs using the caffeine dependency we have?  Or are we doing something with the timestamp value that isn't available in caffeine?

meiao:
It is not trivial to use Caffeine in this module due to the way it is loaded to the JVM.

PR #1126, opened 2023-01-20
Nr 59596 jboss logging

jasonjkeller:
Adds support for JBoss Logging.

Also adds a new `instrumentation` attribute on log events for all logging frameworks so that we can easily identify which instrumentation module was responsible for reporting a given log event. This makes troubleshooting potential issues much easier.

The `logging.py` AIT contains most of the testing for this instrumentation since a JBoss/Wildfly server needs to be started up to test it properly. It's been run successfully locally a bunch of times. https://github.com/newrelic/java-agent-integration-tests/pull/153


jtduffy (approved):
+1

I like the `INSTRUMENTATION` attribute too.

PR #1140, opened 2023-02-01
Instrument methods with dispatcher=true when they invoke token.link() outside of a transaction 

sdaubin:
### Overview
There are cases for async transactions in which customers call `Token.link()` outside of a transaction, which causes linking to fail.  We can detect this happening and re-transforming the class so that the method is traced with transactions enabled.

To determine the calling class/method, we generate an exception and use the stack trace elements.  Because we're doing this automatically, I added rate limiting so that we only try this once per second.  We could make that time configurable, and we could add a config switch to completely disable this behavior.

### Related Github Issue
https://issues.newrelic.com/browse/NR-74916

### Testing
The referenced ticket contains an application which can be used to reproduce the issue and verify the fix.

### Checks

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/instrumentation/InstrumentationImpl.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/instrumentation/InstrumentationImpl.java
@@ -18,6 +18,7 @@ import com.newrelic.agent.bridge.AgentBridge;
 import com.newrelic.agent.bridge.ExitTracer;
 import com.newrelic.agent.bridge.Instrumentation;
 import com.newrelic.agent.bridge.NoOpTransaction;
+import com.newrelic.agent.config.ClassTransformerConfig;
 import com.newrelic.agent.instrumentation.classmatchers.DefaultClassAndMethodMatcher;
 import com.newrelic.agent.instrumentation.classmatchers.ExactClassMatcher;
 import com.newrelic.agent.instrumentation.classmatchers.HashSafeClassAndMethodMatcher;
@@ -45,6 +46,7 @@ import com.newrelic.agent.tracers.UltraLightTracer;
 import com.newrelic.agent.tracers.metricname.MetricNameFormat;
 import com.newrelic.agent.tracers.metricname.MetricNameFormats;
 import com.newrelic.agent.util.InsertOnlyArray;
+import com.newrelic.api.agent.NewRelic;
 import org.objectweb.asm.Opcodes;
 import org.objectweb.asm.Type;
 
@@ -55,7 +57,9 @@ import java.lang.reflect.Modifier;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
+import java.util.function.Supplier;
 import java.util.logging.Level;
 
 import static com.newrelic.agent.Transaction.SCALA_API_TRACER_FLAGS;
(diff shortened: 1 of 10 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/instrumentation/InstrumentationImpl.java
```diff
@@ -540,8 +572,79 @@ public void instrument(String className, String metricPrefix) {
...
+    public void instrument() {
+        final boolean check = autoInstrumentCheck.get();
+        if (check) {
+            final StackTraceElement stackTraceElement = getApplicationStackTraceElement(
+                    new Exception().fillInStackTrace().getStackTrace());
+            if (stackTraceElement != null) {
+                instrument(stackTraceElement, TraceDetailsBuilder.newBuilder().setAsync(true).build());
+            }
+        }
+        NewRelic.recordMetric("Supportability/InstrumentationImpl/instrument", check ? 1f : 0f);
```
tbradellis, line 585: It looks like this might over report this metric in some cases where ` instrument(StackTraceElement stackTraceElement, TraceDetails traceDetails)` is called and would return immediately because instrumentedStackTraceElements.contains(stackTraceElement) is true.  Should we also check a returned boolean from instrument(...)?

I'm thinking this line is what we'd miss in the check.
    private void instrument(StackTraceElement stackTraceElement, TraceDetails traceDetails) {
        if (instrumentedStackTraceElements.contains(stackTraceElement)) {
            return;
        }

Would doing something like this make sense?

    private boolean instrument(StackTraceElement stackTraceElement, TraceDetails traceDetails) {
        if (instrumentedStackTraceElements.contains(stackTraceElement)) {
            return false;
        }
        ...
        ...
        ...
 And then checking again inside the first predicate `check`       
Or do I misunderstand the intent of the metric?
@meiao

sdaubin: I'm performing a little trick with that metric, which is a technique we often use in custom instrumentation.  The count of the metric is how many times we hit the line of code and the value is the count of how many times `check` is true.  Summing the value give you the overall count of `check == true`, but the average of the metric tells you the percentage of time check was true.

meiao: We got that. But we ran a test that should instrument only one method. And when we checked the metric, the sum was 11.

What happened was:
- a call was made to a method that needs to be instrumented
- that method is scheduled to be instrumented
- the time limit to instrument another method elapses
- another call to the same method is made
- [1]
- method is instrumented

in [1] the rateLimiter will say it is fine to instrument that method. Which will make the metric be sent as 1.0f. But when the instrument method is invoked, it will check that that method is in the `instrumentedStackTraceElements` and will return not scheduling another instrumentation.

sdaubin: Sure.  That can happen.  But it won't actually try to reinstrument because of this, right?

         if (instrumentedStackTraceElements.contains(stackTraceElement)) {
             return;
         }
         instrumentedStackTraceElements.add(stackTraceElement);

meiao: Correct, it wont instrument again. But the metric will be sent with 1 in that case.

meiao: It is a little nitpicking, but the metric currently means how many times the rate limiter allowed execution thru, not how many methods were instrumented. But its name is instrumented.

tbradellis (approved):

patrickjamesbarry:
"we could add a config switch to completely disable this behavior."
What is the jvm arg we could use to disable this?

jtduffy:
Setting the system property `-Dnewrelic.config.auto_async_link_rate_limit=0` will disable this behavior.

Issue #1115, opened 2023-01-03
NullPointerException thrown when calling addCustomAttributes


dhilpipre:
[NOTE]: # In a least one use case, a NullPointerExeception is thrown when calling any of the addCustomAttribute or addCustomAttributes method on TracedMethod.  The exception is being thrown in the shouldAddAttribute() of AbstractTracer.

## Description
If the following is true then a NullPointerExeception is  thrown by the Agent,  The async token is created but no transaction is started (No-op token created).   If a method is called with the @Trace attribute async set to true and the token is linked then if you try to add an attribute to TracedMethod it results in a NullPointerExeception.    This seems to happen because when the resulting TransactionActivity is created its Transaction object is set to null.
 In some cases it results in the application not working properly.

## Expected Behavior
The code for shouldAddAttribute should check if the getTransaction is returning null if if it is return false.

## Steps to Reproduce
Attached is a simple Java application that illustrates the problem
[code.tar.gz](https://github.com/newrelic/newrelic-java-agent/files/10339137/code.tar.gz)

(Replicate to Jira)

## Your Environment
Simply running the attached code from the command line with the Java Agent enabled.

## Additional context
All that is probably needed is make a simple update to the code for shouldAddAttribute to include a check for a null transaction

kford-newrelic:
@dhilpipre thanks for letting us know about this issue!

PR #1141, opened 2023-02-01
NullPointerException thrown when calling addCustomAttributes

tbradellis:
[_Before contributing, please read our [contributing guidelines](https://github.com/newrelic/newrelic-java-agent/blob/main/CONTRIBUTING.md) and [code of conduct](https://github.com/newrelic/.github/blob/main/CODE_OF_CONDUCT.md)._](https://issues.newrelic.com/browse/NEWRELIC-6338)

### Overview
null guards shouldAddAttribute
### Related Github Issue

[NOTE]: # In a least one use case, a NullPointerExeception is thrown when calling any of the addCustomAttribute or addCustomAttributes method on TracedMethod. The exception is being thrown in the shouldAddAttribute() of AbstractTracer.

Description

If the following is true then a NullPointerExeception is thrown by the Agent, The async token is created but no transaction is started (No-op token created). If a method is called with the @Trace attribute async set to true and the token is linked then if you try to add an attribute to TracedMethod it results in a NullPointerExeception. This seems to happen because when the resulting TransactionActivity is created its Transaction object is set to null.
In some cases it results in the application not working properly.
### Testing
Testing against provided repro app

### Checks

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/tracers/AbstractTracer.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/tracers/AbstractTracer.java
@@ -8,7 +8,6 @@
 package com.newrelic.agent.tracers;
 
 import com.newrelic.agent.Agent;
-import com.newrelic.api.agent.AttributeHolder;
 import com.newrelic.agent.MetricNames;
 import com.newrelic.agent.Transaction;
 import com.newrelic.agent.TransactionActivity;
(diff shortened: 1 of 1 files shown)


tbradellis:
confirmed resolved in repro app attached here:
https://github.com/newrelic/newrelic-java-agent/issues/1115

jtduffy (approved):
+1

PR #1146, opened 2023-02-02
Infinite Tracing payload compression and batching

toddwest:
### Overview

This PR adds support for Infinite Tracing payloads to be both compressed and batched by default. The following changes are included as part of this work:

- Implements support for compressing infinite tracing payloads
- `gzip` compression is enabled by default for batching **and** non-batching configurations
- Implements support for batching spans within infinite tracing payloads
- Defaults for batching are:
  - Batching `enabled` by default
  - Up to `100` spans bundled per batch
  - Waiting (linger) for up to `5s` to fill up the batch before sending

The span data is still streaming over HTTP/2 using gRPC but utilizing compression and batching will significantly reduce payload sizes and bytes over the wire, resulting in significantly lower network transfer costs for customers. 

### Testing

All new functionality has been covered with unit tests where possible.

### Checks

- Are your contributions backwards compatible with relevant frameworks and APIs?
  - Yes

- Does your code contain any breaking changes?
  - No (although infinite tracing defaults are changing)

- Does your code introduce any new dependencies?
  - Only a `testImplementation` dependency for parameterized tests

### DIFF
--- a/infinite-tracing/build.gradle.kts
+++ b/infinite-tracing/build.gradle.kts
@@ -16,6 +16,7 @@ dependencies {
     implementation(project(":newrelic-api"))
 
     testImplementation("org.junit.jupiter:junit-jupiter-api:5.6.2")
+    testImplementation("org.junit.jupiter:junit-jupiter-params:5.6.2")
     testImplementation("org.mockito:mockito-junit-jupiter:3.3.3")
     testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:5.6.2")
 }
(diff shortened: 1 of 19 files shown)


jasonjkeller:
Whoever reviews this PR, make sure that it is in alignment with the agent spec: https://source.datanerd.us/agents/agent-specs/pull/597

kford-newrelic:
@toddwest thanks for the PR! We've got a number of things in our ToDo list that we have to take care of first but we're going to try and get this in this Jan-Mar quarter!

kford-newrelic:
@toddwest we had to bump this out to the Apr-Jun quarter since we're over-capacity for the current quarter

toddwest:
Thanks for the reminder @jasonjkeller. Went ahead and updated to bring the PR in line with the spec.

@kford-newrelic I'm happy to spend an hour walking through the PR with the team if that could help expedite the review process? I think a lot of customers would get some significant relief from these changes.

jtduffy (approved):

jtduffy:
- All PR tests passed from the branch I created from the fork.
- Tested infinite tracing successfully

PR #1148, opened 2023-02-06
Make sure TokenAndRefCount.token is never null

sdaubin:
### Overview

Code such as this expect that `TokenAndRefCount.token` is never null:

https://github.com/newrelic/newrelic-java-agent/blob/main/newrelic-agent/src/main/java/com/newrelic/agent/Transaction.java#L1403

This sets the `token` to the no-op implementation if a null token is passed in.  We could add some logging when the token is null since ignoring this case could mask an issue, but I'll leave that as a future enhancement.

### Related Github Issue

### Testing

### Checks


tbradellis (approved):
lgtm!

meiao (approved):

meiao:
Checks passed on https://github.com/newrelic/newrelic-java-agent/pull/1151/checks

PR #1154, opened 2023-02-16
Remove agent build code from each verify instrumentation

meiao:

### Overview
When running verify instrumentation, the first job is to build and cache the agent.
Then when the verify instrumentation task is run for each module, if the agent is not in the cache, it will build the agent.
But to build the agent there are some steps that are not dependent on the cache hit, so they execute every time.

This backup strategy is not necessary, as if the agent is not in the cache, likely the whole workflow has to be run again, or running that failed job should get the agent from the cache.

By removing these unnecessary steps, there is some decrease in requests to Github which could help prevent the 429s this job has been receiving.


tbradellis (approved):
lgtm

jtduffy (approved):

PR #1164, opened 2023-02-28
NEWRELIC-4037 Mongodb reactive instrumentation

jtduffy:
Instrumentation for MongoDB reactive, versions 4.2.0 up to but excluding 4.8.0.
Count operation is excluded from this instrumentation, but Aggregate is present. Without the exclusion, we would need another module to cover only [4.2.0-4.3.0)


instrumentation/mongodb-reactive-streams-4.2/src/main/java/com/nr/agent/mongo/MongoUtil.java
```diff
@@ -0,0 +1,156 @@
...
+
+    public static <T> SingleResultCallback<T> instrumentSingleResultCallback(SingleResultCallback<T> callback, String collectionName,
+            String operationName, String databaseName, String host) {
+        if (callback instanceof NRCallbackWrapper) {
+            return callback;
+        }
+
+        String hostname = null;
+        int port = DEFAULT_PORT;
+        String [] hostAndPort = host.split(":");
```
meiao, line 102: This split call will compile a Pattern everytime it is called.
To improve performance, it is better to do a static variable and use that to split here.

```
    private static final Pattern COLON = Pattern.compile(":");
    
    public static ... instrumentSingleResultCallback(...) {
      String[] hostAndPort = COLON.split(host);
    
    }
```

instrumentation/mongodb-reactive-streams-4.2/src/main/java/com/nr/agent/mongo/MongoUtil.java
```diff
@@ -0,0 +1,156 @@
...
+        wrapper.token = NewRelic.getAgent().getTransaction().getToken();
+        wrapper.segment = NewRelic.getAgent().getTransaction().startSegment(operationName);
+        return wrapper;
+    }
+
+    public static String determineBulkWriteOperation(WriteModel writeModel) {
+        String operationName = OP_DEFAULT;
+
+        if (writeModel != null) {
+            String [] tokens = writeModel.getClass().toString().split("\\.");
```
meiao, line 130: Same here.

instrumentation/mongodb-reactive-streams-4.2/src/main/java/com/nr/agent/mongo/NRCallbackWrapper.java
```diff
@@ -0,0 +1,54 @@
...
+    public Segment segment = null;
+    public DatastoreParameters params = null;
+    private SingleResultCallback<T> delegate = null;
+
+    public NRCallbackWrapper(SingleResultCallback<T> d) {
+        delegate = d;
+        if (isNotTransformed.get()) {
+            AgentBridge.instrumentation.retransformUninstrumentedClass(getClass());
+            isNotTransformed.set(false);
+        }
```
meiao, line 32: You might not need this logic if you change the package this class is in.

instrumentation/mongodb-reactive-streams-4.2/src/main/java/com/nr/agent/mongo/NRCallbackWrapper.java
```diff
@@ -0,0 +1,54 @@
...
+        // could be NoOpSegment if a txn wasn't started
+        if (segment != null) {
+            if (params != null) {
+                segment.reportAsExternal(params);
+            }
+            segment.end();
+        } else if (params != null) {
+            NewRelic.getAgent().getTracedMethod().reportAsExternal(params);
+        }
+        delegate.onResult(result, t);
```
meiao, line 52: The metric name could be derived from the delegate class. Maybe it would be more useful than our class name.

instrumentation/mongodb-reactive-streams-4.2/src/main/java/com/nr/agent/mongo/NRCallbackWrapper.java
meiao: For some reason, changing the package did not work as intended. It did instrument the class, but when the time came to link the token it said that no transaction was available, even though the running bytecode had the @Trace(async=true) annotation and the exit tracer was being created.

meiao (approved):

tbradellis:
related:
https://github.com/newrelic/newrelic-java-agent/pull/1172

Snapshot available for here in case anyone wants to run or test with this instrumentation.
https://github.com/newrelic/newrelic-java-agent/actions/runs/4356852735

PR #1171, opened 2023-03-02
NR-92387 Add toggle "newrelic.jdbchelper.vanilla_map" to revert map impl

jtduffy:
Temporary configuration option that will allow the agent to use a `ConcurrentHashMap` implementation rather than the `ExpiringValueMap` implementation for the urlToFactory and urlToDatabaseName maps. To revert to the `ConcurrentHashMap`, set a Java system property of `newrelic.jdbchelper.vanilla_map`.

Also renamed `ExpiringValueCache` to `ExpiringValueMap`.

This config option will be removed in a future Agent release.

### DIFF
--- a/agent-bridge-datastore/src/main/java/com/newrelic/agent/bridge/datastore/JdbcHelper.java
+++ b/agent-bridge-datastore/src/main/java/com/newrelic/agent/bridge/datastore/JdbcHelper.java
@@ -8,12 +8,12 @@
 package com.newrelic.agent.bridge.datastore;
 
 import com.newrelic.agent.bridge.AgentBridge;
+import com.newrelic.api.agent.NewRelic;
 
 import java.sql.Connection;
 import java.sql.DatabaseMetaData;
 import java.sql.Statement;
 import java.time.Duration;
-import java.time.Instant;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


agent-bridge-datastore/src/main/java/com/newrelic/agent/bridge/datastore/JdbcHelper.java
```diff
@@ -42,27 +42,42 @@ protected Boolean initialValue() {
...
     private static final Map<Connection, String> connectionToURL = AgentBridge.collectionFactory.createConcurrentWeakKeyedMap();
     public static final String UNKNOWN = "unknown";
 
+    // Config to toggle the use of the ExpiringValueCache Map implementation or vanilla ConcurrentHashMap implementation
+    // for the urlToFactory and urlToDatabaseName maps. The default is to use the ExpiringValueCache. Set a Java System
+    // property of newrelic.jdbchelper.vanilla_map=true to revert to the ConcurrentHashMap implementation.
+    private static final Map<String, ConnectionFactory> urlToFactory;
+    private static final Map<String, String> urlToDatabaseName;
+    static {
+        boolean useVanillaMap = Boolean.getBoolean("newrelic.jdbchelper.vanilla_map");
```
kford-newrelic, line 67: Is the naming of this sysprop generally in alignment with other naming conventions?

jtduffy: Looking at it now, I think it should be `newrelic.config.jdbchelper.vanilla_map`. That would better fit the established pattern we have.

agent-bridge-datastore/src/main/java/com/newrelic/agent/bridge/datastore/JdbcHelper.java
```diff
@@ -42,27 +42,42 @@ protected Boolean initialValue() {
 
     /**
      * Any values in the urlToFactory or urlToDatabaseName Maps older than this value will
-     * be evicted from the cache when the {@link ExpiringValueCache} timer fires. This
+     * be evicted from the cache when the {@link ExpiringValueMap} timer fires. This
      * will also be used as the timer interval value.
      */
```
meiao, line 47: I didn't catch this on the first review, but it is likely that this will be used by most instances of the map.
So maybe a static factory for this would make this easier to use.

agent-bridge-datastore/src/main/java/com/newrelic/agent/bridge/datastore/JdbcHelper.java
jtduffy: That's fair.  Since this was going to be a one-off, temporary thing I just added it as such.
I'll revisit this so it aligns with standard config vars.

meiao, line 67: Can't you use the config harness here?

PR #1174, opened 2023-03-03
NR-92387 Add toggle "newrelic.jdbchelper.vanilla_map" to revert map impl

jtduffy:
Temporary configuration option that will allow the agent to use a `ConcurrentHashMap` implementation rather than the `ExpiringValueMap` implementation for the urlToFactory and urlToDatabaseName maps. To revert to the `ConcurrentHashMap`, set a top level agent config of `use_jdbchelper_vanilla_map`.

Also renamed `ExpiringValueCache` to `ExpiringValueMap`. It now extends the `AbstractMap` class so that we only need variables of type `Map` that can be assigned to either the `ConcurrentHashMap` or `ExpiringValueMap` instances. Otherwise we would have unnecessary if blocks scattered in the methods that use these objects.

This config option will be removed in a future Agent release.

### DIFF
--- a/agent-bridge-datastore/src/main/java/com/newrelic/agent/bridge/datastore/JdbcHelper.java
+++ b/agent-bridge-datastore/src/main/java/com/newrelic/agent/bridge/datastore/JdbcHelper.java
@@ -60,16 +60,11 @@ public class JdbcHelper {
 
     // Config to toggle the use of the ExpiringValueCache Map implementation or vanilla ConcurrentHashMap implementation
     // for the urlToFactory and urlToDatabaseName maps. The default is to use the ExpiringValueCache. Set a config
-    // property of use_jdbchelper.vanilla_map=true to revert to the ConcurrentHashMap implementation.
+    // property of use_jdbchelper_vanilla_map=true to revert to the ConcurrentHashMap implementation.
     private static final Map<String, ConnectionFactory> urlToFactory;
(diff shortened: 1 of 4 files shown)


agent-bridge-datastore/src/main/java/com/newrelic/agent/bridge/datastore/JdbcHelper.java
```diff
@@ -42,27 +42,47 @@ protected Boolean initialValue() {
...
+    // for the urlToFactory and urlToDatabaseName maps. The default is to use the ExpiringValueCache. Set a config
+    // property of use_jdbchelper.vanilla_map=true to revert to the ConcurrentHashMap implementation.
+    private static final Map<String, ConnectionFactory> urlToFactory;
+    private static final Map<String, String> urlToDatabaseName;
+    static {
+        boolean useVanillaMap = false;
+        try {
+            useVanillaMap = NewRelic.getAgent().getConfig().getValue("use_jdbchelper_vanilla_map");
+        } catch (Exception ignored) {
+            // Just in case the getValue call throws an NPE
```
meiao, line 71: You can use the getValue(String key, T defaultValue)
This way it will never return null and never call a NPE.

jtduffy: Corrected

meiao (approved):

Issue #1104, opened 2022-12-16
Java agent: Include configured labels as tags in forwarded application log events


chris-gilbert-2:
### Description
I use the Java agent, and would like additional static information (git branch/commit hash) that I have available in environment variables to be made available as a dimension in Log events. The labels configuration allows me to easily inject that information into events other than forwarded log events, where they are visible with `tags.` prefix, I would like those tags to be included in logs that are forwarded by the agent.

I note that @StanlieK contributed an enhancement to include MDC in log events (https://github.com/newrelic/newrelic-java-agent/pull/866) - this is one option but MDC isn't really the right tool here - the information I have is the same for all threads, and using MDC would mean ensuring that it gets set in every thread (http request handling, background jobs that fire etc) - that is hard to do reliably and doesn't feel right given that the value is fixed at runtime.

Another option is to use an external forwarding mechanism, and I understand that any arbitrary information can be added alongside the New Relic expected attributes, but I feel that it would be more straightforward to use the agent.

I may have missed another obvious way to achieve what I want - I am quite new to New Relic, so there may be an existing way to enhance forwarded application log events with arbitrary additional data that will be exposed as dimensions in the Log event, but I haven't managed to find it.
 
### Acceptance Criteria

The above PR demonstrates how to look up configuration, and then add additional data to log events that come from log4j and logback forwarded data. 

Intention would be to use a similar approach to add any key value pairs that are specified in the top level labels configuration to forwarded log events with a `tags.` prefix for consistency with other events. I'd be happy to raise a PR along those lines.

### Design Consideration/Limitations
This feels like a consistent approach. I don't know if it could/should be applied to other agents, but it feels like it is in line with the previous accepted PR, as well as being aligned with the intent and usage of existing configuration.

I'm not sure if it should also be possible to switch off or explicitly enable it under the `application_logging` stanza to avoid any impact on existing users.

christopher-gilbert:
Hi, I had a couple of days off and made some changes on a fork, just applying to Logback for the moment - the outcome is as I'd expect - labels configuration is available as metadata when looking at forwarded logs; and if I query the Log table, the structure is the same as other entities such as Transaction which pick up labels config and interpret it as tags (I can add some screenshots after checking that I have obscured work related data).

[...]

PR #1177, opened 2023-03-10
Feature/add tags to forwarded logs

christopher-gilbert:
Full details of the changes are at https://github.com/newrelic/newrelic-java-agent/issues/1104 - important points about the change and approach are repeated here:

I use the Java agent, and would like additional static information (git branch/commit hash) that I have available in environment variables to be made available as a dimension in Log events. The labels configuration allows me to easily inject that information into events other than forwarded log events, where they are visible with tags. prefix, I would like those tags to be included in logs that are forwarded by the agent.

I note that @StanlieK contributed an enhancement to include MDC in log events (https://github.com/newrelic/newrelic-java-agent/pull/866) - this is one option but MDC isn't really the right tool here - the information I have is the same for all threads, and using MDC would mean ensuring that it gets set in every thread (http request handling, background jobs that fire etc) - that is hard to do reliably and doesn't feel right given that the value is fixed at runtime.

The rationale for the changes I made is as follows:

I note that agent-bridge can only access config via the flattened properties rather than, say, accessing LabelsConfig and so the approach I took is to extend the existing flattening approach, which currently will parse Strings into Boolean or Integer. My changes preserve the existing behaviour, and extend it to cover `List<String>` and `Map<String, String>` using the existing parsing logic used within specific config types.

I have extracted map parsing from `LabelsConfigImpl` into `BaseConfig` so that List and Map parsing are now in the same place, and to avoid applying labels specific validation to any other arbitrary 'String as Map' data - the parsing approach could be extended to pass validation rules in, or define a static map of keys -> validators but I decided to keep things simple.

In terms of preserving existing behaviour, although the flattening doesn't apply any validation, the config is still also parsed into `LabelsConfigImpl` and so labels are still validated (and all existing unit tests pass). The only behaviour change I have made is to trim the keys and values BEFORE the length is validated - previously trimming whitespace happened afterwards - I think it is more correct to validate the value that is actually sent by the agent - at the moment we might see a warning that labels have been truncated to 255 characters but the value could be shorter if the 255 included leading/trailing whitespace. I can amend this though if needed to avoid any change.

Another point to note is that only the tags that are defined in labels config are added to forwarded Log metadata, none of the tags that are added by the agent (eg instrumentation.name, instrumentation.provider etc) are included so it means forwarded logs aren't quite consistent with other entities, but looking at the code I couldn't see those tags being applied within the agent so I'm wondering if they are added at the server end? The config tags are fine for our use case but if there is a need to garnish forwarded logs with other tags, I'd need a steer on where that information is added to other entities.


newrelic-agent/src/main/java/com/newrelic/agent/config/AgentConfigImpl.java
```diff
@@ -599,29 +599,19 @@ public <T> T getValue(String path) {
...
         if (value == null) {
             value = flattenedProperties.get(path);
         }
-
         if (value == null) {
             return defaultValue;
         } else if (value instanceof ServerProp) {
             value = ((ServerProp) value).getValue();
             return castValue(path, value, defaultValue);
-        } else if (value instanceof String && defaultValue instanceof Boolean) {
```
christopher-gilbert, line 614: This coercion to Boolean or Integer has been shifted out to the ParsedConfigValue class

newrelic-agent/src/main/java/com/newrelic/agent/config/BaseConfig.java
```diff
@@ -239,6 +248,30 @@ public static List<String> getUniqueStringsFromString(String valuesString, Strin
...
+     * Interprets a {@link String} as a {@link Map} using a format of 'key1:value1;key2:value2', with
+     * resulting String keys and values trimmed of whitespace.
+     * <p>
+     * Each parsed Map entry is passed to the consumer function, which may for example add the entry to
+     * a Map, or carry out validation.
+     * <p>
+     * If the String format cannot be parsed as a sequence of Map entries, or if the consumer throws a
+     * {@link ParseException} then a {@link ParseException} will be thrown.
+     */
+    public static void parseMapEntriesFromString(String mapAsString, MapParsingBiConsumer<String, String> parsedEntryHandler) throws
```
christopher-gilbert, line 261: This was in `LabelsConfigImpl` - I've shifted it up to this base class as a static method to align with the existing similar methods in this class eg `getUniqueStringsFromString`. 

This means that this logic is available to the flattened config access as well as the original typesafe style Configs (LabelsConfig etc)

However, I have taken out labels config specific validation, and just allowed it and other users, to pass in their own handler to do what they want with the parsed map entries.

newrelic-agent/src/main/java/com/newrelic/agent/config/ParsedConfigValue.java
```diff
@@ -0,0 +1,105 @@
...
+public class ParsedConfigValue {
+
+    private static final Pattern MAP_TYPE_VALUE = Pattern.compile(".+:.+");
+    private final Object originalValue;
+    private List<String> valueAsList;
+    private Map<String, String> valueAsMap;
+    private Boolean valueAsBoolean;
+    private Integer valueAsInteger;
+
+    public ParsedConfigValue(Object value) {
```
christopher-gilbert, line 47: Really this is just pulling together existing functionality into one place. Basic idea is that the value is interpreted in as many ways as possible at time of construction, and can then be retrieved as any of the valid types by specifying a default of the required type

newrelic-agent/src/test/java/com/newrelic/agent/config/LabelsConfigImplTest.java
```diff
@@ -140,6 +140,9 @@ public void testInvalidString() {
 
         config = new LabelsConfigImpl("foo:bar;;many:delimiters");
         assertTrue(config.getLabels().isEmpty());
+
+        config = new LabelsConfigImpl("foo:bar;empty:    ");
```
christopher-gilbert, line 144: Just a missing test case added in

kford-newrelic:
@christopher-gilbert thanks for this PR! We intend to review/test this during our Apr-Jun quarter's roadmap work

kford-newrelic:
@chris-gilbert-2 our engineers thought your suggestion was a good one - but they felt that it should be considered for standardization in our agent specification (a good thing; and subsequently implemented across all our APM agents). We're going to trigger those discussions during this Apr-Jun quarter but because that takes time to get buy in from other engineers, we'll have to keep your PR on hold for the moment.

PR #1178, opened 2023-03-14
Updated DockerData to use a single regex to determine container IDs.

jbedell-newrelic:
Added pertinent test for NR-77066.
Removed unnecessary test for a missing ".scope"


jtduffy (approved):

tbradellis (approved):
lgtm

PR #1180, opened 2023-03-15
DO NOT MERGE - Incresing AIT timeout

meiao:
This is a test to check how the AITs will fare with an increased timeout.


kford-newrelic:
@meiao I assume that with the `DO NOT MERGE` still here, that this PR isn't ready yet, correct?

meiao:
This PR should never be merged. It only makes the AITs checkout a different branch, which has an increased timeout.
It also enables all the suppressed AITs, which did not benefit from the increased timeout.

PR #1181, opened 2023-03-16
NEWRELIC-5420 Webflux 6 instrumentation

jtduffy:
Based on the Webflux 5 instrumentation module with a few changes for visibility modifiers and method signatures.


instrumentation/spring-webflux-6.0.0/build.gradle
```diff
@@ -0,0 +1,38 @@
...
+}
+
+java {
+    toolchain {
+        languageVersion.set(JavaLanguageVersion.of(17))
+    }
+}
+
+test {
+    // These instrumentation tests only run on Java 17 regardless of the -PtestN gradle property that is set.
```
tbradellis, line 29: We need these to run on 17, 19+, so probably need to use the inverse logic or list them right?

tbradellis (approved):
looks good!

PR #1183, opened 2023-03-17
NR-99103 Supply a factory to make cleanup thread a daemon thread.

jtduffy:
Supply a factory to make cleanup thread a daemon thread.


meiao (approved):

tbradellis (approved):
+1

meiao (approved):

Issue #1056, opened 2022-10-26
Java Agent creating too many NewRelicMetricsReporter threads when used with Kafka instrumentation


kford-newrelic:
Current agent design/implementation causes an unnecessary number of `NewRelicMetricsReporter` thread to be created when the agent is used with Kafka 3.x instrumentation. Refactoring this implementation would be best - perhaps to leverage a singleton instance, etc.

kford-newrelic:
Will consider for the Jan-Mar roadmap

kford-newrelic:
Targeted for the Jan-Mar quarter

obenkenobi:
A solution lives on the branch [NEWRELIC-4832-reduce-thread-count](https://github.com/newrelic/newrelic-java-agent/tree/NEWRELIC-4832-reduce-thread-count). This bug is put on hold for higher priority tasks.

PR #1187, opened 2023-03-20
Newrelic 4832 reduce thread count in our Kafka instrumentation

obenkenobi:
### Overview
Creates a single daemon thread for metrics reporters in our Kafka Streams instrumentation and a single daemon thread for Kafka Clients instrumentation. This fixes an issue where a single thread is created per a Kafka consumers/producer causing too many threads to be created.

### Related Github Issue
Github Issue Link: [https://github.com/newrelic/newrelic-java-agent/issues/1056](https://github.com/newrelic/newrelic-java-agent/issues/1056)


instrumentation/kafka-clients-metrics-3.0.0/src/main/java/com/nr/instrumentation/kafka/MetricsScheduler.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    public static void removeMetricsReporter(NewRelicMetricsReporter metricsReporter) {
+        ScheduledFuture<?> task = metricReporterTasks.remove(metricsReporter);
+        task.cancel(false);
+    }
+
+    private static ScheduledThreadPoolExecutor createScheduledExecutor() {
+        return (ScheduledThreadPoolExecutor)Executors.newScheduledThreadPool(
+                1, new DefaultThreadFactory("NewRelicMetricsReporter-Kafka", true));
```
jtduffy, line 47: To eliminate the `DefaultThreadFactory` class here and in the streams module, you could do what I did [here](https://github.com/newrelic/newrelic-java-agent/pull/1183/files) and supply an inline lambda for the thread factory.

```
Executors.newSingleThreadScheduledExecutor(runnable -> {
                final Thread thread = new Thread(runnable);
                thread.setDaemon(true);
                thread.setName(THREAD_NAME);
                return thread;
            }
```

jtduffy (approved):

instrumentation/kafka-clients-metrics-3.0.0/src/main/java/com/nr/instrumentation/kafka/MetricsScheduler.java
obenkenobi: I updated my solution to use the inline lambda

PR #1200, opened 2023-04-05
Instrumentation priority

meiao:
### Overview
Adding priority to instrumentation modules so the order in which they apply can be defined.


instrumentation/order-one/build.gradle
```diff
@@ -0,0 +1,9 @@
+dependencies {
+    implementation(project(":newrelic-api"))
+    implementation(project(":newrelic-weaver-api"))
+}
+
+jar {
+    manifest { attributes 'Implementation-Title': 'com.newrelic.instrumentation.order-one', 'Priority': 1 }
```
meiao, line 7: I created these 3 modules for manual testing. They should not be in the final commit.

The Priority attribute is the addition to allow the order of the instrumentation modules. If it is not set, 0 is assumed.

instrumentation/order-one/src/main/java/com/example/Order.java
```diff
@@ -0,0 +1,22 @@
...
+import com.newrelic.api.agent.NewRelic;
+import com.newrelic.api.agent.Segment;
+import com.newrelic.api.agent.Trace;
+import com.newrelic.api.agent.weaver.Weave;
+import com.newrelic.api.agent.weaver.Weaver;
+
+@Weave
+public class Order {
+    public void method() {
+        Segment segment = NewRelic.getAgent().getTransaction().startSegment("one");
```
meiao, line 12: created segments to see that the order is followed.

instrumentation/order-one/build.gradle
jasonjkeller: Nice. Setting the priority in the build file seems like a fairly elegant approach. My first thought was that you were doing the ordering based on the fully qualified module name, essentially assigning a higher priority for those modules that don't start with `com.newrelic.instrumentation.security`.

Regardless, the question I have for either approach is, how do we enforce modules being configured or named correctly to ensure that the ordering makes sense? My guess is that both teams will just have to adopt a process that guarantees that it's always done the same way, whether that's manual documented processes or maybe some kind of automated linter.

meiao: My idea was: the agent is always 0* and it is up to other modules to add their priority.
If the priority is not added to an external module, it is on them. We did provide the functionality.

* unless we do need to order our instrumentation modules

meiao: Also, to make it deterministic, if there are 2 modules that have the same priority, the order will be alphabetical.

And priorities could be negative, in case the external instrumentation module should wrap the agent's instrumentation.

jasonjkeller (approved):

tbradellis (approved):

PR #1210, opened 2023-04-17
retries on failed instrumentation test

tbradellis:
### Overview
Reduce developer toil.  We often have to repeatedly re-run tests by observing a failure, logging onto github. triggering a re-run.  Often, this rerun has to happen multiple times. This will retry up to 2 times if the first test, and subsequently the 2nd retry fails. This should reduce the number of times a developer has to manually trigger the job significantly.  Even if the 3 tries doesn't get past a false failure, then a second manual run still nets up to 6 runs.  
This also gives someone a good opportunity to intervene and make sure the failure is not a legit failure before manually kicking off the job after the 1st 3 tries.

*Note, this does not restart a workflow in the case where it is cancelled due to a timeout or some other cancelled condition. It only reruns if there is a test failure.  There's not a direct way to rerun a cancelled workflow.

### Checks

### DIFF
--- a/.github/workflows/Java-Instrumentation-Tests.yml
+++ b/.github/workflows/Java-Instrumentation-Tests.yml
@@ -143,7 +143,6 @@ jobs:
         id: run_tests_1
         continue-on-error: true
         env:
-
           JAVA_HOME: ${{ env.ORG_GRADLE_PROJECT_jdk8 }}
         run: |
           timeout 35m ./gradlew $GRADLE_OPTIONS --console=plain :instrumentation:test -Ptest${{ matrix.jre }} --continue
--- a/.github/workflows/Java-Instrumentation-Tests.yml
+++ b/.github/workflows/Java-Instrumentation-Tests.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         jre: [8, 11, 17, 19]
     name: Java ${{ matrix.jre }} Instrumentation Tests
-    timeout-minutes: 60
+    timeout-minutes: 120
     # needs: install-all-java
     runs-on: ubuntu-20.04
     env:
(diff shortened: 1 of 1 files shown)


meiao (approved):
This is interesting, also because the retrigger in the job will prevent redownloading everything and setting up the system again.
One thing though, we might need to check the timeouts. For example, if the tests take 25 minutes in average, and the timeout is 60 minutes, then there will never be enough time for the 3rd attempt to finish.

We might also want to set a step time out. This way the timeout for a single step will not take up to the timeout of the whole job.
There is no native GHA functionality for that, but it can be achieved by running:
```
        - name: Run instrumentation tests for Java ${{ matrix.jre }} w/o GE (attempt 2)
        id: run_tests_2
        continue-on-error: true
        if: steps.run_tests_1.outcome == 'failure'
        env:
          JAVA_HOME: ${{ env.ORG_GRADLE_PROJECT_jdk8 }}
        run: |
          timeout 30m ./gradlew $GRADLE_OPTIONS --console=plain :instrumentation:test -Ptest${{ matrix.jre }} --continue
```

.github/workflows/Java-Instrumentation-Tests.yml
```diff
@@ -139,7 +139,26 @@ jobs:
           rm gradle.properties
           mv gradle.properties.gha gradle.properties
 
-      - name: Run instrumentation tests for Java ${{ matrix.jre }} w/o GE
+      - name: Run instrumentation tests for Java ${{ matrix.jre }} w/o GE (attempt 1)
+        id: run_tests_1
+        continue-on-error: true
+        env:
+
```
jasonjkeller, line 146: Looks like an extra line snuck in here.

jasonjkeller (approved):
> *Note, this does not restart a workflow in the case where it is cancelled due to a timeout or some other cancelled condition. It only reruns if there is a test failure. There's not a direct way to rerun a cancelled workflow.

That's really a bummer since we see a good amount of timeouts.

Nonetheless, rerunning on failures is still a great improvement!

obenkenobi (approved):

meiao (approved):

PR #1212, opened 2023-04-19
retry scala instrumentation

tbradellis:
### Overview
This PR adds retry logic to the scala instrumentation GHA workflows.  It is essentially the same as what was added to the Java Instrumentation workflows. Timeouts are slightly different because the number of jobs and average finish times.

*This also removes the old github enterprise portion of the workflow.

Example run where first job failed and the retry succeeds. 
https://github.com/newrelic/newrelic-java-agent/actions/runs/4751877319/jobs/8441538352

### DIFF
--- a/.github/workflows/GHA-Scala-Instrumentation-Tests.yaml
+++ b/.github/workflows/GHA-Scala-Instrumentation-Tests.yaml
@@ -90,7 +90,7 @@ jobs:
       - name: Setup Gradle
         uses: gradle/gradle-build-action@v2
 
-      - name: Build newrelicJar w/o GE
+      - name: Build newrelicJar
         env:
           JAVA_HOME: ${{ env.ORG_GRADLE_PROJECT_jdk8 }}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/GHA-Scala-Instrumentation-Tests.yaml
```diff
@@ -92,27 +90,7 @@ jobs:
...
-          SCALA_VERSION: scala${{ matrix.scala }}
-          TEST_TYPE: instrumentation
-        run: |
-          # echo "JAVA_HOME=${ORG_GRADLE_PROJECT_jdk8}" >> $GITHUB_ENV
-          echo "REVIEW ANY NEW ITEMS IN WORKSPACE"
-          ls -la
-          ./gradlew clean jar --parallel
-          ls -la newrelic-agent/build/
-
       - name: Build newrelicJar w/o GE
```
meiao, line 93: you could remove the `w/o GE`

.github/workflows/GHA-Scala-Instrumentation-Tests.yaml
```diff
@@ -130,25 +108,33 @@ jobs:
...
-          JRE: jre${{ matrix.java-version }}
-          LANGUAGE: scala
-          SCALA_VERSION: scala${{ matrix.scala }}
-          TEST_TYPE: instrumentation
-        run: ./gradlew --console=plain --parallel :instrumentation:test -Ptest${{ matrix.java-version }} -PnoInstrumentation -Pscala-${{ matrix.scala }} --continue --scan -Dscan.uploadInBackground=false --build-cache
-
-      - name: Run instrumentation tests for Java ${{ matrix.java-version }} and scala ${{ matrix.scala }} w/o GE
-        if: ${{ env.GRADLE_KEY == '' }}
+        run: ./gradlew --console=plain --parallel :instrumentation:test -Ptest${{ matrix.java-version }} -PnoInstrumentation -Pscala-${{ matrix.scala }} --continue
+        timeout-minutes: 25
```
meiao, line 117: Nice, I didn't know there was a timeout on steps.
I'd just add it in the beginning, in most cases the run has multiple commands and the timeout at the bottom would look lost.

.github/workflows/GHA-Scala-Instrumentation-Tests.yaml
```diff
@@ -130,25 +108,33 @@ jobs:
...
+
+      - name: Run instrumentation tests for Java ${{ matrix.java-version }} and scala ${{ matrix.scala }} (attempt 2)
+        id: run_tests_2
+        if: steps.run_tests_1.outcome == 'failure'
+        continue-on-error: true
         env:
           JAVA_HOME: ${{ env.ORG_GRADLE_PROJECT_jdk8 }}
-        run: ./gradlew --console=plain --parallel :instrumentation:test -Ptest${{ matrix.java-version }} -PnoInstrumentation -Pscala-${{ matrix.scala }} --continue
+        run: |
+          sleep 60
```
meiao, line 126: why do you need this sleep?

tbradellis: I put the sleep in to give a pause for any potential transient problems that might have caused the failure.  Maybe unnecessary?

tbradellis: I didn't do that in the Java Inst. test, but I saw it used on an example of retries that gave that reason. I'm not tied to it if you don't think it's necessary.

meiao: I would do without it. If we see that many retries are failing, then we can introduce.

meiao (approved):

PR #1218, opened 2023-04-21
retry on unit tests

tbradellis:
Adds retry on Unit Tests.  (also sneaks in a minor tweak to the order of some steps)

### DIFF
--- a/.github/workflows/GHA-Functional-Tests.yaml
+++ b/.github/workflows/GHA-Functional-Tests.yaml
@@ -108,9 +108,9 @@ jobs:
           ls -la newrelic-agent/build/
 
       - name: Run functional tests on ${{ matrix.java-version }} (attempt 1)
-        timeout-minutes: 50
         id: run_tests_1
         continue-on-error: true
+        timeout-minutes: 50
         env:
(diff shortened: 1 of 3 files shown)


.github/workflows/GHA-Functional-Tests.yaml
```diff
@@ -108,9 +108,9 @@ jobs:
           ls -la newrelic-agent/build/
 
       - name: Run functional tests on ${{ matrix.java-version }} (attempt 1)
-        timeout-minutes: 50
         id: run_tests_1
         continue-on-error: true
+        timeout-minutes: 50
```
tbradellis, line 113: This was not in the same order as every other test, so just flipped it for consistency.

.github/workflows/Java-Instrumentation-Tests.yml
```diff
@@ -121,7 +121,7 @@ jobs:
 
       # Is newrelicJar present in newrelic-agent/build
       # TO DO: Below version number has to be dynamic
-      - name: Build newrelicJar w/o GE
+      - name: Build newrelicJar
```
tbradellis, line 124: Also slipping this change in because it is left over cruft from old workflows.

meiao (approved):

jtduffy (approved):

PR #1130, opened 2023-01-26
Instrument the Kafka bootstrap nodes

fryckbos:
### Overview
This PR adds a new metric that captures the Kafka bootstrap nodes.

### Testing
The agent includes a suite of tests which should be used to
verify your changes don't break existing functionality. These tests will run with
Github Actions when a pull request is made. More details on running the tests locally can be found
[here](https://github.com/newrelic/newrelic-java-agent/blob/main/CONTRIBUTING.md),

### Checks

[x] Are your contributions backwards compatible with relevant frameworks and APIs? Yes
[x] Does your code contain any breaking changes? No
[x] Does your code introduce any new dependencies? No

PR #1223, opened 2023-04-26
Add new Nodes metrics that include the consumed/produced topics

javimb:
### Overview

These changes are a second iteration of [PR#1130](https://github.com/newrelic/newrelic-java-agent/pull/1130).
Since a service could have multiple Kafka consumers and producers, and each one could connect to different clusters, we want to introduce these new metrics to understand the relationship between consumers/producers, topics and nodes.

### Testing

Although `Kafka3MessageTest` is disabled because it's flaky on GHA, I added some more checks around the metrics I introduced and I checked that the tests pass locally.

### Checks

✅ Are your contributions backwards compatible with relevant frameworks and APIs? Yes
✅ Does your code contain any breaking changes? No
✅ Does your code introduce any new dependencies? No

### DIFF
--- a/instrumentation/kafka-clients-metrics-3.0.0/src/main/java/com/nr/instrumentation/kafka/MetricsScheduler.java
+++ b/instrumentation/kafka-clients-metrics-3.0.0/src/main/java/com/nr/instrumentation/kafka/MetricsScheduler.java
@@ -77,15 +77,15 @@ public class MetricsScheduler {
                         }
                     }
                 }
-                if (METRICS_AS_EVENTS) {
-                    for (NewRelicMetricsReporter.NodeMetricName nodeMetricName : nrMetricsReporter.getNodes().values()) {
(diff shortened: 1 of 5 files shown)


kford-newrelic:
@javimb Thanks for the PR - we'll be reviewing!

meiao:
I've converted a few `Iterable.forEach` calls with loops.
forEach calls will force an extra method call and the creation of a lambda.
These are usually negligible for an application, but we try to keep the agent as lean as possible.

There is another point of improvement which would be not to store both metric and event names, since only one of those is ever used.
This requires more code changes and possibly more CPU overhead. And the decrease in memory may not be meaningful. So just leaving this note here and in the future that can be revisited.

Testing is done on: https://github.com/newrelic/newrelic-java-agent/actions/runs/5524205746

meiao (approved):

PR #1228, opened 2023-05-01
Use RollingFileAppender only when log_limit_in_kbytes greater than 0

jtduffy:
This corrects the agent logger agent config so that a RollingFileAppender is only created if the setting `log_limit_in_kbytes` is > 0. Before, if the `log_limit_in_kbytes` was 0 but `log_file_count` was > 1, a RollingFileAppender was created, but with a NoOp trigger policy, no rolling ever actually occurred.  

Created unit tests for these scenarios.

The docs will be clarified with the follow info:
- if `log_daily` is true, it will use the `file_count value`, with a minimum value of 1; if `log_limit_in_kbytes` > 0, a composite trigger policy will be configured: daily roll and size based roll
- if `log_limit_in_kbytes` > 0, a rolling file appender strategy will be configured, using the configured `log_file_count` (minimum of 1)
- Otherwise, no log rolling strategy will be configured

This is a non-breaking change, since any environment with `log_limit_in_kbytes` == 0 and `log_file_limit` > 1 will result in a regular, non-rolling file appender, which is equivalent to prior behavior.

### DIFF
--- a/newrelic-agent/src/test/java/com/newrelic/agent/logging/FileAppenderFactoryTest.java
+++ b/newrelic-agent/src/test/java/com/newrelic/agent/logging/FileAppenderFactoryTest.java
@@ -1,0 +1,92 @@
+package com.newrelic.agent.logging;
+
+import org.apache.logging.log4j.core.appender.AbstractOutputStreamAppender;
+import org.apache.logging.log4j.core.appender.FileAppender;
+import org.apache.logging.log4j.core.appender.FileManager;
+import org.apache.logging.log4j.core.appender.RollingFileAppender;
+import org.apache.logging.log4j.core.appender.rolling.CompositeTriggeringPolicy;
+import org.apache.logging.log4j.core.appender.rolling.CronTriggeringPolicy;
+import org.apache.logging.log4j.core.appender.rolling.NoOpTriggeringPolicy;
+import org.apache.logging.log4j.core.appender.rolling.SizeBasedTriggeringPolicy;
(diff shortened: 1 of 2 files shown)


tbradellis (approved):

newrelic-agent/src/test/java/com/newrelic/agent/logging/FileAppenderFactoryTest.java
```diff
@@ -0,0 +1,92 @@
...
+    public void build_withLogDailyFalseAndByteLimitEqualZero_createsSizeBasedRollingFileAppender() {
+        FileAppenderFactory factory = new FileAppenderFactory(1, 0, FILENAME + "3", false);
+        AbstractOutputStreamAppender<? extends FileManager> fileAppender = factory.build();
+        FileAppender rollingFileAppender = (FileAppender) fileAppender;
+
+        Assert.assertEquals(FileAppender.class, fileAppender.getClass());
+        Assert.assertEquals(FILENAME + "3", rollingFileAppender.getFileName());
+
+        try {
+            System.out.println(new File(FILENAME + "0").getCanonicalPath());
```
meiao, line 80: Should these be asserts?
Or is this just debug code?

If not debug code and these throw an exception, it should fail the test.

jtduffy: It was debug I forgot to remove...

meiao (approved):

PR #1230, opened 2023-05-02
GHA workflow documentation

meiao:
### Overview
Adding information about each workflow in the .github/workflow readme.

### DIFF
--- a/.github/workflows/TestSuite-Release.yml
+++ b/.github/workflows/TestSuite-Release.yml
@@ -1,8 +1,5 @@
 name: Test Suite - Release
 on:
-  release:
-    types:
-      - published
   workflow_dispatch:
     inputs:
       agent-ref:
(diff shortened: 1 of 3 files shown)


.github/workflows/TestSuite-Release.yml
```diff
@@ -1,8 +1,5 @@
 name: Test Suite - Release
 on:
-  release:
-    types:
-      - published
```
meiao, line 5: When the release is already published is a little late to run this workflow.
We have to run it manually to make sure that the release is all good.

.github/workflows/VerifyInstrumentation-Single.yml
```diff
@@ -1,5 +1,3 @@
-# Reusable workflow to run Verify Instrumentation on a single module.
-
```
meiao, line 2: This is not really a reusable workflow.

meiao:
That is not controlled by the workflow. All manually runnable workflows have that feature.

PR #1248, opened 2023-05-09
NR-115333 - Agent API changes for csec

meiao:
### Overview
Adds methods to the API that are required by the csec agent.


newrelic-agent/src/main/java/com/newrelic/agent/Transaction.java
```diff
@@ -2562,4 +2562,8 @@ int getCountOfRunningAndFinishedTransactionActivities() {
             return runningChildren.size() + finishedChildren.size();
         }
     }
+
+    public Object getSecurityMetaData() {
+        return null;
```
meiao, line 2567: This differs a little from the implementation on the k2-integration branch.
In there this method returns a SecurityMetaData. But it would require that we have a dependency on the csec api.
So for now, this implementation only returns null. Since nothing uses this method yet, this should be fine.

jasonjkeller (approved):

PR #1261, opened 2023-05-10
NR-106006 disable userid when high security

obenkenobi:
Disabled user tracking when high security mode is enabled

### DIFF
--- a/instrumentation/servlet-user/src/main/java/javax/servlet/Servlet.java
+++ b/instrumentation/servlet-user/src/main/java/javax/servlet/Servlet.java
@@ -26,7 +26,7 @@ public abstract class Servlet {
         if (request instanceof HttpServletRequest) {
             Principal principal = ((HttpServletRequest) request).getUserPrincipal();
             if (principal != null) {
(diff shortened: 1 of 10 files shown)


instrumentation/servlet-user/src/main/java/javax/servlet/Servlet.java
```diff
@@ -26,7 +26,7 @@ public void service(ServletRequest request, ServletResponse response) {
         if (request instanceof HttpServletRequest) {
             Principal principal = ((HttpServletRequest) request).getUserPrincipal();
             if (principal != null) {
-                AgentBridge.getAgent().getTransaction().getAgentAttributes().put("user", principal.getName());
+                NewRelic.setUserName(principal.getName());
```
obenkenobi, line 29: This does the same thing but adds additional useful logs and a supportability metric, `SetUserName`

jtduffy (approved):

instrumentation/servlet-user/src/main/java/javax/servlet/Servlet.java
meiao: didn't this use a different parameter name?

meiao (approved):

PR #1265, opened 2023-05-11
NR-106006 fix how user attribute is set in servlet-user from previous PR

obenkenobi:
Fixes a problem that is missed in the previous PR for the ticket because GHA was not working so tests did not run. The "user" attribute was changed to a custom attribute but I reverted it to be an agent attribute.


jbedell-newrelic (commented):
Looks like you've got some leftover java.util.logging.Level imports.

jtduffy (approved):

PR #1267, opened 2023-05-12
Update the default config file for K2 settings

kford-newrelic:


newrelic-agent/src/main/resources/newrelic.yml
```diff
@@ -380,34 +380,31 @@ common: &default_settings
...
+    agent:
+      enabled: false
+
+    # These are the category of csec security events that can be disabled from generating.
+    detection:
+      rci:
+        enabled: true
+      rxss:
+        enabled: true
+    deserialization:
```
jasonjkeller, line 406: I believe `deserialization` should be indented under the `detection` stanza, not on the same level as it.

jasonjkeller (changes requested):

newrelic-agent/src/main/resources/newrelic.yml
kford-newrelic: Thank you - corrected!

jasonjkeller (approved):

PR #1272, opened 2023-05-16
Java 8 improvements

jeffalder:

### Overview
Three improvements related to Java 8 functionality.

1. *Lambdas*. The agent cannot use lambdas in instrumentation modules due to relocation/naming issues that have never been solved in the weaver. However, the rest of the agent is "just code" and shouldn't have issues using lambdas or method references.
2. *Comparators*. A few places define Comparators using anonymous classes when a helper could be combined with a lambda.
3. *Less Guava*. Remove unnecessary Guava collection constructors. Before the advent of the diamond operator, Guava's collection constructors like `Maps.newHashMap` allowed users to have a shorter constructor call. Now, many of these constructors are unnecessary. They're only an advantage in the calls where you pass in an expected number of elements and they expand the expectation by a factor to reduce resizes.

### Related Github Issue

### Testing

### Checks

[x] Does your code contain any breaking changes? - No
[x] Does your code introduce any new dependencies? - No


meiao (approved):

meiao:
Running tests on https://github.com/newrelic/newrelic-java-agent/actions/runs/4995173280

PR #1274, opened 2023-05-17
Preload Security used classes to avoid complete application thread bl…

lovesh-ap:
### Overview
Preload Security Used Classes

### Related Github Issue
https://issues.newrelic.com/browse/NR-103247
Preload Security used classes to avoid complete application thread blocking in rare scenarios.


newrelic-agent/src/main/java/com/newrelic/agent/instrumentation/ClassTransformerServiceImpl.java
```diff
@@ -157,6 +159,10 @@ private PointCutClassTransformer startClassTransformer(InstrumentationProxy inst
         // java-io-stream.
         NewRelic.getAgent().getTransaction();
 
+        // Preload Security used classes to avoid complete application thread blocking in rare scenarios.
+        StringUtils.startsWithAny("abc", "z", "b", "a");
```
meiao, line 163: would just setting the empty string here work?

lovesh-ap: yes it will work, if you want i can change

meiao: Yes, please. I think it will look less hacky.

lovesh-ap:
@meiao We have updated CSEC version to 1.0.3-SNAPSHOT

meiao (approved):

PR #1277, opened 2023-05-18
Spotbugs corrections from NR-118554 and NR-118558

jtduffy:
Added `final` to certain variables:
- com.newrelic.agent.TransactionApiImpl.INSTANCE
- com.newrelic.agent.attributes.AgentAttributeSender.ATTRIBUTE_TYPE
- com.newrelic.agent.attributes.CustomAttributeSender.ATTRIBUTE_TYPE
- com.newrelic.agent.errors.ErrorMessageReplacer.NO_REPLACEMENT
- com.newrelic.agent.service.analytics.SpanEventFactory.DEFAULT_SYSTEM_TIMESTAMP_SUPPLIER
- com.newrelic.agent.tracers.AbstractTracer.ATTRIBUTE_TYPE
- com.newrelic.agent.tracers.jasper.GeneratorVisitTracerFactory.RUM_STATE_PROCESSOR_KEY
- com.newrelic.agent.utilization.GCP.PROVIDER

Other corrections:
-com.newrelic.agent.trace.RandomTransactionSampler.harvest(String)
  - Comparision of String param using == or != , line 60
- com.newrelic.agent.TransactionApiImpl.clearTransaction()
  - Null check of tx at line 477 of value previously de-ref'ed in clearTransaction(), line 481
- com.newrelic.agent.tracers.MethodExitTracerNoSkip(ClassMethodSignature, Transaction)
  - Null check of transaction at line 44 of value previously de-ref'ed in com.newrelic.agent.tracers.MethodExitTracerNoSkip(ClassMethodSignature, Transaction), line 44
- com.newrelic.agent.profile.ProfilerParameters.setProfilerFormat(String)
  - Return value of String.trim() ignored, line 94
- com.newrelic.agent.instrumentation.classmatchers.ManyClassMatcher.toString()
  - Invocation of toString(), line 75
- com.newrelic.agent.instrumentation.PointCutConfiguration.isEnabled()
  - Useless condition - it's known that classExplicitlyEnabled == false at this point, line 91

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/instrumentation/PointCutConfiguration.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/instrumentation/PointCutConfiguration.java
@@ -88,10 +88,6 @@ public class PointCutConfiguration {
 
         boolean groupExplicitlyDisabled = !pointCutGroupConfig.getProperty("enabled", true);
         boolean classExplicitlyDisabled = !pointCutConfig.getProperty("enabled", true);
-        if (groupExplicitlyDisabled && classExplicitlyEnabled) {
-            Agent.LOG.info(MessageFormat.format("Disabled point cut \"{1}\" (\"{2}\")",  getName(), getGroupName()));
-            return false;
-        }
         if (groupExplicitlyDisabled || classExplicitlyDisabled) {
             Agent.LOG.info(MessageFormat.format("Disabled point cut \"{1}\" (\"{2}\")",  getName(), getGroupName()));
             return false;
--- a/newrelic-agent/src/main/java/com/newrelic/agent/trace/RandomTransactionSampler.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/trace/RandomTransactionSampler.java
@@ -15,6 +15,7 @@ import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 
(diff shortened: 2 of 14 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/instrumentation/PointCutConfiguration.java
```diff
@@ -88,10 +88,6 @@ public boolean isEnabled() {
 
         boolean groupExplicitlyDisabled = !pointCutGroupConfig.getProperty("enabled", true);
         boolean classExplicitlyDisabled = !pointCutConfig.getProperty("enabled", true);
-        if (groupExplicitlyDisabled && classExplicitlyEnabled) {
```
jtduffy, line 91: This is always false because if `classExplicitlyEnabled` was `true`, the if on line 84 would eval to true and return

newrelic-agent/src/main/java/com/newrelic/agent/trace/RandomTransactionSampler.java
```diff
@@ -57,7 +58,7 @@ public List<TransactionTrace> harvest(String appName) {
         if (td == null) {
             return Collections.emptyList();
         }
-        if (td.getApplicationName() != appName) {
```
jtduffy, line 60: Do a null safe comparison with `Objects.equals`

jbedell-newrelic (approved):

newrelic-agent/src/main/java/com/newrelic/agent/trace/RandomTransactionSampler.java
meiao: There are some places where we do String comparison using `==` as a performance improvement.
In those cases, we make sure that we always use the same set of Strings so we know that if the Strings are going to be `equals`, then they are going to be the same reference.
Not a coding best practice, but it lessens the overhead of the agent.

PR #1286, opened 2023-05-25
Adding missing environment variable on Create Custom Jar workflow

meiao:
### Overview
The custom jar workflow was missing an environment variable which prevented the AWS credentials from being set.

### DIFF
--- a/.github/workflows/Create-Custom-Jar.yml
+++ b/.github/workflows/Create-Custom-Jar.yml
@@ -15,6 +15,9 @@ jobs:
   create_custom_jar:
     name: Create jar
     runs-on: ubuntu-20.04
+    env:
+      # we use this in env var for conditionals (secrets can't be used in conditionals)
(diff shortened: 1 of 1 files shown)


.github/workflows/Create-Custom-Jar.yml
```diff
@@ -15,6 +15,9 @@ jobs:
   create_custom_jar:
     name: Create jar
     runs-on: ubuntu-20.04
+    env:
+      # we use this in env var for conditionals (secrets can't be used in conditionals)
+      AWS_KEY: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
meiao, line 20: This is used in an if statement on the step that sets up the AWS credentials, if this is missing (or has no value) the credentials are not set.

jtduffy (approved):

tbradellis (approved):

PR #1315, opened 2023-06-14
implement getStatus for nrwrapper

tbradellis:
NRResponseWrapper class in our Servlet 5 and Servlet 6 instrumentation doesn't implement the getStatus method.
This came up in testing Jakarata 10 EE.  Testing shows that the http.StatusCode method was not being added to agent attributes (an assertion on attribute map length revealed the issue).

The method in the wrapper always returns 0 causing the put to the attributes to be skipped.
https://github.com/newrelic/newrelic-java-agent/blob/aeb1c4e0d8d0852909d8012fc92721c57ade1136/newrelic-agent/src/main/java/com/newrelic/agent/dispatchers/WebRequestDispatcher.java#L107

### DIFF
--- a/instrumentation/servlet-5.0/src/main/java/com/nr/instrumentation/servlet5/NRResponseWrapper.java
+++ b/instrumentation/servlet-5.0/src/main/java/com/nr/instrumentation/servlet5/NRResponseWrapper.java
@@ -20,14 +20,9 @@ public class NRResponseWrapper implements Response {
         this.response = response;
     }
 
-    /**
-     * Since {@link #getStatus()} does not exist prior to Servlet 3.0, return 0, meaning "unknown".
-     * 
-     * @since Servlet 3.0
-     */
     @Override
     public int getStatus() throws Exception {
-        return 0;
+        return response.getStatus();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


newrelic-agent/src/test/java/com/newrelic/agent/dispatchers/WebRequestDispatcherTest.java
```diff
@@ -161,4 +162,18 @@ private WebRequestDispatcher createDispatcher(MockHttpRequest httpRequest) throw
         return dispatcher;
     }
 
+    @Test
```
tbradellis, line 165: Just sneaking some coverage since I was looking at this class while debugging.

jtduffy (approved):

instrumentation/servlet-5.0/src/main/java/com/nr/instrumentation/servlet5/NRResponseWrapper.java
```diff
@@ -20,14 +20,9 @@ public NRResponseWrapper(HttpServletResponse response) {
...
 
-    /**
-     * Since {@link #getStatus()} does not exist prior to Servlet 3.0, return 0, meaning "unknown".
-     * 
-     * @since Servlet 3.0
-     */
     @Override
     public int getStatus() throws Exception {
-        return 0;
+        return response.getStatus();
```
meiao, line 25: Good find.

meiao (approved):

PR #1324, opened 2023-06-16
Kafka connect

meiao:
### Overview
Adds instrumentation for Kafka Connect.
There are 3 new instrumentation modules.
The first adds support for extracting metrics from Kafka Connect.
The other 2 add transaction tracing, one for versions 2.0.0 up to 3.2.3, and the other for versions 3.3.0 and later.


instrumentation/kafka-connect-metrics-1.0.0/src/main/java/com/nr/instrumentation/kafka/connect/KafkaConnectMetricsReporter.java
```diff
@@ -0,0 +1,176 @@
...
+                this::harvest, 0L, REPORTING_INTERVAL_IN_SECONDS, TimeUnit.SECONDS);
+    }
+
+    // default visibility for testing
+    void harvest() {
+        try {
+            Map<String, Object> eventData = METRICS_AS_EVENTS ? new HashMap<>() : Collections.emptyMap();
+            for (final Map.Entry<String, KafkaMetric> metric : metrics.entrySet()) {
+                Object metricValue = metric.getValue().metricValue();
+                if (metricValue instanceof Double) {
```
obenkenobi, line 78: Why not use `metricValue instanceof Number`? There could be missing metrics that cast to an integer.

obenkenobi: In addition you can then get your float value like this: `final float value = ((Number) metricValue).floatValue();`

instrumentation/kafka-connect-metrics-1.0.0/src/main/java/com/nr/instrumentation/kafka/connect/KafkaConnectMetricsReporter.java
```diff
@@ -0,0 +1,176 @@
...
+
+    private static final AtomicBoolean INITIALIZED = new AtomicBoolean(false);
+
+    // key is the metric/event name
+    private final Map<String, KafkaMetric> metrics = new ConcurrentHashMap<>();
+
+    private ScheduledFuture<?> scheduledFuture;
+
+    // the metrics are shared between Workers. So only one reporter needs to be attached.
+    public static void initialize(ConnectMetrics.MetricGroup metricGroup) {
```
obenkenobi, line 51: Nice. I'd like to refactor some of the other Kafka metrics reporters in streams/clients to use this approach.

obenkenobi (approved):

jtduffy (approved):
+1

instrumentation/kafka-connect-metrics-1.0.0/src/main/java/com/nr/instrumentation/kafka/connect/KafkaConnectMetricsReporter.java
meiao: This was copied from the original Kafka metrics instrumentation.
But that is a good change. I checked and apparently they do have integer metrics.

PR #1326, opened 2023-06-20
unit tests; update mockito dependency

jtduffy:
I updated the mockito dependency to use the "inline-mockito" rather than mockito-core to add the ability to mock static methods. I want to see if this breaks any unit tests.


jtduffy:
Closing for now to test mockito upgrade

PR #1351, opened 2023-06-23
Unit tests; update mockito dependency for java-agent module

jtduffy:
The dependency update required some method/package name updates in unrelated test classes.


jbedell-newrelic (approved):

jtduffy:
Mockito upgrade caused issues

PR #1406, opened 2023-07-21
Update CHANGELOG.md for 8.5.0

jtduffy:


CHANGELOG.md
```diff
@@ -4,6 +4,25 @@ Noteworthy changes to the agent are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Version 8.5.0
+## New features and improvements
+- Kafka Connect instrumentation: Adds metrics and transaction tracing for version 2.0.0+ [#1324](https://github.com/newrelic/newrelic-java-agent/pull/1324)
+- Security Agent: Limiting the supported version range for Apache log4j due to the new version release of Apache log4j on 21 June 2023 [#80](https://github.com/newrelic/csec-java-agent/pull/80/files)
```
kford-newrelic, line 10: I don't think we should be referencing "security agent", since we have a consolidated agent.

Since this is IAST related, what about changing this to "IAST :"? Similar to what we might do with a change for our DT feature?

PR #1408, opened 2023-07-21
Got Mockito-Inline working in the newrelic-agent module, we should be…

jbedell-newrelic:
Updated the mockito to mockito-inline 4.11 for the newrelic-agent module so we can use static mocks.


jtduffy (approved):
This is awesome...

PR #1410, opened 2023-07-24
Add instrumentation for r2dbc-postgresql-0.9.2

koenpunt:

### Overview

Adds updated instrumentation for r2dbc-postgresql-0.9.2, since instrumentation was disabled in https://github.com/newrelic/newrelic-java-agent/pull/999.

### Related Github Issue

### Testing

### Checks


koenpunt:
I tried running the tests locally, but couldn't get that to work, and was hoping to get the workflow running here on CI, but I see that maintainer requires approval.

koenpunt:
Okay, got the tests to run by not running gradle with the `--parallel` flag. Had to update the EmbeddedPostgres usage to assign a free port to not conflict with my local postgres installation.

meiao (approved):

meiao:
I've encountered some problems with segments not ending properly.
Investigating.

koenpunt:
> I've encountered some problems with segments not ending properly.

We've been experiencing the same, also the reason newrelic is currently unusable for, and thus has been disabled for the past month already.

meiao:
The segments that were not ended were related with the creation of Connections.
We will open a different ticket to investigate that.

@koenpunt did you encounter something different?

koenpunt:
We didn't investigate too much in what the problem was, also because we didn't really know where to look. What didn't help was the noise the token linker produces in the traces/profiling.

koenpunt:
Also; we've been using it with v1.0.0 of the r2dbc library, not sure if there it's because of the same reasons?

meiao:
The issue that I found is the same across all versions.

PR #1413, opened 2023-07-25
Add instrumentation for r2dbc-postgresql-1.0.0

koenpunt:

### Overview

Adds updated instrumentation for r2dbc-postgresql-1.0.0, since instrumentation was disabled in https://github.com/newrelic/newrelic-java-agent/pull/999.

Similar to #1410

~One caveat is that 1.0.x introduced support for multiple hosts, but AFAIK there no way to know which host is used at the point where the code is instrumented, so for now I defaulted to the first host in the configuration if there are multiple.~

Edit: I mimicked the implementation of the other r2dbc-*sql implementations to get the active socket address, so now the hostname should be correct at all times.

### Related Github Issue

### Testing

### Checks


meiao:
While testing this PR, it was found out that the instrumentation for 1.0.0 also worked for 0.9.2.
So the 1.0.0 was changed to be the 0.9.2 instrumentation and that will work up to the latest version.

koenpunt:
Hm, maybe because my initial approach used different apis in those versions, but when later updated the same components were used again..

instrumentation/r2dbc-postgresql-0.9.2/build.gradle
```diff
@@ -10,8 +10,8 @@ jar {
 }
 
 verifyInstrumentation {
-    passesOnly 'org.postgresql:r2dbc-postgresql:[0.9.2,1.0.0.RELEASE)'
-    excludeRegex(".*(M1|M2|RC).*")
+    passesOnly 'org.postgresql:r2dbc-postgresql:[0.9.2,)'
+    exclude 'org.postgresql:r2dbc-postgresql:1.0.0.RC1'
```
meiao, line 14: Something was odd with this release.
Removed the regex so if there is an incompatibility with any preview of 2.0.0 we will be notified then, instead of when the final version is released.

instrumentation/r2dbc-postgresql-0.9.0/src/test/java/com/nr/agent/instrumentation/r2dbc/PostgresqlNoInstrumentationTest.java
meiao, line 1: This was a test that ran without instrumentation. Not sure what was the intent of it.
Removing because it took 8 seconds.

meiao:
This PR was merged by #1556

PR #1422, opened 2023-07-28
Updated the getInputStream method on EmbeddedJars to just use the Jar…

jbedell-newrelic:
…File.getInputStream for a entry by name, rather than parsing thru them.

This Is fixing a problem where sometimes a JAR file in constructed in such a way that JarInputStream can't seem to process it correctly, possibly due to the order of the entries inside.  (Possibly related to this: https://bugs.openjdk.org/browse/JDK-8031748 )

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/service/module/EmbeddedJars.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/service/module/EmbeddedJars.java
@@ -47,7 +47,7 @@ public class EmbeddedJars {
                 String path = url.toExternalForm().substring(index + entry.getKey().length());
                 // add 1 to skip past the `.` and the value length, which is the length of the file extension
                 url = new URL(url.toExternalForm().substring(0, index + 1 + entry.getValue().length()));
-                // For some reason, some JAR files cannot be read properly by JarInputStream, at least the getNextJatEntry methodI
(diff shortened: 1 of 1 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/service/module/EmbeddedJars.java
```diff
@@ -46,36 +47,16 @@ public static InputStream getInputStream(URL url) throws IOException {
...
                 url = new URL(url.toExternalForm().substring(0, index + 1 + entry.getValue().length()));
-                InputStream inputStream = url.openStream();
-                JarInputStream jarStream = new JarInputStream(inputStream);
-                
-                if (!readToEntry(jarStream, path)) {
-                    inputStream.close();
-                    throw new IOException("Unable to open stream for " + path + " in " + url.toExternalForm());
-                }
-                return jarStream;
+                // For some reason, some JAR files cannot be read properly by JarInputStream, at least the getNextJatEntry methodI
```
meiao, line 50: Typo. getNextJarEntry

meiao (approved):
Nice find.

jtduffy (approved):

newrelic-agent/src/main/java/com/newrelic/agent/service/module/EmbeddedJars.java
jbedell-newrelic: fixed

PR #1466, opened 2023-08-29
Instrumentation of the Spring Cache abstraction layer

### LINKED ISSUE #1458: Implement spring-cache instrumentation
Add instrumentation for the spring-cache framework. This instrumentation will report metrics for cache hits, misses, evictions and clears. The underlying cache provider classname will be part of the metric name.

jtduffy:
Resolves #1458 

Adds counter metrics for cache hits/misses/evicts/clears for caches wrapped via the [Spring Cache](https://docs.spring.io/spring-framework/docs/4.1.x/spring-framework-reference/html/cache.html) library.  Supports Spring framework version 3.1.0 - current.

The format of the metrics will be:
`Spring/Cache/<cache-provider>/<cache-name>/[Hits/Misses/Clear/Evict]`

For example:
`Spring/Cache/java.util.ConcurrentHashMap/my_cache/Hits`


instrumentation/spring-cache-3.1.0/src/main/java/org/springframework/cache/Cache_Instrumentation.java
```diff
@@ -0,0 +1,42 @@
...
+import com.newrelic.api.agent.weaver.Weaver;
+import org.springframework.cache.Cache;
+
+import java.util.logging.Level;
+
+@Weave(type = MatchType.Interface, originalName = "org.springframework.cache.Cache")
+public abstract class Cache_Instrumentation {
+    public Cache.ValueWrapper get(Object key) {
+        Cache.ValueWrapper result = Weaver.callOriginal();
+        NewRelic.incrementCounter("Spring/Cache/" + getProviderClassName() + "/" + getName() + (result == null ? "/Misses" : "/Hits"));
```
meiao, line 16: I'd start the metric name with Cache, then Spring.
This way if we instrument other cache libraries, the metrics will all start with Cache, similar to Datastore or External.

instrumentation/spring-cache-3.1.0/src/main/java/org/springframework/cache/Cache_Instrumentation.java
```diff
@@ -0,0 +1,42 @@
...
+        return result;
+    }
+
+    public void evict(Object key) {
+        Weaver.callOriginal();
+        NewRelic.incrementCounter("Spring/Cache/" + getProviderClassName() + "/" + getName() + "/Evict");
```
meiao, line 22: The operation name usually starts with a lowercase character, probably because it is the method name, which starts with lowercase.

meiao (commented):
Good job.
It is a very simple instrumentation module, but delivers great value.

instrumentation/spring-cache-3.1.0/src/main/java/org/springframework/cache/Cache_Instrumentation.java
jtduffy: toLower'ed the operation names

meiao (approved):

dyorgio:
Hey guys, how to disable it to evict much data ingestion?

jtduffy:
@dyorgio You're welcome.

The configuration options for the agent are documented [here](https://docs.newrelic.com/docs/apm/agents/java-agent/configuration/java-agent-configuration-config-file/).  There are a couple that are missing, but we're in the process of updating the page.

You can also reference the [default yml file](https://download.newrelic.com/newrelic/java-agent/newrelic-agent/current/newrelic.yml) which has pretty good descriptions of most of the options available.

If you're looking to enable/disable other instrumentation modules, the name to use in the config option will match the module name under the `instrumentation` [folder](https://github.com/newrelic/newrelic-java-agent/tree/main/instrumentation) in the project.

Issue #1463, opened 2023-08-25
Create a workflow to run a single AIT


obenkenobi:
**Problem Statement**
The main way we run AITs in the Java Agent is running Test - AITs. This runs every AIT which can be time consuming if your focus is getting a single AIT to pass. 

**Proposed Solutiom**
The solution to this problem is to create a new workflow that only runs a single AIT.

obenkenobi:
Different workflow runs

All AITs
https://github.com/newrelic/newrelic-java-agent/actions/runs/6029454517

server/tomcat.py TomcatTest.test_tomcat
https://github.com/newrelic/newrelic-java-agent/actions/runs/6029457421

server/tomcat.py
https://github.com/newrelic/newrelic-java-agent/actions/runs/6029459740

A failing case so we know running a single AIT properly reports test failures
https://github.com/newrelic/newrelic-java-agent/actions/runs/6029471948

From the PR test suite so the workflow runs as intended
https://github.com/newrelic/newrelic-java-agent/actions/runs/6029663027

obenkenobi:
This has been accomplished by modifying the Test-AITs workflow to allow a new input parameter to run a single test

PR #1468, opened 2023-08-30
1463 run single ait

obenkenobi:
### Overview

Modify the Test - AITs workflow to add a new additional input to specify only a single test will run by specifying a specific file/script and possibly a test case in that test file.

### Related Github Issue
[1463](https://github.com/newrelic/newrelic-java-agent/issues/1463)

### DIFF
--- a/.github/workflows/Test-AITs.yml
+++ b/.github/workflows/Test-AITs.yml
@@ -15,9 +15,9 @@ on:
         description: "Specify cache branch/tag/sha (main is default)"
         required: false
         default: 'main'
-      test-case:
-        description: "Specify the test case. 
-        For example you can write 'server/tomcat.py' to run all tomcat tests or run 'server/tomcat.py TomcatTest.test_tomcat' to run a specific test case"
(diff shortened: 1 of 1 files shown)


.github/workflows/Test-AITs.yml
```diff
@@ -32,6 +37,12 @@ on:
         required: false
         default: 'main'
         type: string
+      single-test:
```
meiao, line 40: I don't think this input is needed in workflow_call.
Right now the use of workflow_call is to run all AITs in the test suite.

Do you see any use case for another workflow run a single AIT?
Or is this necessary because the input is referenced in the code?

meiao (commented):
Good idea to use conditional jobs.
What I had in mind was way more complicated.

.github/workflows/Test-AITs.yml
obenkenobi: I will remove it then

jbedell-newrelic (approved):

PR #1505, opened 2023-09-17
fix spring 6 pointcuts, after migration from javax to jakarta

sokolek:
I've observed that since migration to spring 6, the newrelic agent no longer sends error message when http requests fails. `SpringExceptionHandlerPointCut` used to fill error message, intercepting calls  to `org.springframework.web.servlet.DispatcherServlet#processHandlerException`.  Pointcut was defined to capture method with signature `"(Ljavax/servlet/http/HttpServletRequest;Ljavax/servlet/http/HttpServletResponse;Ljava/lang/Object;Ljava/lang/Exception;)Lorg/springframework/web/servlet/ModelAndView;"`.  Spring 6 moved to jakarta, so `jakarta` instead of `javax` is now used and method signature from `SpringExceptionHandlerPointCut` no longer captures `org.springframework.web.servlet.DispatcherServlet#processHandlerException`.

I've added new pointcuts reflecting new `jakarta` package names, used since Spring 6.0.


jtduffy:
Closed for duplicate #1544

PR #1529, opened 2023-09-29
Cgroupsv2 container

### LINKED ISSUE #1358: Support cgroup v2 for getting container ID
##  The issue 
In 2021 linux distributions rolled out cgroup v2.
This resulted in a change to how docker container ID is exposed.
In cgroup v1, we read `/proc/self/cgroup` and pattern match to find the container id.

https://github.com/newrelic/newrelic-java-agent/blob/e9f8ce7c2a16deb8873f2c0e9acff3f1c1dada41/newrelic-agent/src/main/java/com/newrelic/agent/utilization/DockerData.java#L32

This results in the agent not sending container id in the connect payload for utilization metrics on hosts using cgroup v2.

## Feature Description
The same information in cgroup c2 can be found in `/proc/self/mountinfo` through a similar pattern match.
The feature request is to support both cgroup v2 and cgroup v1
A reference can be found here with the open telemetry project:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/7167

## Describe Alternatives
There is an open issue on the open container initiative site to address this in a more generalized way so that changes like this do not break this as easily.
The main alternative would be to wait for this open issue to play out and use that solution.
https://github.com/opencontainers/runtime-spec/issues/1105

Because this is used for for utilization calculations, we need to move faster on implementing a solution. Also, there is no guarantee that a generalized abstraction will be implemented.

## Additional context

This information is used for utilization and is provided as part of the connect payload.  An agent-spec issue as also been created, as this impacts all language agents.

There's additional helpful information and links on this stackoverflow post:
https://stackoverflow.com/questions/68816329/how-to-get-docker-container-id-from-within-the-container-with-cgroup-v2

kanderson250:
Resolves #1358 

This PR adds support for getting the containerId from a docker container with Linux cgroup v2. 

Previously, the containerId was grabbed from `proc/self/cgroup` for containers with cgroup V1. Now, the id is grabbed from `proc/self/mountinfo`, with fallthrough to `proc/self/cgroup` if that file is missing.  

- This is the widely accepted approach for now, though the API is expected to change and a long-term solution is supposedly in the works. 
- The current implementation is based on best-effort research towards the contents of `proc/self/mountinfo`. Because this solution is a bit hacky, there isn't clear documentation about what's supposed to be in there in different environments/providers. The matching logic may need to be updated in the future.

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/utilization/DockerData.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/utilization/DockerData.java
@@ -41,39 +41,33 @@ public class DockerData {
         if (isLinux) {
             //try to get the container id from the v2 location
             File containerIdFileV2 = new File(FILE_WITH_CONTAINER_ID_V2);
-            String idResultV2 = cgroupV2GetDockerIdFromFile(containerIdFileV2);
+            String idResultV2 = getDockerIdFromFile(containerIdFileV2, CGroup.V2);
             if (idResultV2 != null) {
                 return idResultV2;
             }
             //try to get container id from the v1 location
             File containerIdFileV1 = new File(FILE_WITH_CONTAINER_ID_V1);
-            return getDockerIdFromFile(containerIdFileV1);
+            return getDockerIdFromFile(containerIdFileV1, CGroup.V1);
         }
         return null;
     }
 
-    /*
-    THIS LOGIC IS NOT BEAUTIFUL.
-    IT SHOULD BE REFACTORED.
-     */
-
-    String cgroupV2GetDockerIdFromFile(File mountInfoFile) {
+    String getDockerIdFromFile(File mountInfoFile, CGroup cgroup) {
         if (mountInfoFile.exists() && mountInfoFile.canRead()) {
             try {
-                return cgroupV2ReadFile(new FileReader(mountInfoFile));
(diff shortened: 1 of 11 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/utilization/DockerData.java
```diff
@@ -115,6 +141,7 @@ private boolean validCpuLine(String segment) {
         return false;
     }
 
+    //Get the match for the capture group of a single-capture-group regex expression.
```
jtduffy, line 144: nit: For method level comments I would change to the standard Javadoc format
```
/**
 * blah blah
 */
```

newrelic-agent/src/main/java/com/newrelic/agent/utilization/DockerData.java
```diff
@@ -123,4 +150,74 @@ private boolean checkAndGetMatch(Pattern p, StringBuilder result, String segment
         }
         return false;
     }
+
```
jtduffy, line 153: Remove commented code

jtduffy (changes requested):
2 minor nits. Nice test coverage!

jtduffy (approved):

PR #1536, opened 2023-10-04
Add stall detection

### LINKED ISSUE #1527: Add stall detection to the NR java agent
### Description
Many years ago the java agent had a very useful stall detection feature.  The agent would notice transactions that started and had not finished after a certain amount of time.  It would then report this as an error with a stack trace for the thread on which the transaction was running.

When it detects a "stall", it will report an event with transaction details similar to those we report in `Transaction` events.  The stack trace will be reported in `code.stacktrace`.

Initially, configuration will include:
- Enable/disable flag for the new SlowTransactionService
- Configuration to control how often (or at all) previously reported slow transactions get reported

@jack-berg will be handling the implementation.  Working branch [here](https://github.com/newrelic/newrelic-java-agent/compare/main...jack-berg:newrelic-java-agent:prototype-slow-transactions?expand=1).

jack-berg:
Resolves #1527.

Some notes:

- Adds a feature which detects slow transactions and reports as `SlowTransaction` analytic events
- Implementation hooks into transaction listener to maintain map of open transactions. Before harvest, it scans the map of open transactions and identifies the slowest that has been open longer than a threshold. It extracts metadata about that transaction, notably including the stacktrace of the thread that initialized it, and emits an event.
- The implementation has a mechanism preventing the same transaction to be reported multiple times.
- Disabled by default
- Threshold is configurable, and defaults to 1s (i.e. any transaction open for longer than 1s is a candidate to be reported).


jack-berg:
The build failures appear to be test flakes?

newrelic-agent/src/main/java/com/newrelic/agent/service/slowtransactions/SlowTransactionService.java
```diff
@@ -0,0 +1,226 @@
...
+            insightsService.storeEvent(
+                    ServiceFactory.getRPMService().getApplicationName(),
+                    new CustomInsightsEvent(
+                            "SlowTransaction",
+                            System.currentTimeMillis(),
+                            attributes,
+                            DistributedTraceServiceImpl.nextTruncatedFloat()));
+            //insightsService.recordCustomEvent("SlowTransaction", attributes);
+        }
+        addReportedTransaction(guid);
```
sdaubin, line 174: Instead of tracking `previouslyReportedTransactions`, we could just remove the guid from `openTransactions` right?  That will also cover the case of only reporting a slow transaction once.

jack-berg: Thanks - that should be a nice simplification! Will push a commit with the update.

jtduffy:
Duplicated this PR in #1542.

Issue #1519, opened 2023-09-25
Add s12y metrics for setUserId API call


meiao:
### Description

To gauge use of the `NewRelic.setUserId` the agent should report a s12y metric every time the API is invoked.

PR #1540, opened 2023-10-05
add s12y metrics for setUserId API call

deleonenriqueta:
Issue #1519


meiao (approved):

newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
```diff
@@ -298,6 +298,7 @@ public void addCustomParameters(Map<String, Object> params) {
      */
     @Override
     public void setUserId(String userId) {
+        MetricNames.recordApiSupportabilityMetric(MetricNames.SUPPORTABILITY_API_SET_USER_ID);
```
jtduffy, line 301: I don't know the specifics of the ticket, but do we want to send the metric if the call is ignored due to high security being enabled? Or have another metric for when it _is_ ignored due to high security?

jtduffy (approved):

PR #1543, opened 2023-10-09
Additional transaction naming logging

### LINKED ISSUE #1530: Add logging messages to the agent
### Description
While troubleshooting a customer's issues, we added logging messages to a snapshot version of our agent that turned out to be helpful. For this issue, we want to review those log messages and where appropriate, add some of those messages permanently to the agent. 

### Acceptance Criteria
Additional, new logging messages are added to the agent and visible in agent log files when the right logging level it set.

Need to make sure that any new log messages added are useful and necessary.

### Dependencies
N/A

### Additional context
Snapshot changes are on [this branch](https://github.com/newrelic/newrelic-java-agent/tree/setTxnName-debugging)

jtduffy:
Resolves #1530 

Adds logging around code which sets a transaction name.

### DIFF
--- a/instrumentation/servlet-2.4/src/main/java/com/nr/instrumentation/servlet24/ServletHelper.java
+++ b/instrumentation/servlet-2.4/src/main/java/com/nr/instrumentation/servlet24/ServletHelper.java
@@ -184,7 +184,9 @@ public class ServletHelper {
             return false;
         }
         txName = normalizeServletName(txName, servlet);
-        NewRelic.getAgent().getLogger().log(Level.FINER, "Setting transaction name using servlet name: {0}", txName);
+        NewRelic.getAgent().getLogger().log(Level.FINEST, "Servlet24/ServletHelper::setTxNameUsingServletName: calling transaction.setTransactionName with " +
+                        "priority: SERVLET_NAME and override false, category: Servlet, txn {0}, ",
+                AgentBridge.getAgent().getTransaction().toString());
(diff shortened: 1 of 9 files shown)


instrumentation/servlet-2.4/src/main/java/com/nr/instrumentation/servlet24/ServletHelper.java
```diff
@@ -184,7 +184,9 @@ private static boolean setTxNameUsingServletName(ServletConfig servletConfig, Se
             return false;
         }
         txName = normalizeServletName(txName, servlet);
-        NewRelic.getAgent().getLogger().log(Level.FINER, "Setting transaction name using servlet name: {0}", txName);
+        NewRelic.getAgent().getLogger().log(Level.FINEST, "Servlet24/ServletHelper::setTxNameUsingServletName: calling transaction.setTransactionName with " +
```
meiao, line 187: Having the + to concatenate the strings will force the compiler to create a StringBuilder everytime this line is executed, even if it is not going to be logged.

I recommend either saving the String as a constant in the class, so the StringBuilder is only ever created once, or wrapping the log statement with a check to see if this is loggable.
```
if (NewRelic.getAgent().getLogger().isLoggable(Level.FINEST)) {
  // log statement
}
```

Pretty much all the other log statements could benefit from one of these. Either because of the +, or because of the call to String.join.

meiao (approved):

PR #1559, opened 2023-10-20
Log4j2 JsonLayout support

jasonjkeller:
This is a followup to a previous PR that was reverted to address verifier failures: https://github.com/newrelic/newrelic-java-agent/pull/1545

This PR re-introduces the Log4j2 `JsonLayout` support but splits Log4j2 instrumentation across two modules.

`apache-log4j-2.6` is a new instrumentation modules for older versions of Log4j2 (versions 2.6 - 2.10.0). It's simply a copy of the `apache-log4j-2 `(versions 2.11.0+) module but with a few tweaks required to avoid verifier failures with those older versions.


instrumentation/apache-log4j-2.6/build.gradle
meiao, line 1: It is a bit odd that the 2.6 instrumentation comes before the 2.

Any reason that changing the name of 2 to 2.11 and adding an alias would not be enough to keep backwards compatibility?

meiao (approved):

instrumentation/apache-log4j-2.6/build.gradle
jasonjkeller: I agree that the naming is a bit weird. I went back and forth on whether to change the name of the module to 2.11 or not but I wasn't sure if that was going to cause headaches when evaluating instrumentation usage with angler. 

I was also thinking that we might want to deprecate the log4j2 versions used in the 2.6 module and ultimately delete that module altogether, which would once again leave us with just the `apache-log4j-2` module in the future.

I don't feel strongly either way but I'm happy to change it if you do.

instrumentation/apache-log4j-2/src/main/java/org/apache/logging/log4j/core/LogEvent_Instrumentation.java
```diff
@@ -0,0 +1,80 @@
...
+public abstract class LogEvent_Instrumentation {
+
+    /*
+     * In cases where the LogEvent is sent to an AsyncAppender, getLinkingMetadata would get called on a new thread and the trace.id and span.id
+     * would be missing. To work around this we save the linking metadata on the LogEvent on the thread where it was created and use it later.
+     */
+    @NewField
+    public Map<String, String> agentLinkingMetadata = isApplicationLoggingLocalDecoratingEnabled() ? NewRelic.getAgent().getLinkingMetadata() : null;
+
+    public abstract LogEvent toImmutable();
```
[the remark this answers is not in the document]
jasonjkeller: I didn't think they were required but oddly when I removed them, leaving the `@NewField` as the only thing in the interface, the instrumentation no longer worked. I expected that abstract methods in the interface didn't matter unless we were explicitly weaving them but the agent behavior seemed to indicate otherwise. I don't fully understand why. If that wasn't the case then this instrumentation wouldn't have to be split into two modules.

instrumentation/apache-log4j-2.6/build.gradle
meiao: It is fine. Let's just add the 2.6 module to the list of deprecations then.

PR #1587, opened 2023-11-13
Provide a noop Agent implementation

sdaubin:

### Overview
Add an accessor for a no-op instance of the agent api.  This allows us to re-use many of the no-op api instances instead of having to redefine them in the bridge.

It also makes it easier to pass an instance of the agent into tests that don't use the full bytecode injecting agent.

Moving forward, as we add more methods to `Agent`, we will have less code to write in the bridge.

### Testing

### Checks

- [X ] Your contributions are backwards compatible with relevant frameworks and APIs.
- [X ] Your code does not contain any breaking changes. Otherwise please describe. 
- [X ] Your code does not introduce any new dependencies. Otherwise please describe.


jtduffy (approved):

sdaubin:
Unfortunately this is broken because of the K2 agent :(

Issue #781, opened 2022-03-11
#39 regression; env vars with sys prop syntax are prioritized


jeffalder:
## Description
I have an application where the _environment variable_ `newrelic.config.labels` is set _and_ `NEW_RELIC_LABELS` is also set. Despite the message that the env var `newrelic.config.labels` is not supported, it appears from the data to actually take _precedence_ over `NEW_RELIC_LABELS`. 

Yes, I'm 100% sure the environment variable with dots and lower-case is used despite it looking like a system property.
Yes, I know. 
No, it's not in my control.

## Expected Behavior
`NEW_RELIC_LABELS` takes precedence; `newrelic.config.labels` is ignored.

## [Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results
I'm using agent 7.4.3; I don't suspect you need more than that?

## Steps to Reproduce
See [test.zip](https://github.com/newrelic/newrelic-java-agent/files/8233163/test.zip)
- download the agent, drop it in the current directory
- run `./run.sh` - it will compile and run the app
- Look at the output

Environment:
```
    env.put("newrelic.config.labels", "dotted-labela:dotted-valuea");
    env.put("NEW_RELIC_LABELS", "caps-labelb:caps-valueb");
```

When I look at the connect payload through audit mode, I see this:
```
% jq '.[0].settings.labels' payload.json
"caps-labelb:caps-valueb"
% jq '.[0].labels' payload.json
[
  {
    "label_type": "dotted-labela",
    "label_value": "dotted-valuea"
  }
]
```

## Your Environment
Temurin Java 17 but repros on Azul Zulu 11 as well
Agent 7.4.3
Mac M1, Monterey 12.2.1

(Migrate to Jira)

kford-newrelic:
Not selected for next quarter's agent roadmap. Will consider for a future agent release.

PR #1598, opened 2023-11-19
No dotted env vars

jeffalder:

Resolves https://github.com/newrelic/newrelic-java-agent/issues/781

### Overview
Stop looking at env vars with dotted property names. The current warning log message is inaccurate and I have provided a repro case that illustrates it.

### Testing

*edit* I also simplified the special case of `NEW_RELIC_LOG`. Working through the single-entry maps made the code harder to understand. I think the current code is much better.

### Checks

- Your code **does** ~not~ contain breaking changes. See below.

### Breaking changes
This pull request removes behavior from the lower-case, dotted environment variable names (such as `newrelic.config.labels`). There are no changes to documented system property behavior (via `newrelic.config.`, server-side config, YAML, or environment variable (via `NEW_RELIC_`).

Customers relying on lower-case, dotted environment variables should switch to the standard upper-case, underscore names.

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/config/SystemPropertyProvider.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/config/SystemPropertyProvider.java
@@ -21,7 +21,6 @@ public class SystemPropertyProvider {
     // environment variables
     private static final String NEW_RELIC_PREFIX_ENV = "NEW_RELIC_";
     private static final String LOG_ENV = NEW_RELIC_PREFIX_ENV + "LOG";
-    private static final String LOG_FILE_NAME = AgentConfigImpl.SYSTEM_PROPERTY_ROOT + AgentConfigImpl.LOG_FILE_NAME;
     private static final String NEW_RELIC_SYSTEM_PROPERTY_ROOT = "newrelic.";
 
     private final Map<String, String> newRelicSystemProps;
@@ -87,7 +86,9 @@ public class SystemPropertyProvider {
             String envVar = entry.getKey();
             if (envVar.startsWith(NEW_RELIC_PREFIX_ENV)) {
                 if (envVar.equals(LOG_ENV)) {
-                    addPropertyWithoutSystemPropRoot(nrProps, LOG_FILE_NAME, entry.getValue());
+                    // The classic NEW_RELIC_LOG corresponds to the log_file_name property
+                    // so we use log_file_name here.
+                    nrProps.put(AgentConfigImpl.LOG_FILE_NAME, entry.getValue());
                 } else {
                     addPropertyWithoutEnvPrefix(nrProps, envVar.toLowerCase(), entry.getValue());
                 }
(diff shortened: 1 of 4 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/config/SystemPropertyProvider.java
```diff
@@ -104,9 +85,10 @@ private void addNewRelicEnvProperties(Map<String, Object> nrProps, Map<String, S
         for (Map.Entry<String, String> entry : allEnvVars.entrySet()) {
             String envVar = entry.getKey();
             if (envVar.startsWith(NEW_RELIC_PREFIX_ENV)) {
-                String envVarNameToReplace = envVarToSystemPropKeyMap.get(envVar);
-                if (envVarNameToReplace != null) {
-                    addPropertyWithoutSystemPropRoot(nrProps, envVarNameToReplace, entry.getValue());
+                if (envVar.equals(LOG_ENV)) {
```
jeffalder, line 88: This block of code has always had a sorting issue; if both `NEW_RELIC_LOG` and `NEW_RELIC_LOG_FILE_NAME` are present and different, one or the other will be chosen. I opted not to try and fix this.

newrelic-agent/src/main/java/com/newrelic/agent/config/SystemPropertyProvider.java
```diff
@@ -128,21 +110,6 @@ private void addPropertyWithoutSystemPropRoot(Map<String, Object> nrProps, Strin
...
-        }
-
-        return getenv(key);
-    }
-
-    private String getenv(String key) {
-        String val = environmentFacade.getenv(key);
-        if (val != null) {
-            return val;
-        }
```
jeffalder, line 143: This is the conditional that was retrieving the dotted name from the environment. The simplest change would be to remove this conditional alone.

newrelic-agent/src/main/java/com/newrelic/agent/config/SystemPropertyProvider.java
```diff
@@ -128,21 +110,6 @@ private void addPropertyWithoutSystemPropRoot(Map<String, Object> nrProps, Strin
     }
 
     public String getEnvironmentVariable(String key) {
-        String propVal = envVars.get(key);
```
jeffalder, line 131: This entry consisted of exactly one value: `envVars.put(LOG_FILE_NAME, getenv(LOG_ENV));`. I opted to remove this special case from here, and pull it up into `AgentConfigImpl`.

newrelic-agent/src/test/java/com/newrelic/agent/config/AgentConfigImplTest.java
```diff
@@ -1293,6 +1296,49 @@ public void getApplicationLoggingConfigSystemProperty() {
                 config.getTransactionTracerConfig().isEnabled());
     }
 
+    @Test
+    public void logFilePriorityOrder() {
```
jeffalder, line 1300: I sequenced the commits so that I verified that this test passed on `main` before layering the remaining changes on it.

newrelic-agent/src/test/java/com/newrelic/agent/config/BaseConfigTest.java
```diff
@@ -26,9 +28,10 @@ public class BaseConfigTest {
     // SystemProperties collections with our key value, so we need to be able to identify this magic test
     // key so we can set up and clean up using @Before and @After methods.
 
-    private static final String PREFIX = "BaseConfig";
```
jeffalder, line 29: I had to rewrite this test pretty significantly; I don't think this worked as intended with the `NEW_RELIC_` or `newrelic.config.` prefixes (particularly the regex-replace from `newrelic.config` to `new.relic`).

jasonjkeller (approved):

jasonjkeller:
Looks good but we're going to hold off on merging this PR at the moment given the breaking changes.

jeffalder:
@jasonjkeller I've rebased the PR.

PR #1606, opened 2023-11-27
Add work queue bound on LicenseException retries in RPMConnectionServ…

### LINKED ISSUE #1428: OOM on API Key change
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description
[NOTE]: # ( Describe the problem you're encountering. )
Running inside docker containers.

We recently had an API key rotation for the new relic agent on our non productive environments as a way to save costs by disabling monitoring and logging.

We suddenly found out that we were having constant OOMs. Troubleshooting a bit we found out that the culprit was the New Relic Agent.

Sadly I cannot share a reproducible project.

## [Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results
[NOTE]: # ( Provide any other relevant log data. )
[TIP]:  # ( Scrub logs and diagnostic information for sensitive information )

## Steps to Reproduce

Very little I can share but it's pretty standard setup.

``` 
## Dockerfile
RUN wget "http://download.newrelic.com/newrelic/java-agent/newrelic-agent/current/newrelic-java.zip"
RUN unzip newrelic-java.zip -d "/app"

##Bash Script that runs on entrypoint
if [ "$USE_NEW_RELIC" = "true" ]; then
        echo -n "Setting New Relic..."
        if [ -f /run/secrets/newrelic_license_key ]; then
            NEWRELIC_LICENSE_KEY=$(cat /run/secrets/newrelic_license_key)
            NEW_RELIC="-javaagent:/app/newrelic/newrelic.jar"
            NEW_RELIC=$NEW_RELIC" -Dnewrelic.environment="${ENVIRONMENT}
            NEW_RELIC=$NEW_RELIC" -Dnewrelic.config.license_key="${NEWRELIC_LICENSE_KEY}
            NEW_RELIC=$NEW_RELIC" -Dnewrelic.config.app_name="${ENVIRONMENT}_${SERVICE_NAME}
            echo "New Relic set."
        else
            echo "New Relic licence key is not set."
            exit 1
        fi
    fi
```

```
apk add openjdk8
jps #pid 547
jcmd 547 GC.class_histogram | less
/app # jstat -gcutil 547 250 10
  S0     S1     E      O      M     CCS    YGC     YGCT    FGC    FGCT     GCT
  0.00   0.00  99.68  99.99  97.34  95.57 164830 1507.445 113654 254514.054 256021.499
  0.00   0.00 100.00  99.99  97.34  95.57 164830 1507.445 113654 254514.054 256021.499
  0.00   0.00 100.00  99.99  97.34  95.57 164830 1507.445 113654 254514.054 256021.499
  0.00   0.00 100.00  99.99  97.34  95.57 164830 1507.445 113654 254514.054 256021.499
  0.00   0.00 100.00  99.99  97.34  95.57 164830 1507.445 113654 254514.054 256021.499
  0.00   0.00 100.00  99.99  97.34  95.57 164830 1507.445 113654 254514.054 256021.499
  0.00   0.00 100.00  99.99  97.34  95.57 164830 1507.445 113654 254514.054 256021.499
  0.00   0.00 100.00  99.99  97.34  95.57 164830 1507.445 113654 254514.054 256021.499
  0.00   0.00 100.00  99.99  97.34  95.57 164830 1507.445 113654 254514.054 256021.499
  0.00   0.00 100.00  99.99  97.34  95.57 164830 1507.445 113655 254516.281 256023.726
```

image

## Your Environment
Docker container using openjdk:8-jre-alpine openjdk:8-jdk-alpine
Current New Relic Agent version

kanderson250:
Fixes #1428 

This PR addresses `OutOfMemory` exceptions that were reported during rotation to an invalid license key. The cause of the memory leak was found to be excessive creation of `RPMConnectionTasks`. To mitigate the issue, the number of `RPMConnectionTasks` created during processing of a `LicenseException` is now bounded. 

- The `ScheduledExecutorService` is replaced with a `ScheduledThreadPoolExecutor`, in order to get access to the size of its work queue. 
- The `LicenseException` handler in the `doConnect` method of `RPMService` limits `reconnect` attempts when the work queue acquires a certain number of tasks. 
- The number of tasks is bounded at 5000. 

There is no way for the agent to recover from a bad license key, so although these mitigation steps will prevent a memory leak, a JVM restart is still required to get the agent back up.

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/rpm/RPMConnectionServiceImpl.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/rpm/RPMConnectionServiceImpl.java
@@ -18,8 +18,7 @@ import com.newrelic.agent.stats.StatsWorks;
 import com.newrelic.agent.util.DefaultThreadFactory;
 import com.newrelic.agent.util.SafeWrappers;
 
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
@@ -45,12 +44,13 @@ public class RPMConnectionServiceImpl extends AbstractService implements RPMConn
     public static final long APP_SERVER_PORT_TIMEOUT = 120L;
     // interval between connection attempts
     public static final long MIN_CONNECT_ATTEMPT_INTERVAL = 5L;
-    private final ScheduledExecutorService scheduledExecutor;
+    private static final int MAX_QUEUED_CONNECT_TASKS = 5000;
+    private final ScheduledThreadPoolExecutor scheduledExecutor;
 
     public RPMConnectionServiceImpl() {
         super(RPMConnectionService.class.getSimpleName());
         ThreadFactory threadFactory = new DefaultThreadFactory(RPM_CONNECTION_THREAD_NAME, true);
-        scheduledExecutor = Executors.newSingleThreadScheduledExecutor(threadFactory);
+        scheduledExecutor = new ScheduledThreadPoolExecutor(1, threadFactory);
     }
 
     @Override
@@ -97,6 +97,14 @@ public class RPMConnectionServiceImpl extends AbstractService implements RPMConn
      */
     public long getAppServerPortTimeout() {
         return APP_SERVER_PORT_TIMEOUT;
+    }
+
+    /**
+     * Checks whether tasks have begun piling up in the scheduled executor's work queue.
+     * Used to prevent repeated reconnect attempts in the event of a LicenseException.
+     */
+    public boolean shouldPreventNewConnectionTask(){
+        return scheduledExecutor.getQueue().size() > MAX_QUEUED_CONNECT_TASKS;
     }
 
     /**
--- a/newrelic-agent/src/main/java/com/newrelic/agent/rpm/RPMConnectionServiceImpl.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/rpm/RPMConnectionServiceImpl.java
@@ -19,6 +19,7 @@ import com.newrelic.agent.util.DefaultThreadFactory;
 import com.newrelic.agent.util.SafeWrappers;
 
 import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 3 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/rpm/RPMConnectionServiceImpl.java
```diff
@@ -45,12 +44,13 @@ public class RPMConnectionServiceImpl extends AbstractService implements RPMConn
     public static final long APP_SERVER_PORT_TIMEOUT = 120L;
     // interval between connection attempts
     public static final long MIN_CONNECT_ATTEMPT_INTERVAL = 5L;
-    private final ScheduledExecutorService scheduledExecutor;
+    private static final int MAX_QUEUED_CONNECT_TASKS = 5000;
+    private final ScheduledThreadPoolExecutor scheduledExecutor;
```
jasonjkeller, line 48: I would switch this back to `ScheduledExecutorService` so that wherever we use `scheduledExecutor` we are dealing with an interface rather than a specific implementation of that interface. In general, this will help prevent us from becoming too dependent on a specific implementation (with the one exception called out below) which will make it easier to change to a different implementation if we should need to do so at some point in the future. As unlikely as that may be it's a good practice to follow.

newrelic-agent/src/main/java/com/newrelic/agent/rpm/RPMConnectionServiceImpl.java
```diff
@@ -99,6 +99,14 @@ public long getAppServerPortTimeout() {
         return APP_SERVER_PORT_TIMEOUT;
     }
 
+    /**
+     * Checks whether tasks have begun piling up in the scheduled executor's work queue.
+     * Used to prevent repeated reconnect attempts in the event of a LicenseException.
+     */
+    public boolean shouldPreventNewConnectionTask(){
+        return scheduledExecutor.getQueue().size() > MAX_QUEUED_CONNECT_TASKS;
```
jasonjkeller, line 107: If `scheduledExecutor` is switched back to `ScheduledExecutorService` then we will need to cast it to `ScheduledThreadPoolExecutor` to get access to the `getQueue()` method. It does mean that we now are dependent on the `ScheduledThreadPoolExecutor` implementation in this method but doing the explicit casting only in this one place lessens the likely hood that it will spread.

```
    public boolean shouldPreventNewConnectionTask(){
        return ((ScheduledThreadPoolExecutor) scheduledExecutor).getQueue().size() > MAX_QUEUED_CONNECT_TASKS;
    }
```

jasonjkeller (changes requested):

newrelic-agent/src/main/java/com/newrelic/agent/rpm/RPMConnectionServiceImpl.java
jasonjkeller: I don't think it matter much but this could also be `>= MAX_QUEUED_CONNECT_TASKS`

jasonjkeller (approved):

PR #1610, opened 2023-11-29
adding httstatus in segment

caioborgesleal:

### Overview
In the current version, we dont have http status code for http clients using spring webclient.

Captura de Tela 2023-11-29 às 14 56 14

### Related Github Issue

### Testing

### Checks

- [X] Your contributions are backwards compatible with relevant frameworks and APIs.
- [X] Your code does not contain any breaking changes. Otherwise please describe. 
- [X] Your code does not introduce any new dependencies. Otherwise please describe.

### DIFF
--- a/instrumentation/spring-webclient-5.0/src/main/java/com/nr/agent/instrumentation/spring_webclient/Util.java
+++ b/instrumentation/spring-webclient-5.0/src/main/java/com/nr/agent/instrumentation/spring_webclient/Util.java
@@ -62,7 +62,7 @@ public class Util {
                             .uri(uri)
                             .procedure("exchange")
                             .inboundHeaders(new InboundResponseWrapper(clientResponse))
-                            .status(clientResponse.statusCode().value())
+                            .status(clientResponse.statusCode().value(), "" + clientResponse.statusCode().value())
                             .build());
                     segment.end();
                 } catch (Throwable e) {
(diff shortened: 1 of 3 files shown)


pauloproni (approved):

instrumentation/spring-webclient-5.0/src/main/java/com/nr/agent/instrumentation/spring_webclient/Util.java
```diff
@@ -62,6 +62,7 @@ public void accept(ClientResponse clientResponse) {
                             .uri(uri)
                             .procedure("exchange")
                             .inboundHeaders(new InboundResponseWrapper(clientResponse))
+                            .status(clientResponse.statusCode().value(), "" + clientResponse.statusCode().value())
```
jtduffy, line 65: Instead of empty string concatenation, we would prefer `String.valueOf(clientResponse.statusCode().value())`.

jtduffy (changes requested):
Thanks for your contribution @caioborgesleal 

Once the change is made and our test suite passes, we'll merge this in for our next release.

instrumentation/spring-webclient-5.0/src/main/java/com/nr/agent/instrumentation/spring_webclient/Util.java
meiao: Actually, I think that value could be null.
I don't think it is very useful to duplicate the code as the message, and I don't think Spring webclient provides the response message.

caioborgesleal: Changed to null.

jtduffy:
Verified PR test suite passes via #1614

Issue #1300, opened 2023-06-07
Instrumentation Support for Jetty Server 12+


tbradellis:
Jetty 12 has not released yet.
It was originally slated for June 1st, but there was no GA version released.
 https://www.eclipse.org/jetty/
Jetty 12 will have support for Jakarta EE 10

Download site:
https://www.eclipse.org/jetty/download.php

We capped our instrumentation as exclusive start 12-alpha
https://github.com/newrelic/newrelic-java-agent/blob/06c4e2ef4c5c66d5573d294acbf729a943d709e2/instrumentation/jetty-11/build.gradle#L13

Jetty Version | Servlet | License | Java
-- | -- | -- | --
Jetty 12.0.x | 6.0, 5.0, 4.0, 3.1 | JakartaEE 10/jakartaee.servlet.*JakartaEE 9/jakartaee.servlet.*JakartaEE 8/javax.servlet.* | Java 17+

Priority:
Must have

seancorfield:
Is this planned to be in 8.7.0 along with JDK 21 support? We're eager to move to JDK 21 + Jetty 12 for the vthreads stuff (we're on JDK 20 + Jetty 11 right now, and we've been testing JDK 21 + Jetty 12 locally).

meiao:
Unfortunately this did not make the cut for 8.7.0.

seancorfield:
Thanks, @meiao -- do you know if Jetty 12 will make 8.8.0? (I assume JDK 21 is the focus right now for 8.7.0, which will at least move us forward one step).

meiao:
We will start working on this soon. We expect it to be in 8.8.0.

PR #1621, opened 2023-12-01
Add support for Jetty 12

obenkenobi:
### Overview
Adds instrumentation for Jetty 12. 

Jetty 12 follows a philosophy of being *Servlet API* independent. As a result Jetty now ships with a Core Server that makes no reference to the Servlet API. In addition, Jetty now supports every Jakarta EE spec from 8 to 10. As a result, an instrumentation module is needed for each EE implementation. This PR also adds instrumentation for the core Jetty Server. As a result 5 new instrumentation modules are added.  These new modules can be described as follows:

#### jetty-12 
This module instruments the core Jetty server that handles tasks that does not require the Servlet API. Since async activity is not managed in the core server, transactions cannot be started and managed in this module. The tasks this module does include:

- Reporting thread-pool metrics
- Reporting the dispatcher name and version as well as the port number for the server.
- Adding outbound CAT headers for outgoing requests. CAT however is deprecated during the time this module is made (i.e. December 2023).

#### jetty-ee*-servlet-12
This refers to any module named based on the pattern `jetty-ee*-servlet-12`.
These modules instrument the Servlet API implementations for Jetty.
As a result, the instrumentation works with tasks that require us to work with the Servlet API such as:

- Start and ending web transactions
- Linking transactions to async activity
- Reporting errors

#### List of new instrumentation modules
- jetty-12 
- jetty-ee8-servlet-12
- jetty-ee9-servlet-12
- jetty-ee10-servlet-12
- jetty-ee10-servlet-12.0.1

### Related Github Issue

https://github.com/newrelic/newrelic-java-agent/issues/1300


instrumentation/jetty-12/src/main/java/com/nr/agent/instrumentation/jetty12/JettySampler.java
```diff
@@ -0,0 +1,85 @@
...
+        final ThreadPool threadPool = server.getThreadPool();
+        if (threadPool != null) {
+            final String name = threadPool.getClass().getSimpleName();
+
+            int totalThreads, busyThreads;
+
+            int threadCount = threadPool.getThreads();
+            int idleCount = threadPool.getIdleThreads();
+
+            // Jetty9 ExecutorThreadPool doesn't expose getMaxThreads(), so just approximate usage
```
jtduffy, line 62: Jetty9 --> Jetty12?

instrumentation/jetty-12/src/main/java/org/eclipse/jetty/server/Response_Instrumentation.java
```diff
@@ -0,0 +1,24 @@
...
+@Weave(originalName = "org.eclipse.jetty.server.Response", type = MatchType.Interface)
+public class Response_Instrumentation {
+    public void write(boolean last, ByteBuffer content, Callback callback) {
+        if (!isCommitted()) {
+            NewRelic.getAgent().getTransaction().addOutboundResponseHeaders();
+        }
+        Weaver.callOriginal();
+    }
+
+    public boolean isCommitted() {
```
jtduffy, line 21: To reduce some instrumentation overhead, you can make your class abstract, and then declare `isCommitted` as abstract.
`public abstract boolean isCommitted();`

instrumentation/jetty-12/src/main/java/org/eclipse/jetty/server/Server_Instrumentation.java
```diff
@@ -0,0 +1,38 @@
...
+public abstract class Server_Instrumentation {
+
+    protected void doStart() {
+        AgentBridge.privateApi.addSampler(new JettySampler(this), 1, TimeUnit.MINUTES);
+        Weaver.callOriginal();
+    }
+
+    // Required so that earlier jetty versions do not apply.
+    // Transactions are managed in the jetty-12-ee* modules.
+    public boolean handle(Request request, Response response, Callback callback) {
```
jtduffy, line 23: Same here.. You can make the following methods `abstract`

jtduffy (changes requested):
Looks good. Aside from my inline comments:
- Most of the class files are missing the copyright header. Those that have the header have the incorrect year (2022).

instrumentation/jetty-12/src/main/java/com/nr/agent/instrumentation/jetty12/JettySampler.java
obenkenobi: Changed it to `Jetty 9+`

obenkenobi:
> Looks good. Aside from my inline comments:
> 
> * Most of the class files are missing the copyright header. Those that have the header have the incorrect year (2022).

Added the copyright headers

jtduffy (approved):

PR #1634, opened 2023-12-06
CSEC Java Agent Release Bump 1.0.7

lovesh-ap:
CSEC Java Agent Release Bump 1.0.6 to 1.0.7 Public Preview


lovesh-ap:
One additional change was added which is not required

PR #1679, opened 2024-01-04
Deprecate browser footer injection APIs;header injection now adds both scripts

### LINKED ISSUE #1657: Browser agent updates
### Description
The Browser agent has some changes to the way it is to be added to a page.

### Acceptance Criteria
Instead of using 2 script tags, a single one should be used and the contents of what were put in the script tags should be separated by a semi-colon and a line break. Eg.
```
// old
<script type="text/javascript">window.NREUM||(NREUM={});NREUM.info=JAVASCRIPT_AGENT_CONFIGURATION;</script>
<script type="text/javascript" nonce="NONCE">JAVASCRIPT_AGENT</script>

// new
<script type="text/javascript" nonce="NONCE">window.NREUM||(NREUM={});NREUM.info=JAVASCRIPT_AGENT_CONFIGURATION;
JAVASCRIPT_AGENT</script>
```
If NONCE is not provided, then the attribute `nonce` should not be set in the script tag.
This change should apply to the code for both `getBrowserTimingHeader` and `getBrowserTimingFooter`.

Deprecate`getBrowserTimingFooter` and update documentation to reflect that this `getBrowserTimingHeader` is the preferred method to be called.

jtduffy:
### Overview
Resolves #1657 

This change deprecates the browser footer API (`getBrowserTimingFooter`) and changes the behavior of the browser header API/instrumentation ('getBrowserTimingHeader`) to now include both scripts in the `<head>` element. 

Calling the browser footer API returns an empty string.

### Testing

### Checks


newrelic-agent/src/main/java/com/newrelic/api/agent/NewRelicApiImplementation.java
kford-newrelic, line 1: For the methods that are now noops, should we also update the Javadoc comments to call this out - and perhaps reference the header-only method that essentially replaces it?

jtduffy: The [javadoc](https://github.com/newrelic/newrelic-java-agent/pull/1679/files#diff-4ec5bf535fb512176b00cc19f94beef5f6d09618ec347f696d544d3ec37a0a64R110) in `PublicApi.java` contains the information regarding the deprecation/replacements of the footer related methods.

newrelic-agent/src/test/java/com/newrelic/agent/browser/JavaScriptAgentConfigTest.java
```diff
@@ -1,10 +1,3 @@
-/*
- *
- *  * Copyright 2020 New Relic Corporation. All rights reserved.
- *  * SPDX-License-Identifier: Apache-2.0
- *
- */
-
 package com.newrelic.agent.browser;
```
meiao, line 1: Copyright was lost.

meiao (approved):

Issue #1689, opened 2024-01-17
Synthetics-Info Header Causing NPE


deleonenriqueta:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

### Description
When a synthetics request is sent, the request contains an `X-NewRelic-Synthetics` header, but the agent is expecting both the `X-NewRelic-Synthetics` header and the `X-NewRelic-Synthetics-Info` header. When an `X-NewRelic-Synthetics-Info` header is not a part of the request, null values are added to the InboundHeaderState where information from the `X-NewRelic-Synthetics-Info` would have been. The agent is then trying to add those null attributes from the InboundHeaderState to the intrinsic attributes of transactions. This is causing the NullPointerException.

  
 
### Acceptance Criteria

1. Add null checks to prevent NPE's when adding attributes from the InboundHeaderState to transaction intrinsic attributes in _Transaction.java_: 
     <img width="500" alt="image" src="https://github.com/newrelic/newrelic-java-agent/assets/64384521/409f5942-eed4-462a-86bc-4082ac469245">
 
2. Add a null check to prevent unnecessary parsing of `X-NewRelic-Synthetics-Info` in _InboundHeaderState.java_:
      <img width="300" alt="image" src="https://github.com/newrelic/newrelic-java-agent/assets/64384521/1ea893b0-38ae-4212-a34b-ae4112c22afd">

PR #1690, opened 2024-01-17
Add Null Checks for Synthetics-Info Header Attributes

deleonenriqueta:
Issue #1689

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/Transaction.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/Transaction.java
@@ -1093,19 +1093,24 @@ public class Transaction {
                             this.getInboundHeaderState().getSyntheticsMonitorId());
                     getIntrinsicAttributes().put(AttributeNames.SYNTHETICS_JOB_ID,
                             this.getInboundHeaderState().getSyntheticsJobId());
-                    getIntrinsicAttributes().put(AttributeNames.SYNTHETICS_TYPE,
-                            this.getInboundHeaderState().getSyntheticsType());
-                    getIntrinsicAttributes().put(AttributeNames.SYNTHETICS_INITIATOR,
-                            this.getInboundHeaderState().getSyntheticsInitiator());
                     getIntrinsicAttributes().put(AttributeNames.SYNTHETICS_VERSION,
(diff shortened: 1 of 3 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/Transaction.java
```diff
@@ -1093,19 +1093,24 @@ private void finishTransaction() {
...
-                    getIntrinsicAttributes().put(AttributeNames.SYNTHETICS_TYPE,
-                            this.getInboundHeaderState().getSyntheticsType());
-                    getIntrinsicAttributes().put(AttributeNames.SYNTHETICS_INITIATOR,
-                            this.getInboundHeaderState().getSyntheticsInitiator());
                     getIntrinsicAttributes().put(AttributeNames.SYNTHETICS_VERSION,
-                            this.getInboundHeaderState().getSyntheticsVersion());
-
-                    Map<String, String> attrsMap = this.getInboundHeaderState().getSyntheticsAttrs();
-                    String attrName;
+                            String.valueOf(this.getInboundHeaderState().getSyntheticsVersion()));
```
jasonjkeller, line 1097: Is the synthetics version supposed to be added to the intrinsic attributes as a `String` or `Number`? And why is it being changed now?

I'm a bit unclear on why there are two different `getSyntheticsVersion()` methods, one that returns an `int` and the other a `String`.

jasonjkeller (approved):

newrelic-agent/src/main/java/com/newrelic/agent/Transaction.java
deleonenriqueta: According to agent specs, all `synthetics-info` attributes added to the intrinsic attributes should be in the form of a `String`. Re-reading the agent specs, the synthetics version is not _required_ to be added to the intrinsic attributes, so this line can be removed.

There are two `getSyntheticsVersion()` methods, because one version is pulling from the original `synthetics` header and is converting it to an `int` in order to confirm supportability. The second is to retrieve the `String` value of the version that exists within the `synthetics-info` header.

jtduffy (approved):

Issue #1715, opened 2024-01-31
New Relic java agent 8.3.0 is Broken for Apache HttpClient5 v5.1.4


patrickjamesbarry:
We are using 

- Apache Http Client 5.1.4 async api 
- org.apache.httpcomponents.core5 5.1.5

We pulled down your New Relic java agent 8.3.0 and immediately had issues. It is probably because you guys are embedding a newer version of the apache lib?  

Here is the output
2024-01-30T15:27:53,509-0500 [21064 84] com.newrelic FINER: Non-API call to setThrowable from asynchronous activity ignored: java.lang.ClassCastException: class org.apache.hc.core5.http.Message cannot be cast to class org.apache.hc.core5.http.message.BasicHttpResponse (org.apache.hc.core5.http.Message and org.apache.hc.core5.http.message.BasicHttpResponse are in unnamed module of loader 'app') with priority TRACER

Debugging, here is full stack trace
java.lang.ClassCastException: class org.apache.hc.core5.http.Message cannot be cast to class org.apache.hc.core5.http.message.BasicHttpResponse (org.apache.hc.core5.http.Message and org.apache.hc.core5.http.message.BasicHttpResponse are in unnamed module of loader 'app')
	at com.nr.agent.instrumentation.httpclient50.WrappedFutureCallback.completed(WrappedFutureCallback.java:36)
	at org.apache.hc.core5.concurrent.BasicFuture.completed(BasicFuture.java:123)
	at org.apache.hc.core5.concurrent.ComplexFuture.completed(ComplexFuture.java:79)
	at org.apache.hc.client5.http.impl.async.InternalAbstractHttpAsyncClient$1$3$1.completed(InternalAbstractHttpAsyncClient.java:299)
	at org.apache.hc.core5.http.nio.support.BasicResponseConsumer$2.completed(BasicResponseConsumer.java:95)
	at org.apache.hc.core5.http.nio.entity.AbstractBinAsyncEntityConsumer.completed(AbstractBinAsyncEntityConsumer.java:84)
	at org.apache.hc.core5.http.nio.entity.AbstractBinDataConsumer.streamEnd(AbstractBinDataConsumer.java:81)
	at org.apache.hc.core5.http.nio.support.BasicResponseConsumer.streamEnd(BasicResponseConsumer.java:127)
	at org.apache.hc.client5.http.impl.async.HttpAsyncMainClientExec$1.streamEnd(HttpAsyncMainClientExec.java:235)
[...]

Issue #1649, opened 2023-12-07
Investigate null response on httpclient5


meiao:
### Description
In some cases, after an async request is made using httpclient5, the call is completed and a null Object is passed as the response.

This causes an NPE when the instrumentation is trying to get the http response code. Further, this prevents a segment from being finished, causing the transaction to appear to have taken 10 minutes (due to segment timeout).

### Acceptance Criteria
Code is guarded to prevent NPE when the response is being read at https://github.com/newrelic/newrelic-java-agent/blob/9f144fd80f85e69fa735c2713cd1bc138337e659/instrumentation/httpclient-5.0/src/main/java/com/nr/agent/instrumentation/httpclient50/InstrumentationUtils.java#L80

Also, end the segment in a finally block.

kford-newrelic:
Additional reference - [NR-214951](https://new-relic.atlassian.net/browse/NR-214951)

jbedell-newrelic:
Resolved by: https://github.com/newrelic/newrelic-java-agent/pull/1719

PR #1719, opened 2024-02-02
1715 http client5 fixes

jbedell-newrelic:
instanceof checks when handling the response
and another null check

### DIFF
--- a/instrumentation/httpclient-5.0/src/main/java/com/nr/agent/instrumentation/httpclient50/InstrumentationUtils.java
+++ b/instrumentation/httpclient-5.0/src/main/java/com/nr/agent/instrumentation/httpclient50/InstrumentationUtils.java
@@ -58,6 +58,7 @@ public class InstrumentationUtils {
     }
 
     public static void processResponse(URI requestURI, HttpResponse response, Segment segment) {
+        if (response == null) return;
         HttpParameters params = createInboundParams(requestURI, response);
         if (segment != null) {
             segment.reportAsExternal(params);
--- a/instrumentation/httpclient-5.0/src/main/java/com/nr/agent/instrumentation/httpclient50/WrappedFutureCallback.java
+++ b/instrumentation/httpclient-5.0/src/main/java/com/nr/agent/instrumentation/httpclient50/WrappedFutureCallback.java
@@ -11,6 +11,8 @@ import com.newrelic.agent.bridge.AgentBridge;
 import com.newrelic.api.agent.Segment;
 import org.apache.hc.core5.concurrent.FutureCallback;
 import org.apache.hc.core5.http.HttpRequest;
+import org.apache.hc.core5.http.HttpResponse;
+import org.apache.hc.core5.http.Message;
 import org.apache.hc.core5.http.message.BasicHttpResponse;
 
 import java.net.URISyntaxException;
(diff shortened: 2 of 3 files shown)


instrumentation/httpclient-5.0/src/main/java/com/nr/agent/instrumentation/httpclient50/InstrumentationUtils.java
```diff
@@ -58,6 +58,7 @@ public static void handleUnknownHost(Exception e) {
     }
 
     public static void processResponse(URI requestURI, HttpResponse response, Segment segment) {
+        if (response == null) return;
```
meiao, line 61: Do you know what triggers the response to be null?
Does it still make sense to create the external in that case?
The HttpParameters could be created calling `noInboudHeaders` instead of `inboundHeaders(Response)`.

instrumentation/httpclient-5.0/src/main/java/com/nr/agent/instrumentation/httpclient50/WrappedFutureCallback.java
```diff
@@ -33,7 +35,14 @@ public WrappedFutureCallback (HttpRequest request, FutureCallback origCallback,
...
     public void completed(T response) {
         try {
-            InstrumentationUtils.processResponse(request.getUri(), (BasicHttpResponse)response, segment);
+            if (response instanceof HttpResponse) {
+                InstrumentationUtils.processResponse(request.getUri(), (HttpResponse) response, segment);
+            } else if (response instanceof Message && ((Message)response).getHead() instanceof HttpResponse) {
+                HttpResponse resp2 = (HttpResponse)(((Message)response).getHead());
+                InstrumentationUtils.processResponse(request.getUri(), resp2, segment);
+            } else {
+                AgentBridge.getAgent().getLogger().log(Level.FINER, "Unhandled response type: "+response.getClass());
```
meiao, line 44: This creates a StringBuilder even if the message is not going to be logged.

Could you use:
```
 AgentBridge...log(Level.FINER, "Unhandled response type: {0}", response.getClass());
```
or guard with a check to whether FINER level is enabled.

And regardless, if response is null here, it will throw a NPE.

instrumentation/httpclient-5.0/src/test/java/com/nr/agent/instrumentation/httpclient/HttpClient5Test.java
```diff
@@ -55,19 +63,28 @@ public class HttpClient5Test {
...
     @Test
     public void testErrorAsync() throws Exception {
-        testError(true);
+        testError(true, false);
     }
 
-    private void testError(boolean async) throws Exception {
+    @Test
+    public void testErrorAsyncWithMessage() throws Exception {
+        testError(true, false);
```
meiao, line 76: shouldn't this be true, true?

instrumentation/httpclient-5.0/src/main/java/com/nr/agent/instrumentation/httpclient50/InstrumentationUtils.java
jbedell-newrelic: Good question, I'm not sure what causes that.  What are your thoughts on creating an external anyway?

instrumentation/httpclient-5.0/src/test/java/com/nr/agent/instrumentation/httpclient/HttpClient5Test.java
jbedell-newrelic: Fixed

meiao (approved):

jbedell-newrelic:
Resolves: https://github.com/newrelic/newrelic-java-agent/issues/1649
Resolves: https://github.com/newrelic/newrelic-java-agent/issues/1715

PR #1729, opened 2024-02-08
Decrease log instrumentation overhead when log forwarding is disabled account wide #1572

### LINKED ISSUE #1572: Decrease log instrumentation overhead when log forwarding is disabled account wide
### Description
When log forwarding is disabled account wide the instrumentation module is not aware of it and will impose some overhead on the system.

To notify that log forwarding is disabled account wide, the collector sends the agent information telling that the `max_samples_stored` for logging is 0.

But the "collector" data is read and  a JSONObject and put in the config as `event_harvest_config`. This info is not sent to the AgentConfig's respective configuration `application_logging.forwarding.max_samples_stored`.

So the instrumentation modules use the wrong value from `application_logging.forwarding.max_samples_stored` and never know when forwarding is disabled account wide.

meiao:
### Overview
To solve the problems on #1572 this PR adds a server property to the base config. The property `application_logging.forwarding.enabled` will be set to false when the harvest limit for log events is 0.

It was added as a property rather than calculated in `ApplicationLoggingForwardingConfig` because the instrumentation modules do not have access to the config objects and have to query for the properties using the Config interface.

### Related Github Issue
Fixes #1572 

### Checks

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/config/AgentConfigFactory.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/config/AgentConfigFactory.java
@@ -9,10 +9,13 @@ package com.newrelic.agent.config;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
+import com.newrelic.agent.Agent;
+import com.newrelic.agent.HarvestServiceImpl;
 import com.newrelic.agent.browser.BrowserConfig;
 import com.newrelic.agent.config.internal.DeepMapClone;
 import com.newrelic.agent.database.SqlObfuscator;
 import com.newrelic.agent.reinstrument.RemoteInstrumentationServiceImpl;
+import com.newrelic.agent.transport.CollectorMethods;
 import com.newrelic.agent.transport.ConnectionResponse;
 
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/config/AgentConfigFactory.java
```diff
@@ -229,7 +231,22 @@ public static void mergeServerData(Map<String, Object> settings, Map<String, Obj
...
+        // when log forwarding is disabled account wide, the backend will inform the agent by setting the harvest limit to 0
+        if (eventHarvestConfig instanceof Map) {
+            Object harvestLimits = ((Map<?,?>) eventHarvestConfig).get(HarvestServiceImpl.HARVEST_LIMITS);
+            if (harvestLimits instanceof Map) {
+                Object logLimit = ((Map<?,?>) harvestLimits).get(CollectorMethods.LOG_EVENT_DATA);
+                if (logLimit instanceof Number && ((Number) logLimit).intValue() == 0) {
+                    String loggingForwardingEnabled = AgentConfigImpl.APPLICATION_LOGGING + "." + ApplicationLoggingConfigImpl.FORWARDING + "." +
+                            ApplicationLoggingForwardingConfig.ENABLED;
+                    addServerProp(loggingForwardingEnabled, false, settings);
+                }
```
meiao, line 246: I had logging when this happens, but there is another log statement that kinda notifies us that log forwarding is turned off.

jtduffy (approved):

PR #1744, opened 2024-02-16
CSEC Version bump to 1.1.1

lovesh-ap:
CSEC Version bump to [1.1.1](https://github.com/newrelic/csec-java-agent/releases/tag/1.1.1) 
Remove config to enable/disable low-priority-instrumentation, from 1.1.1 onwards these modules are enabled by default.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,6 @@
 # The agent version.
 agentVersion=8.10.0
-securityAgentVersion=1.1.0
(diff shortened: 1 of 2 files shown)


jtduffy:
@lovesh-ap When will v1.1.1 be available in maven central?

lovesh-ap:
It was released out from sonatype repo on same day. I'm not sure why it's not available on maven yet.
We are investigating

PR #1762, opened 2024-02-26
NPE guard in AgentAttributeSender.removeAttribute()

### LINKED ISSUE #1749: NullPointerException if setUserId is called with an empty argument and no transaction is active
## Description
We occasionally encounter an exception from calls to `NewRelic.setUserId()`. This appears to be caused by [this line](https://github.com/newrelic/newrelic-java-agent/blob/d49c2ca5b667f7dc0c3b68111ae1da96993e8a58/newrelic-agent/src/main/java/com/newrelic/agent/attributes/AgentAttributeSender.java#L32) in `AgentAttributeSender`, which returns a null map reference when no transaction is active.

```
java.lang.NullPointerException: Cannot invoke "java.util.Map.remove(Object)" because the return value of "com.newrelic.agent.attributes.AgentAttributeSender.getAttributeMap()" is null
    at com.newrelic.agent.attributes.AgentAttributeSender.removeAttribute(AgentAttributeSender.java:37)
    at com.newrelic.api.agent.NewRelicApiImplementation.setUserId(NewRelicApiImplementation.java:270)
    at com.newrelic.api.agent.NewRelic.setUserId(NewRelic.java:229)
```

## Expected Behavior
Calls to `setUserId()` should succeed regardless of value of user ID argument or transaction state.

## Your Environment
* NewRelic Java agent v8.9.1
* JDK Eclipse Temurin 21.0.2

jtduffy:
Resolves #1749 

Add an NPE check prior to calling `.remove()` on the return value from `getAttributeMap()`.

This corrects the issue where `NewRelicApiImplementation.setUserId()` was throwing an exception when called outside of a transaction.

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/attributes/AgentAttributeSender.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/attributes/AgentAttributeSender.java
@@ -34,6 +34,10 @@ public class AgentAttributeSender extends AttributeSender {
     }
 
     public void removeAttribute(String key) {
+        Map<String, Object> attributeMap = getAttributeMap();
+        if (attributeMap == null) {
+            return;
+        }
         getAttributeMap().remove(key);
     }
 
--- a/newrelic-agent/src/main/java/com/newrelic/agent/attributes/AgentAttributeSender.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/attributes/AgentAttributeSender.java
@@ -38,7 +38,7 @@ public class AgentAttributeSender extends AttributeSender {
         if (attributeMap == null) {
             return;
         }
-        getAttributeMap().remove(key);
+        attributeMap.remove(key);
     }
 
 }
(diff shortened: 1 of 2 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/attributes/AgentAttributeSender.java
```diff
@@ -34,6 +34,10 @@ protected Map<String, Object> getAttributeMap() {
     }
 
     public void removeAttribute(String key) {
+        Map<String, Object> attributeMap = getAttributeMap();
```
meiao, line 37: you can use the attributeMap from this line in line 40. Will save 1 method call until JIT optimizes.

meiao (approved):

newrelic-agent/src/main/java/com/newrelic/agent/attributes/AgentAttributeSender.java
jtduffy: Yup, totally forgot to change that.

meiao (approved):

Issue #1648, opened 2023-12-07
Review how the Java Agent handles HTTP responses


kford-newrelic:
### Description
The Java Agent _could be_ unnecessarily holding onto HTTP responses, which translates into a larger amount of memory use. Need to investigate this a bit more to determine whether we can avoid this or whether it's a necessary cost.

If we determine that holding onto the response is necessary, we should further investigate a possible optimization by preventing child traces from being generated? Should that be something that we allow to be configurable by customers? What's the impact to the trace in doing this?

### Acceptance Criteria
Need to determine whether the agent _is_ in fact holding onto HTTP responses too long. If yes, then we need to determine whether an optimized solution can be implemented.

### Design Consideration/Limitations
N/A

### Dependencies
N/A

### Additional context
Original Jira issue - [NR-82809](https://new-relic.atlassian.net/browse/NR-82809)

kford-newrelic:
Should be configuration driven

jbedell-newrelic:
Resolved by https://github.com/newrelic/newrelic-java-agent/pull/1775

PR #1775, opened 2024-02-29
1648 http response fixes

jbedell-newrelic:
The PR addresses the latest version of each module that handles external HTTP calls.  The various wrapper classes used to access the headers of the response have been changed to only hold onto the headers rather than the entire response object.


jbedell-newrelic:
Resolves https://github.com/newrelic/newrelic-java-agent/issues/1648

jtduffy (approved):

vikinghawk:
is there a reason this change wasn't made to spring_webclient5 as well?

jbedell-newrelic:
Hi @vikinghawk , thanks for the question.  At the time, in order to limit scope and potential impact, we made the decision to only update the latest version for most modules.  As a result of your question, we discussed this today and decided to open a new ticket to address it for spring-webclient-5.0 as well.  [Here](https://github.com/newrelic/newrelic-java-agent/issues/2029) is a link to the ticket, so you can follow the progress.

Thanks for getting in touch!

PR #1783, opened 2024-03-05
Prevent duplicate DT headers for grpc

### LINKED ISSUE #1768: Prevent duplicate DT headers for grpc
## Description 

We got a report from [NR-231941](https://new-relic.atlassian.net/browse/NR-231941) that distributed tracing was broken between two apps using grpc. Specifically, traceIds from the upstream app were not being used by the downstream app, and the two apps were not being linked in the UI. 

After investigation we discovered that the distributed tracing headers `newrelic`, `traceparent`, and `tracestate` were being duplicated (ie same key, same value) in the upstream app's outbound headers. Because the headers were duplicated, the DT payload was not accepted by the downstream app.

## To do:

1. Prevent duplicate distributed tracing headers from being added to the `metadata` object in the `OutboundHeadersWrapper` class of the [grpc instrumentation](https://github.com/newrelic/newrelic-java-agent/blob/c495abc03caa9b51b9e0932e261147a55c4c60e5/instrumentation/grpc-1.40.0/src/main/java/com/nr/agent/instrumentation/grpc/OutboundHeadersWrapper.java#L30). This needs to be done for all versions of grpc. 
2. In `W3CTraceParentParser.parseHeaders` we check that the `traceparent` header has size 1. Within this check, add additional logic that will validate the headers if their size is not 1 but their values agree. The supportability metric  in that check should still be generated. 
3. Add a WARNING level log message within`W3CTraceParentParser.parseHeaders` when multiple trace parent headers are found. 
## Considerations

Are there any valid cases for duplicate DT headers? Is overwriting the DT headers acceptable, or should we just prevent adding new duplicate headers?

## Acceptance Criteria

- Outbound headers are not duplicated
- The issue is fixed for all versions of the grpc instrumentation
- New tests for duplicate headers are added
- All tests passing

kanderson250:
Resolves #1768 

This PR addresses a bug with grpc apps using distributed tracing. Previously our grpc instrumentation allowed duplicate DT headers, which caused DT payloads to be rejected by downstream services. This is now prohibited. 

Changes:
- Prevent adding duplicate headers in all grpc instrumentation modules
- Allow parsing of duplicate (same value) parent headers in `W3CTraceParentParser`. Also add a log message when this happens to help diagnose the issue should it recur.

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/tracing/W3CTraceParentParser.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/tracing/W3CTraceParentParser.java
@@ -24,16 +24,11 @@ public class W3CTraceParentParser {
             ServiceFactory.getStatsService().getMetricAggregator().incrementCounter(MetricNames.SUPPORTABILITY_TRACE_CONTEXT_INVALID_PARENT_HEADER_COUNT);
             Agent.LOG.log(Level.WARNING, "Multiple traceparent headers found on inbound request.");
             // Multiple values ok if all are equal
-            boolean allHeadersEqual = true;
             String first = traceParentHeaders.get(0);
             for (String header : traceParentHeaders) {
                 if (!header.equals(first)) {
-                    allHeadersEqual = false;
-                    break;
+                   return null;
                 }
-            }
-            if (!allHeadersEqual) {
-                return null;
             }
         }
         String traceParentHeader = traceParentHeaders.get(0);
--- a/newrelic-agent/src/main/java/com/newrelic/agent/tracing/W3CTraceParentParser.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/tracing/W3CTraceParentParser.java
@@ -7,19 +7,25 @@
 
 package com.newrelic.agent.tracing;
 
+import com.newrelic.agent.Agent;
 import com.newrelic.agent.MetricNames;
 import com.newrelic.agent.service.ServiceFactory;
 
 import java.util.List;
+import java.util.logging.Level;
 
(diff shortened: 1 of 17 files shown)


jtduffy (approved):

newrelic-agent/src/main/java/com/newrelic/agent/tracing/W3CTraceParentParser.java
```diff
@@ -7,19 +7,25 @@
...
 
 public class W3CTraceParentParser {
 
     static W3CTraceParent parseHeaders(List<String> traceParentHeaders) {
         if (traceParentHeaders.size() != 1) {
             ServiceFactory.getStatsService().getMetricAggregator().incrementCounter(MetricNames.SUPPORTABILITY_TRACE_CONTEXT_INVALID_PARENT_HEADER_COUNT);
-            return null;
+            Agent.LOG.log(Level.WARNING, "Multiple traceparent headers found on inbound request.");
+            // Multiple values ok if all are equal
+            boolean allHeadersEqual = traceParentHeaders.stream().allMatch(h -> h.equals(traceParentHeaders.get(0)));
```
jasonjkeller, line 24: You might want to consider refactoring this to avoid using streams. The stream API can cause significantly more overhead when being used on small data structures than more traditional looping approaches. Ideally you should avoid using streams unless you're doing complex processing on a large data set.

kanderson250: Noted, I think the modified loop is readable enough that performance should be the foremost concern. Updated to remove the streams API, plus a small refactoring and test for the empty list edge case.

PR #1798, opened 2024-03-12
Auto-configure the opentelemetry SDK

sdaubin:
### Overview
This PR allows users to use the OpenTelemetry SDK and auto-configuration to send dimensional metrics to New Relic.

In the service code, include the OpenTelemetry SDK:

```
    implementation 'io.opentelemetry:opentelemetry-sdk-extension-autoconfigure:1.33.0'
    implementation 'io.opentelemetry:opentelemetry-sdk:1.33.0'
    implementation 'io.opentelemetry:opentelemetry-exporter-otlp:1.33.0'
```

Access `OpenTelemetry` using `GlobalOpenTelemetry`:

    GlobalOpenTelemetry.get().getMeterProvider().get("test").histogramBuilder("test.metric").build();

When running the instrumented service, set `otel.java.global-autoconfigure.enabled`:

    -Dotel.java.global-autoconfigure.enabled=true

The java agent will configure the proper endpoint for sending data to the New Relic service.

Test app: https://github.com/sdaubin/kafka-examples

## Value add

This makes it easy for customers to report dimensional metrics into their APM service with no special configuration.  This is difficult today, especially trying to get data reported with the correct `entity.guid` stamp.

This also makes it easy to use the OTel metrics SDK today and later switch to an OTel agent without changing any code.

### Related Github Issue

### Testing

### Checks


instrumentation/opentelemetry-sdk-extension-autoconfigure-1.28.0/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdk.java
```diff
@@ -0,0 +1,49 @@
...
+            final String host = agent.getConfig().getValue("host");
+            final String endpoint = "https://" + host + ":443";
+            final String licenseKey = agent.getConfig().getValue("license_key");
+            final Map<String, String> properties = new HashMap<>();
+            properties.put("otel.exporter.otlp.headers", "api-key=" + licenseKey);
+            properties.put("otel.exporter.otlp.endpoint", endpoint);
+            properties.put("otel.exporter.otlp.protocol", "http/protobuf");
+            properties.put("otel.exporter.otlp.headers", "api-key=" + licenseKey);
+            properties.put("otel.span.attribute.value.length.limit", "4095");
+            properties.put("otel.exporter.otlp.compression", "gzip");
```
jack-berg, line 33:
```suggestion
            properties.put("otel.exporter.otlp.compression", "gzip");
            properties.put("otel.exporter.otlp.metrics.temporality.preference", "DELTA");
            properties.put("otel.exporter.otlp.metrics.default.histogram.aggregation", "BASE2_EXPONENTIAL_BUCKET_HISTOGRAM");
            properties.put("otel.experimental.exporter.otlp.retry.enabled", "true");
            properties.put("otel.experimental.resource.disabled.keys", "process.command_line");
```

instrumentation/opentelemetry-sdk-extension-autoconfigure-1.28.0/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdk.java
```diff
@@ -0,0 +1,49 @@
...
+
+            final Object appName = agent.getConfig().getValue("app_name");
+            properties.put("otel.service.name", appName.toString());
+
+            String resourceAttributes = System.getProperty("otel.resource.attributes");
+            if (resourceAttributes == null) {
+                resourceAttributes = System.getenv("OTEL_RESOURCE_ATTRIBUTES");
+            }
+            resourceAttributes = (resourceAttributes == null ? "" : resourceAttributes + ",")
+                    + "entity.guid=" + entityGuid;
```
jack-berg, line 43: Would be good to set `service.instance.id` if its not set.

instrumentation/opentelemetry-sdk-extension-autoconfigure-1.28.0/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdk.java
```diff
@@ -0,0 +1,49 @@
...
+            properties.put("otel.resource.attributes", resourceAttributes);
+            builder.setConfig(DefaultConfigProperties.create(properties));
```
jack-berg, line 45: If you can access the `DefaultConfigProperties` class, should consider calling `builder.addPropertiesCustomizer` instead. It will give you the the ability to say things like:
```
if (configProperties.getString("otel.exporter.otlp.endpoint") == null) {..}
```
Which checks env vars and system properties at the same time. Just a bit more of a reliable way to inspect / change config.

instrumentation/opentelemetry-sdk-extension-autoconfigure-1.28.0/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdk.java
jack-berg, line 43: Adding `entity.guid` here instead of by instrumenting `Resource` directly means that users that want to configure the SDK themselves (i.e. to send data to a collector) won't benefit from all the data being associated with the same entity.

instrumentation/opentelemetry-sdk-extension-autoconfigure-1.28.0/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdk.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+@Weave(type = MatchType.ExactClass)
+public class AutoConfiguredOpenTelemetrySdk {
+
+    public static AutoConfiguredOpenTelemetrySdkBuilder builder() {
+        final AutoConfiguredOpenTelemetrySdkBuilder builder = Weaver.callOriginal();
+        if (System.getProperty("otel.exporter.otlp.endpoint") == null &&
```
jack-berg, line 19: This is a decent way to determine if they are trying to configure the SDK themselves, but not perfect. There are signal-specific overrides for the endpoint that they may be setting of the form `otel.exporter.otlp.metrics.endpoint`. If the user sets this and doesn't set `otel.exporter.otlp.endpoint`, then this if-block will be executed, but metric traffic will go to the user-configured endpoint. Not the end of the world, but potentially confusing.

I wonder if we can think of a more reliable signal to trigger this configuration? 🤔

instrumentation/opentelemetry-sdk-extension-autoconfigure-1.28.0/build.gradle
```diff
@@ -0,0 +1,29 @@
+plugins {
+    id("java")
+}
+
+dependencies {
+    implementation(project(":agent-bridge"))
+    implementation(project(":newrelic-api"))
+    implementation(project(":newrelic-weaver-api"))
+    implementation("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure:1.28.0")
```
jack-berg, line 9: What's the thinking behind this specific version? Are we trying to use the earliest version that has the method we need to instrument?

sdaubin: I just picked a version.  I could go back further, but I also think it's reasonable to say that this works with newer versions of the SDK.

newrelic-agent/src/main/java/com/newrelic/agent/AgentImpl.java
```diff
@@ -23,6 +24,8 @@
 import com.newrelic.api.agent.TraceMetadata;
 
 import java.util.Map;
+import java.util.concurrent.CountDownLatch;
```
jtduffy, line 27: nit: unused import

jtduffy (approved):

PR #1808, opened 2024-03-21
Excluding Glassfish milestone

meiao:
### Overview
New Glassfish milestone is incompatible with the agent. Historically, they released some milestones that were not compatible, but later released others that were. So just leaving as a single exclude for the latest milestone.


meiao:
Duplicate

Issue #1571, opened 2023-10-31
Solr 8+ JMX metrics enhancement


kford-newrelic:
## Is your feature request related to a problem? Please describe.
The agent should be enhanced to be able to find the sharded JMX beans. Rather than hard-coding a handful of fixed names, the agent should adapt to solr 8 and be able to list or otherwise enumerate the bean names and match the arbitrary-depth domain names, as shown above.

These beans should be queried/monitored by the agent and reported to New Relic for display in the NR1 UI.

When Solr8 is using cloud sharding, some of the JMX metrics are not reporting. It would be great if they would!

## Feature Description
Feature Description

The agent should be enhanced to be able to find the sharded JMX beans. Rather than hard-coding a handful of fixed names, the agent should adapt to solr 8 and be able to list or otherwise enumerate the bean names and match the arbitrary-depth domain names, as shown above.

These beans should be queried/monitored by the agent and reported to New Relic for display in the NR1 UI.

## Describe Alternatives
Solr uses [dropwizard/codahale metrics](https://lucene.apache.org/solr/guide/8_6/metrics-reporting.html) internally, and so the [New Relic dropwizard reporter](https://github.com/newrelic/dropwizard-metrics-newrelic) might be able to be used to get the same telemetry. Some experimentation/exploration would be required to verify that the same information can be obtained...and also how it might map to a cohesive user experience.

## Additional context
If we look in the [Solr7JmxValues.java class](https://github.com/newrelic/newrelic-java-agent/blob/main/newrelic-agent/src/main/java/com/newrelic/agent/jmx/values/Solr7JmxValues.java) we can see that the `updateHandler' is looking for beanName:

`solr:dom1=core,dom2=,category=UPDATE,scope=updateHandler,name=`

but when Solr is configured for sharding, the names are [dynamically generated](https://github.com/apache/lucene-solr/blob/bd8905150d6aabd3485874f52572279d9220a8c1/solr/core/src/java/org/apache/solr/metrics/reporters/jmx/JmxObjectNameFactory.java#L90) to an arbitrary depth, and may look more like this:

`solr:dom1=core,dom2=solr-inventory,dom3=shard1,dom4=replica_n1,category=UPDATE,scope=updateHandler,name=adds`

For caching, the NR Solr 7 jmx support is looking for

`solr:dom1=core,dom2=*,category=CACHE,scope=searcher,name=documentCache`

but Solr8 might look more like

[...]

PR #1811, opened 2024-03-21
Pekko support

kanderson250:
Resolves #1692 

This PR adds support for Apache Pekko. 

Pekko is a fork of Akka 2.6. As such, the instrumentation is a direct lift of our existing Akka instrumentation. 

Corresponding AITs have also been added.

PR #1812, opened 2024-03-21
Enhance Solr 7+ JMX metrics by including clustered Solr collections which include the collection name, shard name, and (replica) core

obenkenobi:
### Overview
Allow the agent to capture clustered Solr cores via our JMX API . 

Single instances of Solr will contain function the same. 

In clustered Solr cores, their JMX MBeans will look like `solr:dom1=core,dom2=dummy_collection,dom3=shard2,dom4=replica_n2,category=CACHE,scope=searcher,name=documentCache` which the agent did not capture. 

Now the agent will export Solr values with nested DOM (i.e. clustered cores) in the formats:
- JMX/solr/{collection name}.{shard}.{core name}/{name_value}/%/{attribute}/
- JMX/solr/{collection name}.{shard}.{core name}/updateHandler/%/{attribute}

Fpr example: 
`JMX/solr/dummy_collection.shard2.replica_n2/documentCache/%/`

The Solr UI will recognize the cores in the format `{collection name}.{shard}.{core name}`

#### New Iteration Syntax Internal To The Agent

This is done by adding new iteration syntax for object name keys for MBeans when formatting them into agent metrics.

You may see it in any class that extends `com.newrelic.agent.jmx.metrics.JmxFrameworkValues` in the `newrelic-agent` module.

To see this in practice, we will query an MBean from JMX with Solr using the string:
```
solr:dom1=core,*,category=CACHE,scope=searcher,name=documentCache
```
which may return an MBean with an object name like:
```
solr:dom1=core,dom2=dummy_collection,dom3=shard2,dom4=replica_n2,category=CACHE,scope=searcher,name=documentCache
```

We take a metric name with the format `JMX/solr/{for:dom[2::.]}/documentCache/%/` such that it becomes the metric `JMX/solr/dummy_collection.shard2.replica_n2/documentCache/%/`

We can use this metric syntax in other ways such that `JMX/solr/{for:dom[2:4:.]}/documentCache/%/` can translate to `JMX/solr/dummy_collection.shard2.documentCache/%/`

In the format we see the following placeholder `{for:dom[start:end:delimiter]}` where `for:` indicate some iterable object name keys and `[start:end]` represents some range where `start` is the start number and `end` represents an optional exclusive end of the range. `[start::delimiter]` means the `end` is unbounded. This is similar to python syntax you may have with accessing sublists. The `delimiter` represents a string that will join the iterated values in a sequence. If left empty, it will default to `/`. 

Currently the solution will work with positive numbers for simplicity but we can eventually extend this to more characters.

Since the syntax is new and there maybe unexpected side effects, we will include new undocumented temporarily configuration to disable it for Solr:

yaml:

```yaml
# newrelic.yml snippet
common: &default_settings
  jmx:
    enable_iterated_objectname_Keys: true # default is true
```
system property:

```
-Dnewrelic.config.jmx.enable_iterated_objectname_Keys=true
```

environment variable:
```
NEW_RELIC_JMX_ENABLE_ITERATED_OBJECTNAME-KEYS=true
```
By default the configuration is true. To disable the syntax, set it to false.

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/1571

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/config/JmxConfig.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/config/JmxConfig.java
@@ -30,4 +30,6 @@ public interface JmxConfig {
      * MBean server.
      */
     boolean registerLinkingMetadataMBean();
+
+    boolean enableIteratedObjectNameKeys();
 }
--- a/newrelic-agent/src/main/java/com/newrelic/agent/config/JmxConfigImpl.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/config/JmxConfigImpl.java
@@ -16,7 +16,9 @@ class JmxConfigImpl extends BaseConfig implements JmxConfig {
     public static final String ENABLED = "enabled";
     public static final String REGISTER_LINKING_METADATA_MBEAN = "linkingMetadataMBean";
     public static final String DISABLED_JMX_FRAMEWORKS = "disabled_jmx_frameworks";
+    public static final String ENABLE_ITERATED_OBJECTNAME_KEYS = "enabledIteratedObjectNameKeys";
     public static final boolean DEFAULT_REGISTER_LINKING_METADATA_MBEAN = false;
+    public static final boolean DEFAULT_ENABLE_ITERATED_OBJECTNAME_KEYS = true;
     public static final Boolean DEFAULT_ENABLED = Boolean.TRUE;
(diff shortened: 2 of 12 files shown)


jtduffy (changes requested):

jtduffy (approved):

obenkenobi:
The pekko AIT failing is because this branch does not have the pekko instrumentation which is in another [PR](https://github.com/newrelic/newrelic-java-agent/pull/1811). So when all other tests pass, this branch will be merged.

PR #1813, opened 2024-03-22
Add transaction GUID to ErrorEvent instances

jtduffy:
Errors inbox team validated that both the `ErrorEvent` and `TracedError` payloads contain the transaction GUID in the proper spots when DT is disabled.


obenkenobi (approved):

jtduffy:
Pekko failure due to changes not being merged to agent main yet.

Issue #1789, opened 2024-03-06
Allow tracking of HTTP2 requests on Netty


meiao:
## Is your feature request related to a problem? Please describe.
When Netty receives an HTTP2 request, no transaction is created. If the same request is made using HTTP 1(.1) everything works fine.

kford-newrelic:
Related [GTSE](https://new-relic.atlassian.net/browse/NR-239167)

Issue #1799, opened 2024-03-13
Investigate support for HTTP2 


jbedell-newrelic:
It was discovered that our Netty instrumentation does not work with HTTP2:
Linked issue: https://github.com/newrelic/newrelic-java-agent/issues/1789

As a result, we want to check other instrumentations as well to see if there are any other deficiencies.  We may also want to explore adding tests specifically for HTTP2.

jasonjkeller:
Great overview of the changes in HTTP/2: https://web.dev/articles/performance-http2

The core change is that each `Message` (i.e. `Request`/`Response`) is being broken down into a collection of `Frames` which are then transferred over a single stream, which may interleave many `Frame`s representing many different `Message`s. This could certainly impact some agent instrumentation.

>At the core of all performance enhancements of HTTP/2 is the new binary framing layer, which dictates how the HTTP messages are encapsulated and transferred between the client and server.
>
>The "layer" refers to a design choice to introduce a new optimized encoding mechanism between the socket interface and the higher HTTP API exposed to our applications: the HTTP semantics, such as verbs, methods, and headers, are unaffected, but the way they are encoded while in transit is different. 
>
>In short, HTTP/2 breaks down the HTTP protocol communication into an exchange of binary-encoded frames, which are then mapped to messages that belong to a particular stream, all of which are multiplexed within a single TCP connection. 
>
>* Stream: A bidirectional flow of bytes within an established connection, which may carry one or more messages.
>* Message: A complete sequence of frames that map to a logical request or response message.
>* Frame: The smallest unit of communication in HTTP/2, each containing a frame header, which at a minimum identifies the stream to which the frame belongs.
>
>The relation of these terms can be summarized as follows:
>* All communication is performed over a single TCP connection that can carry any number of bidirectional streams.
>* Each stream has a unique identifier and optional priority information that is used to carry bidirectional messages.
>* Each message is a logical HTTP message, such as a request, or response, which consists of one or more frames.
>* The frame is the smallest unit of communication that carries a specific type of data—e.g., HTTP headers, message payload, and so on. Frames from different streams may be interleaved and then reassembled via the embedded stream identifier in the header of each frame.

jasonjkeller:
HTTP/3 overview: https://blog.cloudflare.com/http3-the-past-present-and-future/

PR #1815, opened 2024-03-25
Add HTTP/2 Support For Netty 4.1.16.Final +

jasonjkeller:
Resolves https://github.com/newrelic/newrelic-java-agent/issues/1789

Adds a new `netty-4.1.16` instrumentation module and caps the range of versions supported by the old `netty-4.0.8` module.

In addition to supporting HTTP/1, this new instrumentation module adds support for HTTP/2 starting with Netty version `io.netty:netty-all:4.1.16.Final` up to the latest `4.1.x `versions (at the time this instrumentation was added the most recent version of Netty was`io.netty:netty-all:4.1.107.Final`).

Note: There are a small number of old Netty versions ([4.1.0.Beta4](https://mvnrepository.com/artifact/io.netty/netty-all/4.1.0.Beta4) - [4.1.15.Final](https://mvnrepository.com/artifact/io.netty/netty-all/4.1.15.Final)) that technically provide HTTP/2 functionality but will be instrumented by the older `netty-4.0.8` module and thus won't have HTTP/2 support from us. This decision was made because those old Netty versions, which were all published from 2015-2017, have an extremely unstable HTTP/2 API which was frequently being refactored, making it difficult to instrument. Chances are extremely low that anyone would actually be using HTTP/2 with any of these early Netty versions.

AIT changes in the following PR will resolve the test failures: https://github.com/newrelic/java-agent-integration-tests/pull/255


jtduffy (approved):

jasonjkeller:
Related to https://github.com/newrelic/newrelic-java-agent/issues/1799

PR #1818, opened 2024-03-25
Updating range for otel-autoconfigure verify instrumentation

meiao:
### Overview
The original range for which the instrumentation applies was missing a few versions.


deleonenriqueta (approved):

meiao:
Pekko AIT failed, but it is being disabled in a different PR.

PR #1827, opened 2024-03-29
1793 dynamo db double externals

jbedell-newrelic:
Mark the DB operations as leaf=true to prevent the subsequent (and duplicate) HTTP external from being recorded.


jtduffy (approved):
Maybe change the PR description to `leaf=true`

Issue #1800, opened 2024-03-14
Support external calls inside Spring Reactor call chains


jbedell-newrelic:
Result of https://github.com/newrelic/newrelic-java-agent/issues/1780

The issue is definitely a missed thread hop.  A simple repro app can be created to reproduce the issue.  Something along the lines of:
`client.callExternal1()
.flatMapMany(this::callExternal2)
.collectList()
.block();`

In this example the external call to External1 will show up in the trace, but the call to External2 will not.

On investigation, the tokenLift() method is not being executed in this scenario, nor is the `Hooks.onEachOperator` being called.

jbedell-newrelic:
Resolved by: https://github.com/newrelic/newrelic-java-agent/pull/1828

PR #1828, opened 2024-03-29
1780 spring reactor chaining

jbedell-newrelic:
This instrumentation should catch call chaining in regards to external calls being made in those chains.


jbedell-newrelic:
Resolves: https://github.com/newrelic/newrelic-java-agent/issues/1800

PR #1868, opened 2024-04-19
Linking DynamoDB queries to infra's entities

### LINKED ISSUE #1802: Add span attribute support for AWS_DYNAMO_DB_TABLE
### Description
Want to populate a `cloud.resource_id` attribute for APM spans accessing DynamoDB.

### Acceptance Criteria
Need to test end-to-end. Scenarios to explicitly test for:

- APM application+updated agent (but AWS CloudWatch and API Polling _not_ enabled)
- APM application+updated agent (AWS CloudWatch enabled)
- APM application+updated agent (API Polling)
- APM application+_non_ updated agent (eg: the `cloud.resource_id` attribute is missing)

May want to use separate NR accounts for each of the scenarios, to make it easier to see results.

Successful end-to-end solution will require one of either CloudWatch or API Polling to be enabled.

### Dependencies
APM Context Team will consume the attribute, generate necessary relationship files to be ingested

### Additional context
N/A

meiao:
### Overview
This PR adds the ability to specify a cloud resource id to some externals.
Having a cloud resource id in a span allows the queried infra entity to be linked to the APM entity.

Note that simply adding the cloud resource id is not enough to make the link. The type of entity must also be supported and instrumented.

To add the cloud resource id, use the new methods `cloudResourceId(String)` available in the builders for DatastoreParameters, MessageConsumeParameters and MessageProduceParameters.

This PR also populates this field for DynamoDB externals.

### Related Github Issue

Fixes #1802 

### Testing

Instrumentation tests were already disabled because they were flaky in GHA. Furthermore, while trying to run the tests, an issue was found with some dependencies. So for now the instrumentation tests are on hold.

A new agent integration test was added to test the DynamoDB instrumentation.

### Checks


jasonjkeller (dismissed):

newrelic-agent/src/main/java/com/newrelic/agent/service/analytics/SpanEventFactory.java
```diff
@@ -399,6 +406,14 @@ public SpanEventFactory setExternalParameterAttributes(ExternalParameters parame
             } else {
                 setAddress(datastoreParameters.getHost(), datastoreParameters.getPathOrId());
             }
+        } else if (parameters instanceof MessageProduceParameters) {
+            MessageProduceParameters messageProduceParameters = (MessageProduceParameters) parameters;
+            setCategory(SpanCategory.generic);
+            setCloudResourceId(messageProduceParameters.getCloudResourceId());
```
jasonjkeller, line 412: I found a bug where the `cloud.resource_id` attribute was never actually getting added to external spans for message producers & consumers when calling `cloudResourceId(arn)` on the `MessageProduceParameters`/`MessageConsumeParameters` builders with a valid ARN. I pushed up a fix and tests for that.

Technically, I think the same problem would exist if reporting an external with a `HttpParameters` type. This doesn’t look like it will be a problem for any of our AWS instrumentation as S3 is the only one that reports externals with `HttpParameters` as I don’t think we can actually build an ARN for S3 anyways. This might be a problem in the future though with other AWS SDKs or other vendors, so maybe we should also just `setCloudResourceId` on  `HttpParameters` as well?

kford-newrelic:
Move to `draft` since the approach of auto account ID discovery is still under discussion (and may have to change).

meiao:
This PR has been superseded by others that implement the same funcionality.

Issue #1780, opened 2024-03-04
Investigate missing async externals when Spring Reactor is in use


kford-newrelic:
### Description
Customer reported issue:

> We just tested the latest APM agent (8.8.1) and we’re still facing the instrumentation gap issue.For context, in all of our services the APM agent seems to be working fine, but we have 1 service which is a bit more complex (heavier reactor usage and more external webservice calls) in which we don’t see all the external calls. The call flow is typical :Spring mvc controller -> Business logic service (where we chain a lot of reactor Fluxes and Monos) -> External service webclientOur traces stop at the controller except for 1 webservice call that we execute at the very start of our reactor chain in the business logic service, all the other calls are absent from the traces.Relevant dependencies :
>
> - spring boot 2.5.14
> - spring mvc, webflux 5.3.20
> - reactor-netty-(http & core) 1.0.19
> - reactor-core (3.4.18)

### Acceptance Criteria
Discovery spike.

### Design Consideration/Limitations
Possible unexpected thread hop?

### Dependencies
N/A

### Additional context
Customer repro available

jbedell-newrelic:
Closing this Spike out with the results detailed in https://github.com/newrelic/newrelic-java-agent/issues/1800

jbedell-newrelic:
The fix here did resolve part of the issue, but more complex call-chaining is still not working.  So, I'm re-opening this ticket to investigate further.  We also believe this may be related to: https://github.com/newrelic/newrelic-java-agent/issues/1450

Issue #1450, opened 2023-08-11
Investigate gap in instrumentation with Spring WebFlux/Reactor when using `subscribeOn(Schedulers.boundedElastic())`


jasonjkeller:
This is a side issue that was noticed during the course of working on https://github.com/newrelic/newrelic-java-agent/issues/1073

Below is a Spring controller that repros the problem and here is a zip of a repro app: [newrelic-trace-on-util.zip](https://github.com/newrelic/newrelic-java-agent/files/12324913/newrelic-trace-on-util.zip)

The problem is that the `Transaction` is lost and thus not linked when calling the `response` method via the `/does-not-work` route. Specifically this seems to be a result of using `subscribeOn(Schedulers.boundedElastic())`. In the past, [we've seen cases where an unexpected thread hop occurred](https://newrelic.slack.com/archives/G96BPDE4Q/p1634570420269400?thread_ts=1634570350.269100&cid=G96BPDE4Q) due to the use of `subscribeOn(Schedulers.boundedElastic())` which caused the token linking failures.

```
package com.example.newrelictraceonutil;

import com.newrelic.api.agent.NewRelic;
import com.newrelic.api.agent.TracedMethod;
import com.newrelic.api.agent.Transaction;
import org.springframework.web.bind.annotation.GetMapping;
import org.springframework.web.bind.annotation.RestController;
import reactor.core.publisher.Mono;
import reactor.core.scheduler.Schedulers;

@RestController
public class ExampleController {

    @GetMapping(path = "/works")
    public Mono<String> works() {
        return Mono.fromCallable(() -> response("works"));
    }

    @GetMapping(path = "/does-not-work")
    public Mono<String> doesNotWork() {
        return Mono.fromCallable(() -> response("does-not-work"))
                .subscribeOn(Schedulers.boundedElastic());
    }

    /*
     * When hitting the /works route the transaction is properly renamed to works-custom-transaction-name as it
     * has reference to a valid TransactionApiImpl and DefaultTracer
     *
     * When hitting the /does-not-work route the transaction is NOT renamed as the Transaction context has been lost,
     * and it has reference to a NoOpTransaction and NoOpTracedMethod
     */
    private String response(String endpoint) {
        NewRelic.setTransactionName("custom", endpoint + "-custom-transaction-name");
        NewRelic.addCustomParameter("custom.param", endpoint);

        Transaction transaction = NewRelic.getAgent().getTransaction();
[...]

PR #1883, opened 2024-05-03
Resolves various thread hop problems with Spring Reactor call chaining.

jbedell-newrelic:
Grab a token when the SubscribeOnSubscriber is created and then link it when it is run().
Also copied this and the FluxMapFuseable instrumentation to the reactor-3.3.0 module.


jbedell-newrelic:
Resolves: https://github.com/newrelic/newrelic-java-agent/issues/1780
Resolves: https://github.com/newrelic/newrelic-java-agent/issues/1450

jtduffy (approved):

instrumentation/netty-reactor-0.9.0/src/main/java/reactor/core/publisher/SubscribeOnSubscriber_Instrumentation.java
```diff
@@ -0,0 +1,47 @@
...
+    Token token;
+    SubscribeOnSubscriber_Instrumentation(Publisher<? extends T> parent,
+            CoreSubscriber<? super T> actual,
+            Scheduler.Worker worker) {
+        if (NewRelic.getAgent().getTransaction() != null && token == null) {
+            token = NewRelic.getAgent().getTransaction().getToken();
+        }
+    }
+
+    public void run() {
```
meiao, line 32: Is this in a transaction?

Or is the method being instrumented after the first run? (due to token.link...() being called outside of a transaction)?

If it is the latter, we should add `@Trace(async=true)` to all these methods.

jbedell-newrelic: Looks like the is working as intended without the @Trace annotations added.

PR #1886, opened 2024-05-07
OTel span support

sdaubin:
### Overview
The java agent supports auto configuring the OTel SDK to report data to New Relic.  Spans that are created with the span api are orphaned from New Relic data structures though.  This returns spans backed by NR traces to generate metric and DT data.

This change hooks the `SdkTracerProvider.tracerBuilder` method and returns our tracer builder.  When spans are started, we attempt to create a tracer.  If successful, we return a `Span` implementation backed by the tracer, otherwise we return a no op span.

There is some special logic based on the span kind:

## consumer

We start an OtherTransaction for `consumer` span kind.

## server

We start a transaction for server spans and turn it into a web transaction.

## client

For client spans, we either turn it into a database span or an external based on the span attributes.

# Configuration

`opentelemetry.sdk.spans.enabled` can be use to disable the span behavior.
`opentelemetry.instrumentation.{instrumentation-scope}.enabled` can be used to disable spans for a specific Opentelemetry library

### Related Github Issue

### Testing

### Checks


agent-bridge/src/main/java/com/newrelic/agent/bridge/Instrumentation.java
```diff
@@ -59,6 +58,10 @@ ExitTracer createTracer(Object invocationTarget, int signatureId, boolean dispat
 
     ExitTracer createScalaTxnTracer();
 
+    default ExitTracer createTracer(String metricName, int flags) {
```
sdaubin, line 61: The `createSegment` API still generates an asynchronous tracer, which is not what we want.  And from the bridge there's no easy way to register a signature to get a `signatureId`.  This was the easiest way to get a normal, synchronous tracer.

jasonjkeller (approved):

jasonjkeller:
While this PR has been reviewed and approved, I'm changing it to a draft to prevent merging as there are ongoing discussions about the broader OTel strategy across APM and we want to make sure that all teams are in alignment on an approach.

jasonjkeller (approved):

PR #1887, opened 2024-05-07
Non-regex SQL checks

kanderson250:
This PR incorporates the changes from the custom jar we built to prevent unnecessary regex checks when parsing SQL. A customer had reported latency with r2dbc drivers, and removing the extra regex checks resolved their issue. 

Instead of performing a regex check for all operations, we now do a simple String `contains` on the operation name, before continuing with the match.

### DIFF
--- a/agent-bridge-datastore/src/main/java/com/newrelic/agent/bridge/datastore/R2dbcOperation.java
+++ b/agent-bridge-datastore/src/main/java/com/newrelic/agent/bridge/datastore/R2dbcOperation.java
@@ -39,7 +39,7 @@ public class R2dbcOperation {
                     for (Pattern pattern : operation.getValue()) {
                         Matcher matcher = pattern.matcher(strippedSql);
                         if (matcher.find()) {
-                            String model = matcher.groupCount() > 0 ? removeBrackets(unquoteDatabaseName(matcher.group(1).trim())) : "unknown";
+                            String model = removeBrackets(unquoteDatabaseName(matcher.group(1).trim()));
                             return new OperationAndTableName(operation.getKey(), VALID_METRIC_NAME_MATCHER.matcher(model).matches() ? model : "ParseError");
                         }
                     }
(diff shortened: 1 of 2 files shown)


jtduffy (approved):

agent-bridge-datastore/src/main/java/com/newrelic/agent/bridge/datastore/R2dbcOperation.java
```diff
@@ -30,14 +30,18 @@ public class R2dbcOperation {
...
-                    if(matcher.find()) {
-                        String model = matcher.groupCount() > 0 ? removeBrackets(unquoteDatabaseName(matcher.group(1).trim())) : "unknown";
-                        return new OperationAndTableName(operation.getKey(), VALID_METRIC_NAME_MATCHER.matcher(model).matches() ? model : "ParseError");
+                String opName = operation.getKey();
+                if (upperCaseSql.contains(opName)) { //non-regex check before pattern matching
+                    for (Pattern pattern : operation.getValue()) {
+                        Matcher matcher = pattern.matcher(strippedSql);
+                        if (matcher.find()) {
+                            String model = matcher.groupCount() > 0 ? removeBrackets(unquoteDatabaseName(matcher.group(1).trim())) : "unknown";
+                            return new OperationAndTableName(operation.getKey(), VALID_METRIC_NAME_MATCHER.matcher(model).matches() ? model : "ParseError");
```
meiao, line 43: I think we can prevent calling the matcher if model is "unknown".
Since the idea is to decrease the amount of regexes run, I think it makes sense to refactor this.

kanderson250: @meiao I went to update this and realized that line 42

```
String model = matcher.groupCount() > 0 ? removeBrackets(unquoteDatabaseName(matcher.group(1).trim())) : "unknown";
```

is redundant. `matcher.groupCount` counts how many capturing groups are in the pattern, not how many matches were found. Since all the patterns have one capturing group, this always evaluates to 1 and "unknown" is never reached. 

Removed in the next commit.

Issue #1876, opened 2024-04-26
Investigate relationship data that can be emitted from apps that use AWS_MQ_BROKER with Apache ActiveMQ


obenkenobi:
### Description
Trying to reduce the entity relationship data gaps when APM applications use AWS services. Need to explore what entity relationship data can be captured and emitted from the agent when `AWS_MQ_BROKER` is in use with ActiveMQ.

### Acceptance Criteria
Research spike.

### Design Consideration/Limitations
Try to capture the ARN in some way if possible but if not, other data - eg: host, port, etc - might be useable for building entity relationships.

### Dependencies
N/A

### Additional context
Once the available data is better understood, we need to flesh out the details of additional work needed to get that into the agent and documented.

obenkenobi:
https://github.com/newrelic/newrelic-java-agent/issues/1832#issuecomment-2079553059

Issue #1832, opened 2024-04-02
Investigate relationship data that can be emitted from apps that use AWS_MQ_BROKER with RabbitMQ


kford-newrelic:
### Description
Trying to reduce the entity relationship data gaps when APM applications use AWS services. Need to explore what entity relationship data can be captured and emitted from the agent when `AWS_MQ_BROKER` with RabbtMQ is in use.

### Acceptance Criteria
Research spike.

### Design Consideration/Limitations
Try to capture the ARN in some way if possible but if not, other data - eg: host, port, etc - might be useable for building entity relationships.

### Dependencies
N/A

### Additional context
Once the available data is better understood, we need to flesh out the details of additional work needed to get that into the agent and documented.

obenkenobi:
AWS MQ is both RabbitMQ and ActiveMQ compatible. Thus the following needs to be investigated:

Rabbit MQ:
  - The RabbitMQ's AMQP client instrumentation (of which we already instrument)
  - RabbitMQ's management api
 Active MQ:
  - The AMQP client 
  - The MQTTP client
  - Apps using STOMP (depends on Spring's web-socket api)
  - ActiveMQ's management api

Doc for using JMS with active MQ as it describes each possible client for ActiveMQ: https://docs.aws.amazon.com/amazon-mq/latest/developer-guide/amazon-mq-working-java-example.html

obenkenobi:
[Link](https://github.com/orgs/newrelic/projects/87/views/1?pane=issue&itemId=61150787) to the Active MQ Ticket for AWS MQ

PR #1898, opened 2024-05-17
Linking AWS MQ infrastructure entities

obenkenobi:
### Overview

> This PR uses the changes from the [DynamoDB PR](https://github.com/newrelic/newrelic-java-agent/pull/1868). Merge the DynamoDB PR first before merging this PR.

This PR adds new span attributes containing endpoint information from a message broker. This is so the Java Agent can properly link with infrastructure entities generated from AWS MQ.

The new span attributes are the following:

- `server.address` which contain the message broker host 
- `server.port` which contain the message broker port.
- `messaging.rabbitmq.destination.routing_key` which is the routing key for a RabbitMQ operation
- `messaging.destination.name` which will be the destination for a RabbitMQ operation. For producers this will be the exchange name. For consumers this is the queue name.
- `messaging.destination_publish.name` is an otel attribute for message consumers used to indicate what the destination name for the producer is. In the case of RabbitMQ, this is the exchange name.

The externals API has new optional parameters to add host and port for message broker spans, via the method `instance(host,port)`.

For example
```java
ExternalParameters messageProduceParameters = MessageProduceParameters
    .library("JMS")
    .destinationType(DestinationType.NAMED_QUEUE)
    .destinationName("MessageQueue")
    .outboundHeaders(outboundMessageHeaders)
    .instance("localhost", 8088)
    .build();
NewRelic.getAgent().getTracedMethod().reportAsExternal(messageProduceParameters);

ExternalParameters messageConsumeParameters = MessageConsumeParameters
    .library("JMS")
    .destinationType(DestinationType.NAMED_TOPIC)
    .destinationName("MessageTopic")
    .inboundHeaders(inboundMessageHeaders)
    .instance("localhost", 8088)
    .build();
NewRelic.getAgent().getTracedMethod().reportAsExternal(messageConsumeParameters);
```

Our JMS (if ActiveMQ is used) and RabbitMQ instrumentation automatically add the span attributes.

To get host and port from ActiveMQ, a new instrumentation module was added called `activemq-client-5.8.0`

On top of that, the`rabbit-amqp` and `jms` instrumentation was altered to automatically include host and port span attributes.

For our RabbitMQ instrumentation, the existing agent attributes `message.routingKey` and `message.queueName` are also added to spans.

In addition the instrumentation was revised so instrumentation module names better match the versions they apply to and to allow host and port information to be reported.

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/1876
https://github.com/newrelic/newrelic-java-agent/issues/1832


newrelic-api/src/main/java/com/newrelic/api/agent/MessageProduceParameters.java
```diff
@@ -128,6 +172,24 @@ public interface OutboundHeadersParameter {
...
+         * Set the cloud provider's id for the message queue.
+         * This method is optional and can be bypassed by calling build directly.
+         */
+        Build cloudResourceId(String cloudResourceId);
+
+        /**
+         * Set the host name for the message queue.
+         * This method is optional and can be bypassed by calling build directly.
+         */
+        Build host(String host);
```
meiao, line 185: DatabaseParameters has a single method for setting the host and port.
Do you think it makes sense to replicate that here?

agent-bridge/src/main/java/com/newrelic/agent/bridge/DefaultCollectionFactory.java
```diff
@@ -11,14 +11,33 @@
...
 
     @Override
     public <K, V> Map<K, V> createConcurrentWeakKeyedMap() {
         return Collections.synchronizedMap(new WeakHashMap<K, V>());
     }
 
+    @Override
+    public <K, V> Function<K, V> memoize(Function<K, V> loader, int maxSize) {
+        Map<K, V> map = new ConcurrentHashMap<>();
+        return k -> map.computeIfAbsent(k, k1 -> {
```
sdaubin, line 34: It's a little more code, but you could do this: 

        return k -> {
            if (map.size() >= maxSize) {
                V value = map.get(k);
                return value == null ? loader.apply(k) : value;
            }
            return map.computeIfAbsent(k, loader::apply);
        };

The performance would be slightly better once you reach the `maxSize` because you wouldn't have to keep creating iterators and removing elements from the map.

meiao: This class is never used by the agent.

It is just there as a placeholder to prevent blowing up an application that uses the agent bridge without the agent.

But indeed, the suggestion is a good one. I will incorporate that in the parent of this PR.

obenkenobi:
aws-mq-rebased is where the MQ changes are pulled to to prevent pollution from dynamo db changes

Issue #1702, opened 2024-01-24
Memory Leak due to newly introduced Postgresql Instrument in 8.8 release of New Relic Java Agent


Maggie552:
We have encountered a memory leak issue when we began using the [new relic 8.8.0 release agent](https://docs.newrelic.com/docs/release-notes/agent-release-notes/java-release-notes/java-agent-880/) which includes the change 
 > Instrument r2dbc-postgresql 0.9.2 till latest [1413](https://github.com/newrelic/newrelic-java-agent/pull/1556).

## Description
When we began use of the 8.8.0 release of the new relic agent we found that our used heap would gradually increase over 3 days until maximum heap limit was reached and we would have to restart our application. This would happen repeatedly and consistently.

## Expected Behavior
N/A

## [Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results
We diagnosed the issue looking at our new relic dashboard and by analysing heap dumps in Eclipse Memory Analyzer. 

In this screen shot from new relic you can see the used heap (the blue line) gradually increasing towards the maximum heap (the green line).
![MicrosoftTeams-image (6)](https://github.com/newrelic/newrelic-java-agent/assets/60889664/7d7897d2-0d65-4943-8096-8637835ac904)

In this screen shot you can see the results of us analysing our heap dump in Eclipse Memory Analyzer and see that the Leak suspect report points towards a package belonging to new relic in relation to postgreql.

![MicrosoftTeams-image (7)](https://github.com/newrelic/newrelic-java-agent/assets/60889664/474ad817-5148-4022-a450-8ae3ae895c82)

We tied this to the newly introduced postgresql instrument as part of the 8.8.0 release for the new relic agent and disabled the instrument, this has stopped the memory leak. However this means we have lost the functionality of this instrument.
## Steps to Reproduce
Run a java web application with connections to a postgresql db using the reactive client(r2dbc) doing simple selects, with the new relic agent version 8.8.0 monitoring metrics. Perform a heap dump and analyse in Eclipse Memory Analyser, you will find the leak suspect pointing towards class com.newrelic.weave.io.r2dbc.postgresql.client.ReactorNettyClient. 

[...]

PR #1916, opened 2024-05-30
R2dbc postgresql 0.9.2 memleak fix

obenkenobi:
### Overview
Changes to the Postgres R2DBC 9.2+ instrumentation that improve memory usage. In particular, runs of this branch of the agent did not cause a growth in heap usage compared to the old agent over time.

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/1702


obenkenobi:
While this memory leak requires running the app for a long time, one thing of note here is that even running a [test app](https://github.com/Obenkenobi-New-Relic-Test-Apps/spring-r2dbc-postgres-stress-app) for almost 19 hours shown the JVM had to more and more increase the committed heap size while the agent with my fix did not (the only exception for the new agent was when my app ran for 3 hours).

image

jtduffy (approved):

obenkenobi:
At a higher load, the app with the old agent reached an out of memory condition and stopped reporting to New Relic.

The agent with my fix did not reach such condition.

image

PR #1950, opened 2024-06-14
Jetty12 core server transactions

jbedell-newrelic:
Start transactions inside the core Jetty server (rather than just servlets).


jtduffy (approved):

obenkenobi (approved):
This did not break async transaction tests, good work!

Issue #1475, opened 2023-09-06
Add New Instrumentation for Jedis-5.0.0


deleonenriqueta:
### Description
Current instrumentation modules are incompatible with the latest jedis version

meiao:
Weave violations
     ReferenceViolation{type=INVALID_REFERENCE, weaveClass=redis/clients/jedis/Connection, originalClass=redis/clients/jedis/Protocol$Command, violationMessage=missing field. weave code expects field: QUITLredis/clients/jedis/Protocol$Command;}
     WeaveViolation{type=EXPECTED_NEW_FIELD_ANNOTATION, clazz=redis/clients/jedis/JedisPubSub, field=client}
     WeaveViolation{type=METHOD_NEW_CALL_ORIGINAL_UNSUPPORTED, clazz=redis/clients/jedis/JedisPubSub, method=proceed(Lredis/clients/jedis/Connection;[Ljava/lang/String;)V}
     WeaveViolation{type=METHOD_NEW_NON_PRIVATE_UNSUPPORTED, clazz=redis/clients/jedis/JedisPubSub, method=proceed(Lredis/clients/jedis/Connection;[Ljava/lang/String;)V}
     WeaveViolation{type=METHOD_NEW_CALL_ORIGINAL_UNSUPPORTED, clazz=redis/clients/jedis/JedisPubSub, method=proceedWithPatterns(Lredis/clients/jedis/Connection;[Ljava/lang/String;)V}
     WeaveViolation{type=METHOD_NEW_NON_PRIVATE_UNSUPPORTED, clazz=redis/clients/jedis/JedisPubSub, method=proceedWithPatterns(Lredis/clients/jedis/Connection;[Ljava/lang/String;)V}
     WeaveViolation{type=METHOD_NEW_INVOKE_UNSUPPORTED, clazz=redis/clients/jedis/JedisPubSub, method=onMessage(Ljava/lang/String;Ljava/lang/String;)V}
     WeaveViolation{type=METHOD_NEW_CALL_ORIGINAL_UNSUPPORTED, clazz=redis/clients/jedis/JedisPubSub, method=onMessage(Ljava/lang/String;Ljava/lang/String;)V}
     WeaveViolation{type=METHOD_NEW_NON_PRIVATE_UNSUPPORTED, clazz=redis/clients/jedis/JedisPubSub, method=onMessage(Ljava/lang/String;Ljava/lang/String;)V}
[...]

PR #1969, opened 2024-07-03
Add Jedis-5.0.0 Instrumentation Module

deleonenriqueta:
Resolves Issue [#1475](https://github.com/newrelic/newrelic-java-agent/issues/1475)


jbedell-newrelic (approved):

instrumentation/jedis-5.0.0/build.gradle
```diff
@@ -0,0 +1,24 @@
...
+dependencies {
+    implementation(project(":agent-bridge"))
+    implementation(project(":agent-bridge-datastore"))
+    implementation(project(":newrelic-api"))
+    implementation(project(":newrelic-weaver-api"))
+    implementation("redis.clients:jedis:5.0.0")
+}
+
+verifyInstrumentation {
+    passes 'redis.clients:jedis:[5.0.0,6.0.0)'
```
meiao, line 14: why not use
```
passesOnly 'redis.clients:jedis:[5.0.0,)
```
If it works for 6.0.0, then we won't have to change this. If it doesn't, the verify instrumentation will tell us about it.

instrumentation/jedis-5.0.0/src/main/java/redis/clients/jedis/Connection_Instrumentation.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Trace
+    public void sendCommand(final ProtocolCommand cmd, final String... args) {
+        Weaver.callOriginal();
+        if (args != null && args.length > 0) {
+            updateDbIndex(cmd, args[0]);
+        }
+    }
+
+    @Trace(leaf = true)
```
meiao, line 54: What is the reasoning behind this method being a leaf and the others not. It would be good to document so later we are not guessing.

instrumentation/jedis-5.0.0/src/main/java/redis/clients/jedis/JedisPubSubBase_Instrumentation.java
```diff
@@ -0,0 +1,76 @@
...
+import com.newrelic.api.agent.weaver.Weave;
+import com.newrelic.api.agent.weaver.Weaver;
+
+@SuppressWarnings({ "ResultOfMethodCallIgnored", "unused" })
+@Weave(type = MatchType.ExactClass, originalName = "redis.clients.jedis.JedisPubSubBase")
+public class JedisPubSubBase_Instrumentation {
+
+  private volatile Connection client;
+
+  public final void proceed(Connection client, Object... channels) {
```
meiao, line 17: does this method need instrumentation? There is no added code and it is not called inside this class.

If it is here so this does not apply to a previous version, you can make this class abstract and also this method.

Else the weaver ends up adding some code to the method IIRC.

instrumentation/jedis-5.0.0/build.gradle
deleonenriqueta: updated!

instrumentation/jedis-5.0.0/src/main/java/redis/clients/jedis/Connection_Instrumentation.java
```diff
@@ -0,0 +1,90 @@
+/*
+ *
+ *  * Copyright 2020 New Relic Corporation. All rights reserved.
```
jasonjkeller, line 3: If this is a net new instrumentation module then the year should be 2024.

The other file (`JedisPubSubBase_Instrumentation.java`) also needs a copyright header.

jasonjkeller (changes requested):

instrumentation/jedis-5.0.0/src/main/java/redis/clients/jedis/Connection_Instrumentation.java
deleonenriqueta: Updated!

jasonjkeller (approved):

PR #2021, opened 2024-08-22
Flyway migration events

### LINKED ISSUE #2008: Spring - Emit events for flyway based DB migrations
For Flyway migrations that happen during application startup, send an event for each migration file that's evaluated, including those migrations that have already been applied. Some attributes that might be candidates to be attached to the event:
- Date/time of migration
- migration filename
etc..

jtduffy:
Resolves #2008 

This instrumentation module will report migration report events to New Relic during Flyway migrations initiated via code, including when running properly configured Spring/Spring Boot applications.

The following Flyway migration events are reported:
- Event.AFTER_EACH_MIGRATE
- Event.AFTER_EACH_MIGRATE_ERROR
- Event.AFTER_EACH_UNDO
- Event.AFTER_EACH_UNDO_ERROR

The custom event type name is `FlywayMigration`, with the following attributes:
- migrationSuccess: `true` if the migration was successful
- migrationFilePhysicalLocation: Path/filename of the migration script
- migrationChecksum: Calculated checksum of the migration
- migrationVersion: Version portion of the migration script filename
- migrationScriptName: Name portion of the migration script filename
- migrationEvent: The underlying Flyway migration event


instrumentation/flyway-core-8.0.0/src/main/java/com/nr/agent/instrumentation/FlywayUtils.java
```diff
@@ -0,0 +1,54 @@
+/*
+ *
+ *  * Copyright 2020 New Relic Corporation. All rights reserved.
```
meiao, line 3: year is wrong

meiao (approved):

PR #2022, opened 2024-08-22
Add support for using env variable for config file location

### LINKED ISSUE #2020: Allow the environment variable NEWRELIC_FILE to specify the location of the agent config file
Currently, to specify the location of the agent config file, the system property `newrelic_config_file` must be used.  Implement a change so that the environment variable `NEWRELIC_FILE` can be used.

This follows the [logic](https://docs.newrelic.com/docs/apm/agents/java-agent/configuration/java-agent-configuration-config-file/#config-options-precedence) of other config options, where environment variable, system property and yml config can be used for configuration. 

Required for SuperAgent / Agent Control implementation.

jtduffy:
Resolves #2020 

Now supports the use of the environment variable `NEWRELIC_FILE` to specify the location of the agent configuration file.

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/config/ConfigFileHelper.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/config/ConfigFileHelper.java
@@ -8,6 +8,7 @@
 package com.newrelic.agent.config;
 
 import com.newrelic.agent.DebugFlag;
+import org.jetbrains.annotations.Nullable;
 
 import java.io.File;
 import java.text.MessageFormat;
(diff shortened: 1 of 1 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/config/ConfigFileHelper.java
```diff
@@ -8,6 +8,7 @@
 package com.newrelic.agent.config;
 
 import com.newrelic.agent.DebugFlag;
+import org.jetbrains.annotations.Nullable;
```
meiao, line 11: We probably don't want to use this annotation.

jtduffy: Nah, it snuck in there when I had IJ do a refactor. Deleted.

meiao (approved):

PR #2025, opened 2024-08-23
Slick 3.5.0 bug fix

### LINKED ISSUE #1996: Instrumentation is bugged for Slick 3.5.0
## Description

I am trying to upgrade a service to Slick 3.5.0. When I run this service with the updated Slick & New Relic Java agent 8.13.0, some Slick queries just…give no response. I believe the instrumentation needs an update.

## Expected Behavior

Previously, the service depended on Slick 3.3.3 and worked with the agent enabled.

## [Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results

 - 

## Steps to Reproduce

So far I have only observed the issue in a large Akka service. Adding 

```yaml
  class_transformer:

    excludes:
      slick/.*
```

to my configuration results in the service behaving as expected.

## Your Environment
[TIP]:  # ( Include as many relevant details about your environment as possible including the running version of New Relic software, JVM vendor and version, framework versions, and any relevant configuration. )

JVM 11, Akka 2.9.3, Slick 3.5.0, MySQL or H2 database.

kanderson250:
Resolves #1996 

Previously, our slick instrumentation wrapped the slick `AsyncExecutor` in our own `NewRelicAsyncExecutor`. However, in slick 3.5.0, the concrete implementation `AsyncExecutorImpl` was introduced, which meant our wrapper ignored `AsynExecutorImpl's` overridden methods. One consequence of that was incorrect creation of `PrioritizedRunnables` and disruption of the executor's work queue, causing the application to stall. 

We also recognized that wrapping slick's `PrioritizedRunnable` in a `Runnable` (only for slick to wrap it *back* into a `PrioritizedRunnable`) was less than ideal. 

This instrumentation avoids wrapping anything so that we don't have the type issues from before. Instead, we now instrument the `prioritizedRunnable` method on `AsyncExecutor` and wrap its generic `run` Function parameter. 

Thanks to @jmreardon-yoppworks for filing the initial bug report and providing a repro to aid our development work. 

** Tests **
There are a few new unit tests, plus an [AIT branch](https://github.com/newrelic/java-agent-integration-tests/blob/1a6c8d13752a2cd4475b01a8d5cb9fe57d4bfd34/tests/java/functionality/framework/play/play2.py#L255) that has been updated with the new slick module.


instrumentation/slick-2.12_3.2.0/src/main/scala/slick/util/SkipDefaultAsyncExecutor.java
```diff
@@ -0,0 +1,7 @@
+package slick.util;
```
jtduffy, line 1: Double check for current copyright headers in each .java file

jtduffy (approved):

PR #2052, opened 2024-09-16
Fix an issue where the Kinesis instrumentation is generating ERROR logs due to a NullPointerException

obenkenobi:
### Overview
Fixes an issue where ERROR logs were generated by the Kinesis instrumentation due to a NullPointerException.
```
2024-09-16T13:37:08,315-0400 [10984 66] com.newrelic ERROR: An error occurred recording external metrics for class putRecord : java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "value" is null

```

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/tracers/DefaultTracer.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/tracers/DefaultTracer.java
@@ -854,7 +854,9 @@ public class DefaultTracer extends AbstractTracer {
 
     private void recordFaasAttributes(CloudParameters cloudParameters) {
         setAgentAttribute(AttributeNames.CLOUD_PLATFORM, cloudParameters.getPlatform());
-        setAgentAttribute(AttributeNames.CLOUD_RESOURCE_ID, cloudParameters.getResourceId());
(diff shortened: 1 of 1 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/tracers/DefaultTracer.java
```diff
@@ -854,7 +854,9 @@ private void recordMessageBrokerMetrics(MessageConsumeParameters messageConsumeP
 
     private void recordFaasAttributes(CloudParameters cloudParameters) {
         setAgentAttribute(AttributeNames.CLOUD_PLATFORM, cloudParameters.getPlatform());
```
jasonjkeller, line 856: Is it possible that `cloudParameters.getPlatform()` could potentially cause an NPE as well?

jasonjkeller (approved):

newrelic-agent/src/main/java/com/newrelic/agent/tracers/DefaultTracer.java
obenkenobi: No because every instance of us creating CloudParameters fills the platform field

PR #2077, opened 2024-09-26
Actuator endpoint naming for Spring Boot 3

### LINKED ISSUE #2035: JAVA APM agent - support Spring Actuator
## Is your feature request related to a problem? Please describe.
Customer is using Spring Actuator for various needs, especially for healthcheck of the API service.

## Feature Description
Customer would like to have OOTB instrumentation for Spring Actuator

## Describe Alternatives
N/A

## Additional context
N/A

## Priority
Please help us better understand this feature request by choosing a priority from the following options: 
[Really Want]

jtduffy:
### Naming of actuator endpoints for Spring Boot 3
Resolves #2035 

By default, built-in actuator endpoints and custom actuator endpoints (using the @Endpoint annotation
and it's subclasses) will all be named as "OperationHandler/handle" in New Relic. Activating this
module will result in the transaction name reflecting the actual base actuator endpoint URI.
For example, invoking "/actuator/loggers" or "actuator/loggers/com.newrelic" will result in the
transaction name "actuator/loggers (GET)". This is to prevent MGI.

To activate actuator naming, set the following configuration to true:
- `class_transformer.name_actuator_endpoints`

The default value is false.


instrumentation/spring-boot-actuator-3.0.0/src/main/java/org/springframework/boot/actuate/endpoint/web/servlet/AbstractWebMvcEndpointHandlerMapping_Instrumentation.java
```diff
@@ -0,0 +1,44 @@
...
+@Weave(type = MatchType.BaseClass, originalName = "org.springframework.boot.actuate.endpoint.web.servlet.AbstractWebMvcEndpointHandlerMapping")
+public class AbstractWebMvcEndpointHandlerMapping_Instrumentation {
+    @Weave(type = MatchType.ExactClass, originalName = "org.springframework.boot.actuate.endpoint.web.servlet.AbstractWebMvcEndpointHandlerMapping$OperationHandler")
+    private static final class OperationHandler_Instrumentation {
+        @Trace
+        Object handle(HttpServletRequest request, Map<String, String> body) {
+            if (SpringActuatorUtils.isActuatorEndpointNamingEnabled) {
+                Transaction transaction = NewRelic.getAgent().getTransaction();
+
+                if (transaction != null) {
```
meiao, line 30: The API call will always return not null.
You could check the class type
```
if (! transaction instanceof NoOpTransaction)
```
Or you can use the agent bridge

```
Transaction transaction = AgentBridge.getAgent().getTransaction(false);
```

instrumentation/spring-boot-actuator-3.0.0/src/main/java/org/springframework/boot/actuate/endpoint/web/servlet/AbstractWebMvcEndpointHandlerMapping_Instrumentation.java
```diff
@@ -0,0 +1,44 @@
...
+                    String uri = SpringActuatorUtils.normalizeActuatorUri(request.getRequestURI());
+                    String reportablePrefix = SpringActuatorUtils.normalizeActuatorUri(uri);
```
meiao, line 32: won't this return the same string?

jtduffy: Duh, yeah good catch, I left that in by accident when I did a refactor.

instrumentation/spring-boot-actuator-3.0.0/src/main/java/org/springframework/boot/actuate/endpoint/web/servlet/AbstractWebMvcEndpointHandlerMapping_Instrumentation.java
jtduffy: Good catch

meiao (approved):

instrumentation/spring-boot-actuator-3.0.0/src/main/java/org/springframework/boot/actuate/endpoint/web/servlet/AbstractWebMvcEndpointHandlerMapping_Instrumentation.java
```diff
@@ -0,0 +1,44 @@
...
+import com.newrelic.api.agent.weaver.MatchType;
+import com.newrelic.api.agent.weaver.Weave;
+import com.newrelic.api.agent.weaver.Weaver;
+import com.nr.agent.instrumentation.actuator.SpringActuatorUtils;
+import jakarta.servlet.http.HttpServletRequest;
+
+import java.util.Map;
+
+@Weave(type = MatchType.BaseClass, originalName = "org.springframework.boot.actuate.endpoint.web.servlet.AbstractWebMvcEndpointHandlerMapping")
+public class AbstractWebMvcEndpointHandlerMapping_Instrumentation {
```
zboris, line 23: what about reactive mappings?
org.springframework.boot.actuate.endpoint.web.reactive.AbstractWebFluxEndpointHandlerMapping$ReactiveWebOperationAdapter

jtduffy: @zboris   Thanks for the suggestion. I'll get an issue created so it gets on our backlog.

PR #2081, opened 2024-10-01
Cloud API

### LINKED ISSUE #2003: Enhance Entity Relationships - design new agent API for AWS account ID
### Description
Need to define a way for a customer to programmatically provide the agent with AWS account ID info, for later use in emitting data for relationship matching for certain services like DynamoDB.

Could be as simple as the `account-id` alone or as complex as a full `ARN` or `ARN-like` data element.

Should assume that a customer application could be invoking one **or more** AWS services, each using a different account ID.

### Acceptance Criteria
Need to define an agent API call that customers can make that will then allow their apps to use a service like DynamoDB and have that relationship get matched/created.

Will need to be implemented across all language agents.

### Dependencies
N/A

### Additional context
Should get familiar with how customers will do this for Open Telemetry.

meiao:
### Overview
Introduces the Cloud API, which allows cloud account information to be passed to the agent.

This information is required by some instrumentation modules to be able to form cloud resource identifiers which are used to link a monitored application to the cloud resources it uses.

### Related Github Issue
Fixes #2003 

### Checks

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/cloud/CloudAccountInfoCache.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/cloud/CloudAccountInfoCache.java
@@ -18,6 +18,7 @@ import com.newrelic.api.agent.NewRelic;
 import java.util.Collections;
 import java.util.EnumMap;
 import java.util.Map;
+import java.util.logging.Level;
 
 /**
  * This class implements the account info methods from the Cloud API.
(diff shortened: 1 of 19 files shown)


jbedell-newrelic (approved):

newrelic-agent/src/main/java/com/newrelic/agent/cloud/CloudAccountInfoCache.java
```diff
@@ -0,0 +1,93 @@
...
+    public void setAccountInfo(Object sdkClient, CloudAccountInfo cloudAccountInfo, String value) {
+        if (sdkClient == null) {
+            return;
+        }
+        if (value == null) {
+            Map<CloudAccountInfo, String> accountInfo = cache.getIfPresent(sdkClient);
+            if (accountInfo != null) {
+                accountInfo.remove(cloudAccountInfo);
+            }
+        }
```
jtduffy, line 52: Should there be a short-circuit return here or is it ok to have a `cloudAccountInfo` with a null value?

jtduffy (approved):

newrelic-agent/src/main/java/com/newrelic/agent/cloud/CloudAccountInfoCache.java
meiao: yes

PR #2114, opened 2024-10-29
Sanitize environment

### LINKED ISSUE #2110: Implement Obfuscate JVM Args Config Feature
Implement a feature to satisfy #2033. 

The feature should support the obfuscation of values of JVM properties after the `=`.  For example: `-Dprop=12345` will now be sent as `-Dprop=obfuscated`. 

This will be configurable via an optional block and allow list.

For example, with the following yml config, 

```
obfuscate_jvm_props:
    allow: -DpropA*, -DpropB*
    block: -DpropB.extended
```
and JVM startup args `-DpropA=one, -DpropB=two, -DpropB.extended=three, -DpropC=four`

We will now send `-DpropA=one, -DpropB=two, -DpropB.extended=obfuscated, -DpropC=obfuscated`

kanderson250:
Resolves #2110 

## Description

This PR introduces a new configuration stanza `obfuscate_jvm_props` to better protect our customers from sending sensitive data in their JVM args. 

Prior to this feature, all JVM args were sent by default on agent startup, and were not filtered or checked in any way. Now, we will obfuscate any data contained behind an `=` in a JVM argument by default (with a few exceptions, see below). 

## Details 

New .yml configuration stanza with descriptions: 

```
obfuscate_jvm_props:
  enabled: If true (default), obfuscate all JVM props data after =, except the default allows set by the agent.
          If false, leave all JVM props as-is. 
  allow: An optional list of comma-separated property names to send unobfuscated. 
          Can be an exact match for the prop name, or wildcarded in the last character. 
  block: An optional list of comma-separated property names to always send obfuscated. 
          Can be an exact match for the prop name, or wildcarded in the last character. 
```
**Default allows set by the agent**: The agent will send any standard or extended JVM prop (those with an `-X*` prefix) as-is, even if they contain an `=` sign. This was done because these props should not contain sensitive data and their values may be used by other New Relic teams, such as the BMDS project. 

To keep the block/allow list conventions consistent with other similar config settings (such as attribute includes/excludes), priority is always given to the more-specific rule (`-DpropB.extended` will take precedence over `-DpropB`, regardless of whether it belongs to the block or allow list).  

This feature has no affect on JVM args that do not contain an `=`, such as `-javaagent`. 

## Example

Setting: 

```
  obfuscate_jvm_props:
    allow: -Dprop.A, -Dprop.B*
    block: -Dprop.B.extended
```

And starting petclinic with: 

```
java -javaagent:newrelic/newrelic.jar -Xms2G -Xmx5G -Dprop.A=val1 -Dprop.B=val2 -Dprop.B.other=val3 -Dprop.B.extended=val4 -DpropC = val5 -XX:+HeapDumpOnOutOfMemoryError -XX:ActiveProcessorCount=1 -jar build/libs/spring-petclinic-3.0.0.jar
```

Sends up to the NR UI:

```
-javaagent:newrelic/newrelic.jar 
-Xms2G 
-Xmx5G 
-Dprop.A=val1 
-Dprop.B=val2 
-Dprop.B.other=val3
-Dprop.B.extended=obfuscated
-DpropC = obfuscated
-XX:+HeapDumpOnOutOfMemoryError 
-XX:ActiveProcessorCount=1
```

## Implementation

The PR introduces a new config class, `ObfuscateJvmPropsConfig`. It enhances the existing `Environment` class using an `ExcludesIncludesFilter` to apply the appropriate rules to JVM properties.


newrelic-agent/src/main/java/com/newrelic/agent/config/AgentConfigImpl.java
```diff
@@ -1241,6 +1250,10 @@ public AttributesConfig getAttributesConfig() {
         return attributesConfig;
     }
 
+    @Override
+
+    public ObfuscateJvmPropsConfig getObfuscateJvmPropsConfig() {return obfuscateJvmPropsConfig;}
```
jtduffy, line 1255: nit: 
```
{
    return obfuscateJvmPropsConfig;
}
```

jtduffy (changes requested):

newrelic-agent/src/main/java/com/newrelic/agent/config/ObfuscateJvmPropsConfigImpl.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.*;
+
+public class ObfuscateJvmPropsConfigImpl extends BaseConfig implements ObfuscateJvmPropsConfig {
+
+    private static final String SYSTEM_PROPERTY_ROOT="newrelic.config.obfuscate_jvm_props.";
+    private static final String ALLOW = "allow";
+    private static final String BLOCK = "block";
+    private static final String ENABLED = "enabled";
+    private static final boolean DEFAULT_ENABLED = true;
+    private static final Set<String> DEFAULT_ALLOW = new HashSet<>();
```
meiao, line 12: Instead of using a HashSet, use a SingletonSet. For things that are going to ever be a single element, this allocates less memory and executes faster.
```suggestion
    private static final Set<String> DEFAULT_ALLOW = Collections.singleton("-X*");
```
If we ever add more to the list, then it will have to go back to your implementation.

meiao (approved):

meiao (approved):

jtduffy (approved):

PR #2128, opened 2024-11-07
Use recordResponseTimeMetric instead recordMetric in NewRelic agent.

Matzz:
It seams that 
```
    public static void recordResponseTimeMetric(String name, long millis) {
        getAgent().getMetricAggregator().recordMetric(name, millis);
    }
```
recordResponseTimeMetric calls recordMetric in MetricAggregator instead of recordResponseTimeMetric.


Matzz:
@jtduffy I just started to wonder if I introduced backward compatibility break here.
As far as I understand when we call `MetricAggregator.recordResponseTimeMetric` the values are converted from millis to seconds. Is that the same case for `recordMetric`? What is the difference between `recordMetric` and `recordResponseTimeMetric`?
Does changing `recordMetric` to `recordResponseTimeMetric` could break some existing charts?

PR #2133, opened 2024-11-08
Add jvm prop settings to example yml file

kanderson250:
Updates the sample yml file with `send_jvm_props` and `obfuscate_jvm_props`

### DIFF
--- a/newrelic-agent/src/main/resources/newrelic.yml
+++ b/newrelic-agent/src/main/resources/newrelic.yml
@@ -356,6 +356,24 @@ common: &default_settings
     # the regex if you have more complex start head elements (attributes for example). The defined pattern is case-insensitive.
     tag_lib_head_pattern: '<head>'
 
+  # By default, the agent sends JVM input arguments to New Relic, where they are visible as Environment data.
+  # Set to false to disable the sending of JVM props.
+  send_jvm_props: true
+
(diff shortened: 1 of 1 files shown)


newrelic-agent/src/main/resources/newrelic.yml
```diff
@@ -356,6 +356,24 @@ common: &default_settings
...
+  # Available since agent 8.16.0.
+  obfuscate_jvm_props:
+    #To disable this feature, and send all JVM props unobfuscated, set enabled to false. The default is true.
+    #enabled: true
+
+    # A comma separated list of JVM property names whose values should be sent to New Relic unobfuscated.
+    #allow: -Dsend.this.prop, -Dsend.this.prop.and.children*
+
+    # A comma separated list of JVM property names whose values should be sent to New Relic obfuscated.
+    #block: -Dhide.this.prop, -Xthis.one.too
```
jasonjkeller, line 375: Can you remove the dummy values from these configs and just leave them empty like below? We don't want to add any dummy values to the actual config file that customers will download, examples like that should only be included in documentation.

```
    # A comma separated list of JVM property names whose values should be sent to New Relic unobfuscated.
    #allow:

    # A comma separated list of JVM property names whose values should be sent to New Relic obfuscated.
    #block:
```

jasonjkeller (changes requested):

jasonjkeller (approved):

Issue #2186, opened 2024-12-27
Add metrics for relationship synthesis on kafka-clients-node instrumentation


meiao:
### Description
The original Kafka instrumentation, recorded the following metrics, that are used for relationship synthesis.

```
MessageBroker/Kafka/Nodes/<broker>
MessageBroker/Kafka/Nodes/<broker>/Produce/<topic>
MessageBroker/Kafka/Nodes/<broker>/Consume/<topic>
```

PR #2195, opened 2025-01-10
kafka-clients-metrics: Refactor node topic metrics

meiao:
## Overview
The previous code executed more verifications than needed for its functionality, it also stored duplicate data for those verifications. This PR simplifies the implementation by reducing the number of verifications and data stored.

## Functionality

The functionality being altered is to generate metrics (or event attributes) with the format:

1. MessageBroker/Kafka/Nodes/<node>
2. MessageBroker/Kafka/Nodes/<node>/<Produce|Consume>/<topic>

For each node a client connects to, a metric of the first format will be created. Then for the cartesian product of the sets of node x topic, a metric of the second format will be created.

So, for a producer client connecting  to nodes (n1, n2) and producing for the topics (t1, t2), the following metrics will be created:
- MessageBroker/Kafka/Nodes/n1
- MessageBroker/Kafka/Nodes/n1/Produce/t1
- MessageBroker/Kafka/Nodes/n1/Produce/t2
- MessageBroker/Kafka/Nodes/n2
- MessageBroker/Kafka/Nodes/n2/Produce/t1
- MessageBroker/Kafka/Nodes/n2/Produce/t2

## Current implementation
For each Kafka client, a `NewRelicMetricsReporter` is instantiated. On its constructor, it creates a map with one `NodeMetricNames` (value) for each node (key) a client connects to. (The key is never used anywhere in the code. All references for this map have a `.value()` call after it).
When `NodeMetricNames` is instantiated, it creates a metric with format 1 for its node.
When a metric is registered on `NewRelicMetricsReporter`, if it has a topic, iterate thru the `NodeMetricNames` registering that topic.
In `NodeMetricNames` if the topic was not seen before, create a metric name (and an event attribute) with the format 2 and mark the topic as seen.

So for each topic, `<node count>` checks are made to create the metric with format 2, also `<node count>` sets are kept to keep the seen topics. Since a new topic will be added to all nodes, there is no reason to make the check for each node, nor keep a set of topics for each node.

Another difference of the new implementation is that it only keeps metric names or event attribute labels instead of both. Since for an agent run only one of either will ever be used, this will save a little memory.

## New implementation
A new class `NodeTopicRegistry` was created. It's functionality is similar to `NodeMetricNames`, but the newer keeps track of all nodes and topics, so only one instance of it is created for each `NewRelicMetricsReporter`.
When `NodeTopicRegistry` is instantiated, it receives all the nodes the Kafka client connects to, and creates all the metric names of format 1.
When a metric is registered with the `NewRelicMetricsReporter`, it will be passed to the `NodeTopicRegistry` and if that topic was not seen before, new metrics of format 2 will be created with the new topic for each node.

So only one check for each topic is done, as well as only one collection of seen topics is kept.

## Related Github Issue
This change was made while adding the same functionality for the `kafka-clients-node-metrics-x.y` (#2186)

### DIFF
--- a/instrumentation/kafka-clients-metrics-3.0.0/src/main/java/com/nr/instrumentation/kafka/NodeTopicRegistry.java
+++ b/instrumentation/kafka-clients-metrics-3.0.0/src/main/java/com/nr/instrumentation/kafka/NodeTopicRegistry.java
@@ -1,0 +1,92 @@
+/*
+ *
+ *  * Copyright 2025 New Relic Corporation. All rights reserved.
+ *  * SPDX-License-Identifier: Apache-2.0
+ *
+ */
+
+package com.nr.instrumentation.kafka;
+
+import org.apache.kafka.common.Node;
+import org.apache.kafka.common.metrics.KafkaMetric;
+
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+import static com.nr.instrumentation.kafka.MetricsConstants.NODE_PREFIX;
+
+/**
+ * This class is used to track all the metric names that are related to a specific node:
+ *
+ * - MessageBroker/Kafka/Nodes/host:port
+ * - MessageBroker/Kafka/Nodes/host:port/Consume/topicName
+ * - MessageBroker/Kafka/Nodes/host:port/Produce/topicName
+ *
+ * At initialization time we only have the node and the mode (is this a metrics reporter
+ * for a Kafka consumer or for a Kafka producer?).
+ *
+ * Then, as topics are discovered through the metricChange method, the topic metric names are
+ * generated. This is the best way we have to get track of the topics since they're not
+ * available when the KafkaConsumer/KafkaProducer is initialized.
+ *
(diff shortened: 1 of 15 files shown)


instrumentation/kafka-clients-metrics-3.0.0/src/main/java/com/nr/instrumentation/kafka/NodeTopicRegistry.java
```diff
@@ -0,0 +1,92 @@
...
+ * Alternatively we could get the topics from the records in KafkaProducer.doSend or
+ * KafkaConsumer.poll, and call NewRelicMetricsReporter.addTopicToNodeMetrics from there.
+ * This approach would have a small impact in performance, and getting the topics from the
+ * KafkaMetrics is a good enough solution.
+ */
+public class NodeTopicRegistry {
+    private final Set<String> recordedTopics = ConcurrentHashMap.newKeySet();
+    // contains the registered metric or event names, according to config
+    private final Set<String> convertedNames = ConcurrentHashMap.newKeySet();
+    private final Set<String> nodes = new HashSet<>();
```
meiao, line 49: This field is not altered after the constructor, so not bothering to add concurrency control to it.

jasonjkeller (approved):
Looks good!

PR #2199, opened 2025-01-13
Instrumentation module for Struts2 v6.7.0

### LINKED ISSUE #2158: Implement Struts 2.0 module that supports struts-core-6.7.0+
### Description
The current struts 2.0 module fails verification against `struts-core-6.7.0`:

`WeaveViolation{type=METHOD_INDIRECT_INTERFACE_WEAVE, clazz=com/opensymphony/xwork2/ActionProxy, method=execute()Ljava/lang/String;}`

jtduffy:
### Overview
New module to support Struts2 v6.7.0 and greater. Mainly package name refactoring in Struts2.

### Related Github Issue
Resolves #2158


settings.gradle
```diff
@@ -108,6 +108,7 @@ include 'instrumentation:apache-pekko-http-2.13_1'
 include 'instrumentation:apache-pekko-http-3_1'
 include 'instrumentation:apache-pekko-http-core-2.13_1'
 include 'instrumentation:apache-struts-2.0'
+include 'instrumentation:apache-struts-2.0_6.7.0'
```
meiao, line 111: any reason to not name this module `apache-struts-6.7.0`?

jtduffy: Only because we named the original module `apache-struts-2.0` (which was my mistake).

I'm actually fine renaming this module and the original struts2 module if everyone else is ok with it.

Issue #2036, opened 2024-09-03
Out of the box DT for SQS


kford-newrelic:
## Is your feature request related to a problem? Please describe.
Would it be possible for the Java Agent to provide support for Distributed Tracing for SQS out of the box?
Customer has described 2 use cases:

1. Simple SQS – I’ve seen a few of our services enqueue message in SQS because of an incoming source HTTP request.  Today, when we pull down the SQS message, it’s not linked to the source HTTP request, which could be helpful.
2. SNS to SQS Subscriptions – A for more common pattern, at least specific to my team, is that we have async services that are interconnected to each other via SNS topics that broadcast out to a set of SQS queues.  Because my team uses this pattern so much, I’m manually writing logic to inject the NR transaction headers into the SNS topic and then reinsert them into the destination service that continues processing the request.

## Feature Description
Distributed Tracing supported across SQS & SNS as described above.

Note : During initial spiking, we should consider other potential scenarios above/beyond those above.

## Describe Alternatives
N/A

## Additional context
N/A

## Priority
Please help us better understand this feature request by choosing a priority from the following options: 
[Really Want]

kford-newrelic:
There may be an agent extension available from the NR Labs team that can serve as a starting point

elucus:
NR Labs version: https://github.com/newrelic-experimental/newrelic-java-aws-sqs

obenkenobi:
It is possible for us to set up distributed tracing for producers where we add new message attributes. For consumers on the other hand it is not possible due to how the SDK works. Whenever we receive messages, it is in bulk. If the `recieveMessage` gives us 1 message from one app but another from a different app, we can only pick one to accept as the parent app. This is fundamental to how distributed tracing works.

obenkenobi:
This is a similar issue faced with our Kafka instrumentation. So the solution to only add out of the box DT for producers replicates an existing pattern.

obenkenobi:
In contrast, NR Lab's approach adds DT for consumers, however it accepts every message's inbound headers and only assumes one entity is producing to a queue. 

However, since you receive more than one message using the sdk, if one or more message comes from two different New Relic entities, only one parent entity will be linked leading to the other(s) not to be linked for a certain trace. In that scenario, manual instrumentation is preferred.

obenkenobi:
Another update, our decision is to support out of the box DT for producers but not for consumers.

PR #2204, opened 2025-01-15
Sqs distributed trace

obenkenobi:
### Overview
Adds distributed trace headers as message attributes for SQS send operations.

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/2036


obenkenobi:
Class loading errors in instrumentation tests forced me to use inner classes for DT Header classes

instrumentation/aws-java-sdk-sqs-1.10.44/build.gradle
```diff
@@ -8,7 +8,8 @@ dependencies {
 
 
 jar {
-    manifest { attributes 'Implementation-Title': 'com.newrelic.instrumentation.aws-java-sdk-sqs-1.10.44' }
+    manifest { attributes 'Implementation-Title': 'com.newrelic.instrumentation.aws-java-sdk-sqs-spans-1.10.44', 'Enabled': 'false',
```
meiao, line 11: instead of disabling the module by default, why not add a config to enable adding the headers?

instrumentation/aws-java-sdk-sqs-1.10.44/src/main/java/com/amazonaws/services/sqs/AmazonSQS_Instrumentation.java
```diff
@@ -7,42 +7,191 @@
...
+                    if(dataType.equalsIgnoreCase("String")) {
+                        String stringValue = value.getStringValue();
+                        if(stringValue != null) return stringValue;
+                    }
+                }
+            }
+            return null;
+        }
+        @Override
+        public Collection<String> getHeaders(String name) {
```
meiao, line 161: Same thing about `Collections` here.

instrumentation/aws-java-sdk-sqs-2.1.0/src/main/java/software/amazon/awssdk/services/sqs/DefaultSqsClient_Instrumentation.java
```diff
@@ -7,43 +7,259 @@
...
         NewRelic.getAgent().getTracedMethod().reportAsExternal(messageConsumeParameters);
+
         return Weaver.callOriginal();
     }
+
+    /*
+     * Headers are used as inner classes to avoid class loading errors when running instrumentation tests
+     * */
+
+    public class SQSRequestHeaders implements Headers {
```
meiao, line 84: Move this inner class to its own file and use the same on the async client.

meiao (changes requested):

instrumentation/aws-java-sdk-sqs-2.1.0/src/main/java/software/amazon/awssdk/services/sqs/DefaultSqsAsyncClient_Instrumentation.java
```diff
@@ -74,8 +103,195 @@ public CompletableFuture<ReceiveMessageResponse> receiveMessage(ReceiveMessageRe
...
+        }
+
+        @Override
+        public String getHeader(String name) {
+            Map<String, MessageAttributeValue> messageAttributes = request.messageAttributes();
+            if(messageAttributes != null) {
+                MessageAttributeValue value = messageAttributes.get(name);
+                if (value != null && value.dataType().equalsIgnoreCase("string")) {
+                    String valueString = value.stringValue();
+                    if (valueString != null) {
```
meiao, line 136: This if check is redundant. If valueString is null, the method will return null anyway.

instrumentation/aws-java-sdk-sqs-2.1.0/src/main/java/software/amazon/awssdk/services/sqs/DefaultSqsAsyncClient_Instrumentation.java
```diff
@@ -74,8 +103,195 @@ public CompletableFuture<ReceiveMessageResponse> receiveMessage(ReceiveMessageRe
...
+        @Override
+        public String getHeader(String name) {
+            Map<String, MessageAttributeValue> attributes = requestEntry.messageAttributes();
+            if(attributes != null) {
+                MessageAttributeValue value = attributes.get(name);
+                if(value != null) {
+                    String dataType = value.dataType();
+                    if(dataType.equalsIgnoreCase("String")) {
+                        String stringValue = value.stringValue();
+                        if(stringValue != null) return stringValue;
```
meiao, line 225: Same redundant null check.

instrumentation/aws-java-sdk-sqs-1.10.44/build.gradle
obenkenobi: This is the existing SQS instrumentation. Disabling would break customer dashboards

instrumentation/aws-java-sdk-sqs-2.1.0/src/main/java/software/amazon/awssdk/services/sqs/DefaultSqsClient_Instrumentation.java
obenkenobi: Problem here is when I did that, I got class-loading errors when running instrumentation tests. I suspect its some magic the SDK is using with classloaders that did not mesh well with the agent. This was the one way I got it to work.

instrumentation/aws-java-sdk-sqs-1.10.44/build.gradle
meiao: Indeed, that is why I am recommending not disabling it by default, which your PR does.

instrumentation/aws-java-sdk-sqs-2.1.0/src/main/java/software/amazon/awssdk/services/sqs/DefaultSqsClient_Instrumentation.java
meiao: You need to add the package that has the header classes to the @InstrumentationTestConfig
In this case it would look like:
```
@InstrumentationTestConfig(includePrefixes = { "software.amazon.awssdk.services.sqs", "com.newrelic.utils" }, configName = "dt_enabled.yml")
```

instrumentation/aws-java-sdk-sqs-1.10.44/build.gradle
meiao: Also, I hadn't noticed, but you added spans to the module name.

instrumentation/aws-java-sdk-sqs-2.1.0/src/main/java/software/amazon/awssdk/services/sqs/DefaultSqsAsyncClient_Instrumentation.java
obenkenobi: Removed the null check

instrumentation/aws-java-sdk-sqs-2.1.0/src/main/java/software/amazon/awssdk/services/sqs/DefaultSqsAsyncClient_Instrumentation.java
obenkenobi: Removed the null check

instrumentation/aws-java-sdk-sqs-2.1.0/src/main/java/software/amazon/awssdk/services/sqs/DefaultSqsClient_Instrumentation.java
obenkenobi: Moved the header classes

obenkenobi: Thanks for the suggestion. I moved the header classes and the tests pass now.

meiao (approved):

PR #2228, opened 2025-01-30
invokeSuspend edge case fix

### LINKED ISSUE #2201: Handle invokeSuspend edge case in the Weaver
In an effort to support Kotlin Coroutines instrumentation, modify the weaver to support the edge case presented by Kotlin bytecode for `invokeSuspend`. 

The issue was investigated as a spike in #2002 , where errors were thrown during weaving of `invokeSuspend`. The investigation found that the Kotlin compiler intentionally duplicates operands on the stack prior to return instructions in the body of `invokeSuspend`. Though innocuous for `return` instructions, these extra operands cause bytecode verification to fail when `return` is replaced with `goto` (which is how the agent performs its bytecode modification). 

We should update the weaver to handle this edge case, so that weaving for `invokeSuspend` succeeds. This fix should be wrapped in a feature flag/some kind of blocker to prevent the fix from applying to any other methods at this time.

kanderson250:
Resolves #2201 

## Description

This PR addresses a bug we found when attempting to weave `invokeSuspend` in Kotlin Coroutines. 

The weaver replaces RETURN instructions with GOTO instructions. This replacement can be unsafe if the operand stack prior to the RETURN instruction is not tidied up (ie, has more operands than needed to perform the return). This issue was previously unknown to us, because most compilers already clear their return stacks. However, the Kotlin compiler intentionally adds extra operands to Coroutines methods (for debugging purposes) and doesn't clean them up. 

The purpose of this PR is to check the offending method (`invokeSuspend`) and tidy up its return stacks before the return instructions are replaced with gotos.  

## Structure

Here is how a method will be processed with this fix:

1. The method is intercepted in `getInliner` of `MethodCallInlinerAdapter`. At this point, the method's bytecode has not been modified. 
2. If the method’s name is `invokeSuspend`, it is visited with a `ClearReturnAdapter`.
3. The adapter calls `ReturnInsnProcessor.clearReturnStacks()`, which :
    -  Computes the operand stacks for the entire method, locates the return instructions, and stores their stack sizes
    -  Adds a STORE - n-many POPs - LOAD sequence to the instructions list, where n is the number of pops needed to tidy the stack. To perform the STORE and LOAD, it creates a new local variable slot. 
4. Returns the method to the normal weaver execution flow. 

## Issues

One issue that I couldn't figure out how to address is how to handle failures of the frame analyzer (will leave a comment on the code). Unfortunately, the weaver can't log anything, and propagating an error upwards is also probably not feasible. That means it is currently a silent failure. 

Another issue we should take seriously whenever bytecode is being modified is thread safety. I have left several comments in the code about this. 

## Testing

New tests can be found in` ReturnInsnProcessorTest` and `WeaveCoroutineTest`. A test dependency on coroutines has been added to the weaver, because coroutines are the only known way to get the compiler to make the weird bytecode. 

## Other

Some utility methods are added to `WeaveUtils`, which I developed during the debugging of this fix and think would be helpful to keep for the future. 

## Feature Flag

I added a non-public system property to disable this feature:

`-Dnewrelic.config.class_transformer.clear_return_stacks=false`

Setting this property to false will allow the AIOOB error through as before.

### DIFF
--- a/newrelic-weaver/src/main/java/com/newrelic/weave/MethodCallInlinerAdapter.java
+++ b/newrelic-weaver/src/main/java/com/newrelic/weave/MethodCallInlinerAdapter.java
@@ -19,12 +19,7 @@ import org.objectweb.asm.commons.LocalVariablesSorter;
 import org.objectweb.asm.commons.MethodRemapper;
 import org.objectweb.asm.commons.Remapper;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 public abstract class MethodCallInlinerAdapter extends LocalVariablesSorter {
     /**
@@ -113,7 +108,6 @@ public abstract class MethodCallInlinerAdapter extends LocalVariablesSorter {
             }
             int access = opcode == Opcodes.INVOKESTATIC ? Opcodes.ACC_STATIC : 0;
             inliner.inliner = new InliningAdapter(api, access, desc, this, mv, inliner.remapper);
-
         }
         inliner.method.accept(inliner.inliner);
     }
@@ -133,8 +127,10 @@ public abstract class MethodCallInlinerAdapter extends LocalVariablesSorter {
             } else {
                 // Copy the MethodNode before modifying the instructions list (which is not thread safe)
                 MethodNode methodNodeCopy = WeaveUtils.copy(method.method);
-                if (name.equals("invokeSuspend")) {
-                    methodNodeCopy = ReturnInsnProcessor.clearReturnStacks(owner, methodNodeCopy);
+                if (shouldClearReturnStacks(owner, name, desc)){
+                    MethodNode result = WeaveUtils.newMethodNode(methodNodeCopy);
+                    methodNodeCopy.accept(new ClearReturnAdapter(owner, methodNodeCopy, result));
+                    methodNodeCopy = result;
                 }
                 methodNodeCopy.instructions.resetLabels();
                 InliningAdapter originalInliner = method.inliner;
@@ -146,6 +142,32 @@ public abstract class MethodCallInlinerAdapter extends LocalVariablesSorter {
             inliners.put(key, method);
         }
         return method;
+    }
+
+    /*
+    Only the invokeSuspend method of kotlin coroutines has been known to leave return stacks untidy.
+     */
+    private boolean shouldClearReturnStacks(String owner, String name, String desc) {
+        final String invokeSuspendOwner = WeaveUtils.INLINER_PREFIX + "kotlin/coroutines/jvm/internal/BaseContinuationImpl";
+        final String invokeSuspendName = "invokeSuspend";
+        final String invokeSuspendDesc = "(Ljava/lang/Object;)Ljava/lang/Object;";
+        return invokeSuspendOwner.equals(owner) && invokeSuspendName.equals(name) && invokeSuspendDesc.equals(desc);
+    }
+
+    class ClearReturnAdapter extends MethodNode {
+        String owner;
+        public ClearReturnAdapter(String owner, MethodNode source, MethodVisitor next) {
+            super(WeaveUtils.ASM_API_LEVEL, source.access, source.name, source.desc, source.signature, source.exceptions.toArray(new String[source.exceptions.size()]));
+            this.mv = next;
+            this.owner = owner;
+        }
+
+        @Override
+        public void visitEnd() {
+            System.out.println("In visitEnd of OtherClearReturnAdapter for owner: " + owner + " and name: " + name + " and desc " + desc);
+            ReturnInsnProcessor.clearReturnStacks(owner, this);
+            accept(mv);
+        }
     }
 
     class InliningAdapter extends LocalVariablesSorter {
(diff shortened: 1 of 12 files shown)


newrelic-weaver/src/main/java/com/newrelic/weave/utils/ReturnInsnProcessor.java
```diff
@@ -0,0 +1,103 @@
...
+     *
+     * @param owner The owning class of the method
+     * @param mn MethodNode representing the method to be modified
+     */
+    public static void clearReturnStacks(String owner, MethodNode mn) {
+        Map<AbstractInsnNode, Integer> stacks;
+        try {
+            stacks = getReturnStacks(owner, mn);
+        } catch (AnalyzerException e) {
+            //Something went wrong calculating return stacks, leave the method unmodified.
```
kanderson250, line 26: This is a silent failure

newrelic-weaver/src/main/java/com/newrelic/weave/MethodCallInlinerAdapter.java
```diff
@@ -7,7 +7,9 @@
 
 package com.newrelic.weave;
 
+import com.newrelic.weave.utils.ReturnInsnProcessor;
 import com.newrelic.weave.utils.WeaveUtils;
+import org.objectweb.asm.tree.*;
```
jasonjkeller, line 12: FYI: If you setup Intellij to use the [java-agent-code-style.xml](https://github.com/newrelic/newrelic-java-agent?tab=readme-ov-file#intellij-idea-setup) you can have it clean up all of the important statements for you. We adhere to the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html#s3.3-import-statements) which   prohibits wildcard imports.

newrelic-weaver/src/main/java/com/newrelic/weave/utils/ReturnInsnProcessor.java
```diff
@@ -0,0 +1,103 @@
+package com.newrelic.weave.utils;
```
jasonjkeller, line 1: Add copyright header

newrelic-weaver/src/test/java/com/newrelic/weave/WeaveCoroutineTest.java
```diff
@@ -0,0 +1,120 @@
+package com.newrelic.weave;
```
jasonjkeller, line 1: Add copyright header

newrelic-weaver/src/test/java/com/newrelic/weave/utils/ReturnInsnProcessorTest.java
```diff
@@ -0,0 +1,116 @@
+package com.newrelic.weave.utils;
```
jasonjkeller, line 1: Add copyright header

newrelic-weaver/src/test/java/com/newrelic/weave/weavepackage/testclasses/SampleCoroutine.kt
```diff
@@ -0,0 +1,42 @@
+package com.newrelic.weave.weavepackage.testclasses
```
jasonjkeller, line 1: Add copyright header

newrelic-weaver/src/test/java/com/newrelic/weave/weavepackage/testclasses/Weave_SampleCoroutine.java
```diff
@@ -0,0 +1,12 @@
+package com.newrelic.weave.weavepackage.testclasses;
```
jasonjkeller, line 1: Add copyright header

newrelic-weaver/src/main/java/com/newrelic/weave/MethodCallInlinerAdapter.java
```diff
@@ -234,6 +239,57 @@ protected int newLocalMapping(final Type type) {
...
+        final String invokeSuspendName = "invokeSuspend";
+        final String invokeSuspendDesc = "(Ljava/lang/Object;)Ljava/lang/Object;";
+        return invokeSuspendName.equals(name) && invokeSuspendDesc.equals(desc);
+    }
+
+    /**
+     * Feature flag to disable return stack processing.
+     */
+    private boolean clearReturnStacksDisabled () {
+        String featFlag = System.getProperty("newrelic.config.class_transformer.clear_return_stacks");
```
jasonjkeller, line 260:
```suggestion
        String featFlag = System.getProperty("newrelic.config.class_transformer.clear_return_stacks", false);
```

I believe you can call `getProperty` with a default value of `false` which will be returned if the system property doesn't exist, rather than having it potentially return `null`. 

I also think that `getProperty` can potentially throw an exception so you should make sure to handle that.

newrelic-weaver/src/main/java/com/newrelic/weave/utils/ReturnInsnProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+     * @param mn MethodNode representing the method to be modified
+     */
+    public static void clearReturnStacks(String owner, MethodNode mn) {
+        Map<AbstractInsnNode, Integer> stacks;
+        try {
+            stacks = getReturnStacks(owner, mn);
+        } catch (AnalyzerException e) {
+            //Something went wrong calculating return stacks, leave the method unmodified.
+            //This is a silent failure that indicates that the source (client) bytecode is invalid.
+            //This should never happen!
```
jasonjkeller, line 43: Famous last words! ha

newrelic-weaver/src/main/java/com/newrelic/weave/MethodCallInlinerAdapter.java
kanderson250: Refactored to have getProperty return a default value of `true` (the feature flag is an off switch with a value of `false`). And hopefully the flag checker is slightly more readable now. 

Leaving the unchecked exceptions for getProperty unhandled, to follow the pattern in the rest of the agent.

jasonjkeller (approved):

Issue #2037, opened 2024-09-03
Java Agent FY25Q4 - Support Adding Labels on APM Logs


kford-newrelic:
## Is your feature request related to a problem? Please describe.
Customer switched from using Logstash to forward logs to New Relic, to using our Java agent to do the same thing.  

## Feature Description
In their Logstash config, they were able to tag on custom attributes to their log events, using the following format:
```
input {
  file {
    path => "/var/log/supervisor/java-*.log"
    add_field => {
      custom_field => "custom_value"
    }
start_position => "beginning"

codec => json {
    }
  }
}
```
They would like to see us implement a similar feature in the Java agent config, not unlike our current Labels solution but specific to Log events.  A simple list of key / value pairs that would be attached to every log event sent up by that Java agent.

## Describe Alternatives
N/A

## Additional context
Before starting any initial spiking, we should try to engage the requesting customer (through their account team), to see if they'd be willing to guide feature development and later try snapshot agent versions.

Review epic description on [NR-315883](https://new-relic.atlassian.net/browse/NR-315883)

Similar Jiras on this topic - [NR-303269](https://new-relic.atlassian.net/browse/NR-303269), [NR-310852](https://new-relic.atlassian.net/browse/NR-310852) & [NR-312135](https://new-relic.atlassian.net/browse/NR-312135)

## Priority
Please help us better understand this feature request by choosing a priority from the following options: 
[Nice to Have] 

[NR-303269]: https://new-relic.atlassian.net/browse/NR-303269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[NR-315883]: https://new-relic.atlassian.net/browse/NR-315883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

jasonjkeller:
The Java agent supports adding custom attributes to log events via MDC. https://docs.newrelic.com/docs/apm/agents/java-agent/configuration/java-agent-configuration-config-file/#log-context-data

PR #2238, opened 2025-02-14
Support Adding Labels to APM Forwarded Logs

deleonenriqueta:
## Overview
Allows the user to add custom tags (labels) to agent-forwarded logs. 

### Related Issue
Resolves Issue #2037

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/config/ApplicationLoggingLabelsConfig.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/config/ApplicationLoggingLabelsConfig.java
@@ -1,0 +1,59 @@
+package com.newrelic.agent.config;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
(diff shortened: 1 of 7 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/config/ApplicationLoggingLabelsConfig.java
```diff
@@ -0,0 +1,59 @@
...
+        return excludeSet;
+    }
+
+    public boolean getEnabled() { return enabled; }
+
+    public Set<String> getExcludeSet() { return excludeSet; }
+
+    public boolean isExcluded(String label) { return getExcludeSet().contains(label); }
+
+    public Map<String, String> removeExcludedLabels(Map<String, String> labels) {
```
jtduffy, line 50: I might be missing some context, but I think the API could be simplified if the excludes list was applied during the call to `initExcludes` call? (Of course that would assume that you have the labels available to pass into the `ApplicationLoggingLabelsConfig` constructor)

```
ApplicationLoggingLabelsConfig(Map<String, Object> props, String parentRoot, Map<String, String> labels) {
    //stuff
    //call removeExcludedLabels() and make that method privare
}
```

jtduffy (approved):

PR #2251, opened 2025-02-28
Making Pattern.compile calls into class constants

### LINKED ISSUE #2249: Move pattern compile out of GeneratedClassDetector#isGenerated
### Description
GeneratedClassDetector#isGenerated compiles three regexes, which might not even be used in its invocations.

Since the values are always going to be the same, these could be constants in the class so the patterns do not have to be compiled on each invocation.

meiao:
### Overview
Moving the creation of some regexes to constants instead of on method invocation.

Having this be constants, improves the method performance by about 30%. It also decreases the allocation/GC overhead of the agent slightly.

The downside is that it makes the agent hold onto 3 Pattern objects forever.

Note that this method is not called that often. When I opened the ticket I thought it was for every loaded class, but it is only called for loaded classes that do not have sources. 

### Related Github Issue
Fixes #2249

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/instrumentation/context/GeneratedClassDetector.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/instrumentation/context/GeneratedClassDetector.java
@@ -15,13 +15,14 @@ import java.util.regex.Pattern;
 
 public class GeneratedClassDetector implements ClassMatchVisitorFactory {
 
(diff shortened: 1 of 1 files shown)


jtduffy (approved):

meiao:
There are other places where Patterns are compiled on method invocation. SystemPropertyProvider is another example that gets called even more often (~2900 times for Spring Pet Clinic). These two only occur during agent startup and there is very little gain by setting the constants.

Issue #2222, opened 2025-01-29
[Spike] to fix broken traces by adding W3C sampling flag support


kmudduluru:
see the [feature request](https://new-relic.atlassian.net/browse/NR-347703) for more details. This spike is to determine the fix and fix as needed.

PR #2279, opened 2025-03-19
Added sampling options when an inbound traceparent exists

jbedell-newrelic:
New config options for how to sample when we have sampling data from an upstream entity.

Possible values:
  default: New Relic's standard sampling rules will be used.
  always_on: Always sample spans in this case.
  always_off: Always skip sampling in this case.

YML config
```
distributed_tracing:
    sampler:
      remote_parent_sampled: default
      remote_parent_not_sampled: default
```

So, setting `remote_parent_sampled` to `always_on` would tell the agent to try and always sample anything that an upstream entity told us was sampled, until the sampling reservoir is full.

Or, setting `remote_parent_not_sampled` to `always_off` would tell the agent to not bother sampling anything if the upstream entity told us that it was not sampled.


jtduffy (approved):

jbedell-newrelic:
Resolves https://github.com/newrelic/newrelic-java-agent/issues/2222

PR #2219, opened 2025-01-24
Micronaut core instrumentatin modules

jtduffy:
### Overview
This is a port of the NR labs Micronaut core instrumentation modules, including core-reactive.

### Related Github Issue
Resolves #2200

PR #2280, opened 2025-03-21
Micronaut http instrumentation modules

gsidhwani-nr:
Overview
This is a port of the NR labs Micronaut http instrumentation modules.

Please review and share feedback .

The fork is also carrying the Micronaut Core module added by @jtduffy 
https://github.com/newrelic/newrelic-java-agent/pull/2219


jtduffy (changes requested):
Everything looks good except maybe the transaction naming. I would double check each call to `setTransactionName` so we can standardized on the names across the agent.

gsidhwani-nr:
fix: updated transaction name format

Issue #2273, opened 2025-03-17
Improve Kafka consumer spans processing


kmudduluru:
review this issue for more details.
https://new-relic.atlassian.net/browse/NR-378724

PR #2283, opened 2025-03-26
Add kafka consumer distributed traces

obenkenobi:
### Overview
Add a new Kafka instrumentation module named `kafka-clients-spans-consumer-0.11.0.0` that accepts distributed trace headers for consumers. 

When a customer enabled the module using the config
```yml
common: &default_settings
...
  class_transformer:
  ...
    kafka-clients-spans-consumer:
      enabled: true
...
```
Transaction's that contain the app calling Kafka-client's consumer `poll(...)` method will accept distributed trace headers from one Kafka record. 

This is done because in distributed tracing, transactions can only have one parent trace.  

As a result, you will have distributed traces from the producer that either have the consumer traces or none at all.

Due to these limitations, this instrumentation module is disabled by default

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/2273


instrumentation/kafka-clients-spans-consumer-0.11.0.0/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer_Instrumentation.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public ConsumerRecords<K, V> poll(final long timeoutMs) {
+        final ConsumerRecords<K, V> records = Weaver.callOriginal();
+        nrAcceptDtHeaders(records);
+        return records;
+    }
+
+    private void nrAcceptDtHeaders(ConsumerRecords<K, V> records) {
+        AgentBridge.getAgent().getLogger().log(Level.INFO, "nrAcceptDtHeaders");
```
jtduffy, line 30: I assume these INFO logs can be removed or changed to FINEST

jtduffy (changes requested):
Also missing copyright headers in the java source files

instrumentation/kafka-clients-spans-consumer-0.11.0.0/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer_Instrumentation.java
obenkenobi: @jtduffy I removed the logs

jtduffy:
@obenkenobi  I think you missed my other comment about missing copyright headers in the java source files

obenkenobi:
@jtduffy Thanks for letting me know, I just added the headers

jtduffy (approved):

PR #2298, opened 2025-04-03
fixed gaps in Micronaut Http instrumentation

dhilpipre:
Found gaps in instrumentation for Micronaut Http

- Fixed missing / in MicronautSubresourceInstrumentation.java where the path was missing to separate the pieces of the path
- Added additional code to methods to name the transaction to give a more meaningful transaction name instead of the default NettyDispatcher


dhilpipre:
Tabs have been changed to spaces as requested

jtduffy (approved):

PR #2307, opened 2025-04-10
Broaden suspends fix

### LINKED ISSUE #2285: Broaden kotlin suspends fix to other methods
Following the release of our fix of an `ArrayIndexOutOfBoundsException` triggered by the Lab Team's instrumentation of Kotlin `invokeSuspend`, there were reports that the exception was popping up in other places, too. 

This is a repro, where the AIOOBE can be seen occurring elsewhere: https://github.com/MT-Jacobs/sample-newrelic-kotlin-failure

Our fix for `invokeSuspend` is narrowly scoped to apply *only* to this method (invokeSuspend). I have already verified against the repro that applying the fix to *all* methods does resolve the AIOOBE exception. 

However, we need to evaluate and implement the best way of opening the scope of this fix. We do not want to apply the fix in non-Kotlin contexts. We would like to avoid applying the fix where we do not have to, as the additional analysis of stack frames may cause some overhead. 

Investigate the source of the new AIOOBE in the repro; attempt to determine all possible Kotlin classes that may cause this error; and implement a workaround. Consider adding configuration to manually specify classes the fix should apply to.

kanderson250:
Resolves #2285 

This PR refactors our check for applying additional return insn processing to depend completely on the feature flag `-Dnewrelic.config.class_transformer.clear_return_stacks`. When this flag is set to `true`, the kotlin weaver fix will apply to **all methods**. When the flag is set to `false` (default/when the flag is unset), the kotlin weave fix will apply to **no methods**. A new warning message will appear in the agent logs instructing users to turn the flag on if the class transformer encounters the AIOOBE. 

This refactor was made because there isn't a consistent method signature that the Weaver can recognize as having the Kotlin problem; this is only known at the bytecode insn level, once the frames have been analyzed.

### DIFF
--- a/newrelic-weaver/src/main/java/com/newrelic/weave/MethodCallInlinerAdapter.java
+++ b/newrelic-weaver/src/main/java/com/newrelic/weave/MethodCallInlinerAdapter.java
@@ -253,15 +253,6 @@ public abstract class MethodCallInlinerAdapter extends LocalVariablesSorter {
         return enabled.equalsIgnoreCase("true");
     }
 
-//    /**
-//     * Feature flag to disable return stack processing.
-//     * To use this feature flag, set -Dnewrelic.config.class_transformer.clear_return_stacks=false at JVM startup.
-//     */
-//    private boolean clearReturnStacksDisabled() {
-//        String enabled = System.getProperty("newrelic.config.class_transformer.clear_return_stacks", "true");
-//        return enabled.equalsIgnoreCase("false");
-//    }
-
     /**
      * This adapter checks a method's return instructions, adding additional POPs prior to return instructions
      * if the return is made with extra (>1) operands on the stack.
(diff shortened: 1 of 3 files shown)


newrelic-weaver/src/main/java/com/newrelic/weave/MethodCallInlinerAdapter.java
```diff
@@ -240,26 +240,17 @@ protected int newLocalMapping(final Type type) {
...
-
-    /**
-     * Feature flag to disable return stack processing.
-     * To use this feature flag, set -Dnewrelic.config.class_transformer.clear_return_stacks=false at JVM startup.
-     */
-    private boolean clearReturnStacksDisabled() {
-        String enabled = System.getProperty("newrelic.config.class_transformer.clear_return_stacks", "true");
-        return enabled.equalsIgnoreCase("false");
+        String enabled = System.getProperty("newrelic.config.class_transformer.clear_return_stacks", "false");
+        return enabled.equalsIgnoreCase("true");
```
jtduffy, line 253: This is perfectly fine, but you can also use `Boolean.getBoolean()` too.
https://docs.oracle.com/javase/8/docs/api/java/lang/Boolean.html#getBoolean-java.lang.String-

jtduffy (approved):

newrelic-weaver/src/main/java/com/newrelic/weave/MethodCallInlinerAdapter.java
kanderson250: yeah, let's do that

kanderson250: Took your suggestion and inlined the method altogether since a whole method call is overkill now

jtduffy (approved):

Issue #852, opened 2022-05-16
Add instrumentation to support spring kafka @KafkaListener annotation


jasonjkeller:
It looks like `spring-kafka` uses this [@KafkaListener](https://docs.spring.io/spring-kafka/docs/current/reference/html/#x28-batch-overrude) in kafka consumers, which is not something that we currently instrument and thus the agent won’t start a transaction which means any external calls made in that process won’t be captured. We should add new instrumentation to support spring kafka.

We should also do some research to figure out how widely used this is versus the default kafka libraries.

(Migrate to Jira)

kford-newrelic:
Not selected for the Jan-Mar quarter.

PR #2312, opened 2025-04-15
Kafka Consumer Distributed Traces and Spring Kafka Instrumentation.

obenkenobi:
### Overview
Adds new instrumentation for Spring-Kafka and distributed tracing for customers using the core Kafka client library.

The instrumentation for Spring-Kafka will create transactions for kafka listeners (i.e. transactions annotated with the @KafkaListener annotation). Distributed tracing is also added by default for listeners that only listen for one message.

For batch processing of Kafka messages via either spring kafka or using Kafka Clients's `consumer.poll(...)` method, distributed tracing is by default turned off.  If customers want to enable it, use the following config:

```
kafka:
  spans:
    distributed_trace: 
      consume_many:
        enabled: true
```

The reason this feature is off by default transactions can only have one parent trace and more than one message may be produced within different traces/transactions. The result here is distributed traces would end up fragmented where we see some producer transactions/traces having child consumer traces while others do not.

### Related Github Issue

https://github.com/newrelic/newrelic-java-agent/issues/852

https://github.com/newrelic/newrelic-java-agent/issues/2273


obenkenobi:
Until there is consensus with other agent teams on how to name kafka transactions with batch calls and after agent 8.20 is released is when I would merge this transaction

jtduffy (approved):

jtduffy (approved):

PR #2316, opened 2025-04-16
Adjust module build settings

jtduffy:
This module now applies to v1.3 --> current. A skip class was added to prevent applying to 1.0 -> 1.2.


dhilpipre (approved):
Everything looks fine

jbedell-newrelic (approved):

PR #2344, opened 2025-05-01
Allow the org.crac JAR to actually be shadowed as requested.

jbedell-newrelic:
The package has to be added to an allow-list for the shadowIntoJar to actually work properly.


vandykej17:
This is also causing issues on our end and would love for this to be merged and released!

jtduffy (approved):

PR #2358, opened 2025-05-09
KafkaConsumerConfig event support for Kafka 3.7+ 

nrlkirby:
### Overview

The existing code to emit KafkaConsumerConfig events, when enabled, has not worked for versions of Kafka 3.7 and onwards. Apache Kafka release versions since then have used a delegation pattern for KafkaConsumer instances, breaking the existing stack-based analysis that the integration applied to confirm that we were only emitting config events attached to actual clients, and not instantiated independently for whatever reason.

### Testing

A separately maintained integration test confirmed the existing agent code failed to emit KafkaConsumerConfig events for versions 3.7.0 onwards, as did manual testing. That integration testing confirmed that the agent, with these fixes, produces the expected KafkaConsumerConfig events. Specific versions of kafka-clients tested: 2.6.2, 2.8.1, 3.3.2, 3.4.1, 3.5.0, 3.6.2, 3.7.0, 3.8.0, 3.9.0, 4.0.0. The 3.9.0 and 4.0.0 releases were also tested with and without the consumers configured to use the new async/consumer group protocol that engages a different consumer delegate.

### Checks

- [X] Your contributions are backwards compatible with relevant frameworks and APIs.
- [X] Your code does not contain any breaking changes. Otherwise please describe. 
- [X] Your code does not introduce any new dependencies. Otherwise please describe.

### DIFF
--- a/instrumentation/kafka-clients-config-1.1.0/src/main/java/org/apache/kafka/kafka/common/config/AbstractConfig_Instrumentation.java
+++ b/instrumentation/kafka-clients-config-1.1.0/src/main/java/org/apache/kafka/kafka/common/config/AbstractConfig_Instrumentation.java
@@ -5,9 +5,6 @@ import com.newrelic.api.agent.weaver.Weave;
 import com.newrelic.api.agent.weaver.Weaver;
 import com.nr.instrumentation.kafka.config.ConfigEmitter;
 import org.apache.kafka.clients.CommonClientConfigs;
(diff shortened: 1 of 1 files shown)


jtduffy (approved):

jtduffy:
All tests passed in https://github.com/newrelic/newrelic-java-agent/pull/2359
Manual tests passed

PR #2454, opened 2025-08-13
Kotlin coroutines

dhilpipre:
Provides instrumentation for Kotlin Coroutines from version 1.4 onwards.


dhilpipre:
@jtduffy Hi Jerry, 
I will look into to it and let you know.
I do want to make note of why the configuration is "hot" deploy and you can let me know what you think.  
The reason for the configuration is that we can get some of each of the ones configured can be very fast and frequently invoked and tracking them can impact performance.  Plus there can also be ones that the customer is not interested in tracking.  With hot deploy, they can tune the configuration on the fly when they see something they want to eliminate from tracking.
Without hot deploy they would need to restart in order for the change to take effect.  
If we eliminate hot deploy then we should add verbiage in the documentation that they should tune it in a non-production environment before deploying to production.

PR #2461, opened 2025-08-19
OTel attribute mapper impl

### LINKED ISSUE #2436: Traces - Map Attributes From Different Versions of OTel Specs
We need to map attributes such that they adhere to different otel spec versions.

In [ExitTracerSpan](https://github.com/newrelic/newrelic-java-agent/blob/01ccb3e8ed0f239b64d3dda1906445e350c5ac29/instrumentation/opentelemetry-sdk-extension-autoconfigure-1.28.0/src/main/java/io/opentelemetry/sdk/trace/ExitTracerSpan.java#L52) we map attributes found on the OTel Spans into different attributes that we want on NR Spans.

We need to implement mapping rules to adhere to several versions of the OTel spec. For example, currently we'll map `server.address` from an OTel span to `host` on a NR Span, but previous versions of the OTel spec would have an attribute named `net.host.name` that should be mapped to `host`. We don't currently do this, but we should. See the examples here from the [Node agent mappings](https://github.com/newrelic/node-newrelic/blob/main/lib/otel/transformation-rules.json#L2-L121) for various OTel spec versions.

jtduffy:
### Overview
Resolves #2436 

This change allows OTel attributes to be mapped across the various OTel semantic conventions.

#### Approach
Categorize the various attributes via:
- SpanKind
    - Attribute Type (For example "Port", "Host", etc)
        - Actual attribute keys with it's correseponding semantic convention version

For example:
```
SERVER
  Port
    server.port - HTTP Server v1.23
    net.host.port - HTTP Server v1.20
  Host
    server.address - HTTP Server v1.23
    net.host.name - HTTP Server v1.20
```

These mappings will be stored in a JSON file, which will be read and parsed one time, when the `AttributeMapper` singleton is accessed. Note that the JSON files are incomplete at this point until we make sure this is a valid approach.

A method on the mapper, `findProperOtelKey` is used to locate the key being used by the OTel SDK so that the correct value can be retrieved.

`public String findProperOtelKey(SpanKind spanKind, AttributeType type, Set<String> otelKeys)`

Which can be used as follows:
```
Object host = attributes.get(AttributeMapper.getInstance()
    .findProperOtelKey(SpanKind.SERVER, AttributeType.Host, otelAttributes));
// Will look for "server.port" and "new.host.name" and return the one that's present
```

Note that the semantic version value isn't currently used. It was added in case it's needed in the future.

### DIFF
--- a/instrumentation/opentelemetry-sdk-extension-autoconfigure-1.28.0/build.gradle
+++ b/instrumentation/opentelemetry-sdk-extension-autoconfigure-1.28.0/build.gradle
@@ -1,6 +1,13 @@
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
+
 plugins {
     id("java")
 }
+
+configurations {
+    shadowIntoJar
+}
+configurations.implementation.extendsFrom(configurations.shadowIntoJar)
 
 dependencies {
     implementation(project(":agent-bridge"))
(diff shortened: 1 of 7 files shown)


Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-453522

PR #2471, opened 2025-08-28
Updates to support sampling_target.

jbedell-newrelic:
Grabs the adaptive_sampling_target config value(s) and modify them before sending up on connect.  These modifications to the value will be removed later when the harvest cycle period changes.

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/RPMService.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/RPMService.java
@@ -47,6 +47,7 @@ import com.newrelic.agent.transport.HostConnectException;
 import com.newrelic.agent.transport.HttpError;
 import com.newrelic.agent.transport.HttpResponseCode;
 import com.newrelic.agent.utilization.UtilizationData;
+import com.newrelic.api.agent.NewRelic;
 import org.json.simple.JSONStreamAware;
 
 import java.lang.management.ManagementFactory;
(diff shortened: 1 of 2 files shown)


kanderson250 (dismissed):

newrelic-agent/src/main/java/com/newrelic/agent/RPMService.java
```diff
@@ -969,6 +973,61 @@ private void handle503Error(Exception e) {
...
+        // but, the harvest cycle is 12 times per minute right now,
+        // so we need to divide this number by 12 until that changes
+        // there are 2 spots we need to do this:
+        //   - settings.distributed_tracing_sampler_adaptive_sampling_target
+        //   - settings.distributed_tracing.sampler.adaptive_sampling_target
+        if (settings == null) return;
+        try {
+            // local settings
+            Map<String, Object> dtConfig = (Map<String, Object>) settings.get("distributed_tracing");
+            NewRelic.getAgent().getLogger().log(Level.INFO, "JGB dtConfig: "+dtConfig);
```
jtduffy, line 987: Remove JGB logging?

jtduffy (changes requested):

jbedell-newrelic:
> I think this may have introduced an issue. We started seeing this after rolling out 8.24.0. Is this simply log noise? or is this an issue?
> 
> ```
> com.newrelic WARN: Unable to parse local setting adaptive_sampling_target setting: java.lang.NullPointerException: Cannot invoke "String.trim()" because "in" is null
> ```

Hi @mattnelson , this is definitely a bug, but I believe it's a rather innocuous one.  The error should only occur if the sampler stanza is present and the adaptive_sampling_target is not.  In which case the only side effect is that the adaptive_sampling_target value will not be adjusted, which is assumedly not desired anyway, since it was not defined.  Everything else should be functioning properly.  The only other effect is the erroneous log message itself.  

I've opened a [ticket](https://github.com/newrelic/newrelic-java-agent/issues/2488) to get the erroneous log message taken care of.  Thanks for the report!

PR #2513, opened 2025-10-07
Period based adaptive sampler 

### LINKED ISSUE #2485: tracing enhancements
Make some tracing enhancements ahead of the core tracing and Otel trace ratio sampling initiatives.

kanderson250:
Resolves #2485 

This PR decouples our adaptive sampler from the harvest cycle, in an effort to support upcoming initiatives for core tracing and otel style sampling algorithms. 

## What's Changed

Previously, our adaptive sampler was tied to the transaction reservoir (and therefore, the harvest cycle) to perform its math. This PR:

**Adds/Updates**
- Refactors the AdaptiveSampler to be fully self-encapsulated. It runs its own timer, and tracks its own stats, regardless of the state of any event reservoir. To support this self-encapsulated math, a Transaction now needs to assign priority as late in its lifecycle as possible. See the below comment on Transaction changes. 
- Refactors all 3 existing samplers (Adaptive, AlwaysOn, AlwaysOff) to implement a common Sampler interface. 
- Refactors the Transaction class to assign priority as late in its lifecycle as possible. This is required to avoid running the adaptive sampler on a transaction that later accepts DT headers (in which case, the adaptive sampler isn't required to make a decision, and the inbound headers alone should be used). 
  - "As late in the lifecycle as possible" means when createDistributedTracePayload is called, or the transaction finishes. 
  - As before, the Transaction also assigns priority when remote parent sampling decisions are found. 
- Refactors a bit of the remote parent sampling logic to comply with the latest agent specs. Most notably, remote parent sampling decisions may now be taken from the New Relic DT Headers (not just the W3C headers). 
- Refactors the DistributedSamplingPriorityQueue to make use of a custom-implemented synchronized queue that supports a peekLast operation. 

**Removes**
- All the stuff that was required to support our old implementation of the adaptive sampler. This primarily means removing the PriorityAware "decided" field and all references to it. This field used to be used by the reservoir to track adaptive-sampler related statistics (decidedCountLast) - it was **not** used when deciding whether to add an event to the reservoir, but rather to power the math of the adaptive sampler. There are many changes related to this update. 

## Configuration

The AdaptiveSampler can be configured with the `adaptive_sampling_target` setting.

## Relevant Specifications

Adaptive Sampler specification: https://source.datanerd.us/agents/agent-specs/blob/main/distributed_tracing/Distributed-Tracing.md#adaptive-sampler 
Remote Parent Sampled Flag specification: https://source.datanerd.us/agents/agent-specs/blob/main/distributed_tracing/Distributed-Tracing.md#determining-whether-the-remote-parent-was-sampled

## Testing

Unit tests have been added for the adaptive sampler, Transaction class, and DistributedTraceServiceImpl to support these changes. Cross agent tests have been added. Existing tests have been updated.

### DIFF
--- a/agent-interfaces/src/main/java/com/newrelic/agent/interfaces/SamplingPriorityQueue.java
+++ b/agent-interfaces/src/main/java/com/newrelic/agent/interfaces/SamplingPriorityQueue.java
@@ -39,4 +39,6 @@ public interface SamplingPriorityQueue<E extends PriorityAware> {
     int size();
 
     void clear();
+
+    void logReservoirStats();
 }
--- a/newrelic-agent/src/main/java/com/newrelic/agent/Transaction.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/Transaction.java
@@ -358,13 +358,16 @@ public class Transaction {
      *
      * If a priority assignment has already been made, this call is ignored. This is a required check to avoid
      * overwriting any priority decision that was made earlier, either because a remote parent was processed or a previous
-     * call to this method was made earlier in the txn's lifecycle.
+     * call to this method was made earlier in the txn's lifecycle. It is also required to avoid accidentally running the
+     * adaptive sampler twice on the same transaction.
      *
-     * If Distributed Tracing is not enabled, a random priority in [0,1) is assigned.
+     * If Distributed Tracing is not enabled, or this is a synthetic transaction, a random priority in [0,1) is assigned.
      */
     public void assignPriorityRootIfNotSet(){
         if (getAgentConfig().getDistributedTracingConfig().isEnabled()){
             if (priority.get() == null){
+                //The "if" check above is required even though we do compareAndSet(null) below.
+                //Its purpose is to avoid running the sampler more than once for the same txn.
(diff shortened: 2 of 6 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/tracing/samplers/AdaptiveSampler.java
```diff
@@ -0,0 +1,123 @@
...
+    protected AdaptiveSampler(int target, int reportPeriodSeconds){
+        this.target = target;
+        this.reportPeriodMillis = reportPeriodSeconds * 1000L;
+        this.startTimeMillis = System.currentTimeMillis();
+        this.seen = 0;
+        this.seenLast = 0;
+        this.sampledCount = 0;
+        this.sampledCountLast = 0;
+        this.firstPeriod = true;
+        NewRelic.getAgent().getLogger().log(Level.FINE, "Started Adaptive Sampler with sampling target " + this.target + " and report period " +
```
jbedell-newrelic, line 36: to be more clear, add "seconds" to the end of the message here?

jbedell-newrelic (dismissed):

jbedell-newrelic (approved):

Issue #1872, opened 2024-04-24
Add a configuration option to support spring transaction naming with Controller Class Name + Method


obenkenobi:
Provide a config option to name all Spring Transaction based on `ControllerClassName.MethodName` instead of using the request mappings. 

This is to help resolve Spring Controller URI naming schemes of which we decided [not to support](https://github.com/newrelic/newrelic-java-agent/issues/126) (for reasons such as preventing too many transaction names or security concerns).

PR #2532, opened 2025-10-22
Add configuration option for Spring transaction naming with /ControllerClassName/Method

sharvath-newrelic:
### Overview
This PR introduces a new configuration option `use_controller_class_and_method_for_spring_transaction_naming` (which defaults to `false` to maintain backwards compatibility) that allows Spring Controller transactions to be named using the controller class name and method name (e.g., `/CustomerController/edit`) instead of request mapping URIs (e.g., `/api/v1/customer (POST)`).

Updated two Spring instrumentation modules to support the new configuration: Spring 4.3.0 and Spring 6.0.0.

Implementation details:
  - Checks config value: `class_transformer.use_controller_class_and_method_for_spring_transaction_naming`
  - When enabled, extracts controller class name and method name
  - Sets transaction name using conventional notation: `ControllerClassName/methodName`
  - Uses `TransactionNamePriority.FRAMEWORK_HIGH`
  - Takes precedence over `enhanced_spring_transaction_naming` configuration

### Related Github Issue
#1872 

### Testing
Two new unit test files have been created to supports these changes for both modules. Each test suite includes 5 test cases:
  - Basic controller method naming with config enabled
  - Precedence over `enhanced_spring_transaction_naming` when both enabled
  - Handling URL parameters
  - An alternative HTTP method (POST)
  - Backward compatibility when config disabled (for existing behavior)

Manual e2e testing was conducted via the PetClinic sample application with a `TestController` utilizing endpoints with the `@RequestMapping` annotation. Verification of the expected behavior was conducted successfully with the toggling of the `use_controller_class_and_method_for_spring_transaction_naming` config property to `true` producing the expected `Class/methodName` formatting.

### Checks

### DIFF
--- a/instrumentation/spring-6.0.0/src/main/java/com/nr/agent/instrumentation/AbstractHandlerMethodAdapter_Instrumentation.java
+++ b/instrumentation/spring-6.0.0/src/main/java/com/nr/agent/instrumentation/AbstractHandlerMethodAdapter_Instrumentation.java
@@ -36,7 +36,7 @@ public class AbstractHandlerMethodAdapter_Instrumentation {
                     NewRelic.getAgent().getConfig().getValue("class_transformer.use_controller_class_and_method_for_spring_transaction_naming", false);
 
             if (useControllerClassAndMethodForNaming) {
-                SpringControllerUtility.setTransactionNameUsingControllerClassAndMethod(transaction, controllerClass, controllerMethod);
+                SpringControllerUtility.assignTransactionNameFromControllerAndMethod(transaction, controllerClass, controllerMethod);
             } else {
                 //If this setting is false, attempt to name transactions the way the legacy point cut
                 //named them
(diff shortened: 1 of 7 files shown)


instrumentation/spring-6.0.0/src/main/java/com/nr/agent/instrumentation/AbstractHandlerMethodAdapter_Instrumentation.java
```diff
@@ -34,42 +31,54 @@ protected ModelAndView handleInternal(HttpServletRequest request,
...
 
-            String httpMethod = request.getMethod();
-            if (httpMethod != null) {
-                httpMethod = httpMethod.toUpperCase();
+            // If the new config is enabled, always use controller class + method name with dot notation
+            if (useControllerClassForNaming) {
+                String controllerName = controllerClass.getSimpleName();
+                String methodName = controllerMethod.getName();
+                String txnName = controllerName + "." + methodName;
+                transaction.setTransactionName(com.newrelic.agent.bridge.TransactionNamePriority.FRAMEWORK_HIGH, false, "SpringController", txnName);
```
sharvath-newrelic: Perhaps this logic is better placed in `SpringControllerUtility`.

sharvath-newrelic:
> I went back and looked at this and I think you might be able to use the `SpringControllerUtility::assignTransactionNameFromControllerAndMethod()` to set the txn name with that config flag turned on(?)
> 
> That returns a String like `/MyController/doSomething` which is what you want I think.

Thanks for the insight. I will go ahead and make the change to use that function.

- [ ] Shift to use `assignTransactionNameFromControllerAndMethod()`

Issue #2581, opened 2025-11-14
UnsupportedOperationException: third-party implementation of CancellableContinuation when using Kotlin coroutines with New Relic Java agent


JeHuiPark:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description
When running a Kotlin coroutine–based application with the New Relic Java agent (v8.25.0) attached, the process throws an exception inside a coroutine `delay()` call:

```
java.lang.UnsupportedOperationException: third-party implementation of CancellableContinuation is not supported
    at kotlinx.coroutines.CancellableContinuationKt.invokeOnCancellation(CancellableContinuation.kt:240)
    at kotlinx.coroutines.CancellableContinuationKt.disposeOnCancellation(CancellableContinuation.kt:417)
    at kotlinx.coroutines.EventLoopImplBase.scheduleResumeAfterDelay(EventLoop.common.kt:238)
    at kotlinx.coroutines.DelayKt.delay(Delay.kt:126)
    ...
```

The exception occurs only when the agent is active (`-javaagent:/newrelic/newrelic-agent.jar`).
If the agent is removed, the coroutine executes normally without any errors.

It seems that the agent’s bytecode instrumentation wraps or modifies the coroutine continuation classes, causing the kotlinx.coroutines runtime to detect a “third-party implementation” and throw this `UnsupportedOperationException`.

## Expected Behavior
[NOTE]: # ( Tell us what you expected to happen. )

Coroutines should operate normally with the New Relic Java agent attached.  

## [Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results
[NOTE]: # ( Provide any other relevant log data. )
[TIP]:  # ( Scrub logs and diagnostic information for sensitive information )

Relevant log excerpt:

```
Picked up JAVA_TOOL_OPTIONS: -javaagent:/newrelic/newrelic-agent.jar ...
com.newrelic INFO: Using default collector host: collector.newrelic.com
Exception in thread "DefaultDispatcher-worker-1" java.lang.UnsupportedOperationException:
third-party implementation of CancellableContinuation is not supported
```

When the same build runs **without** the agent, this exception does not appear.

## Steps to Reproduce
[NOTE]: # ( Please be as specific as possible. )
[...]

PR #2582, opened 2025-11-14
fixes Coroutine error about third party implementation of Cancellable…

dhilpipre:
…Continuation

There is an error being generated by the Coroutines instrumentation where it throws an error because the instrumentation tries to implement the CancellableCotinuation interface and Coroutines disallows this.
Workaround is to disable the reporting of calls to the delay function.  It can be configured in newrelic.yml

&nbsp;&nbsp;coroutines:
&nbsp;&nbsp;&nbsp;&nbsp;delayed:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;enabled: false

This code removes the class and references to it so that the error no longer is generated.

Resolves: https://github.com/newrelic/newrelic-java-agent/issues/2581


jtduffy (approved):

dhilpipre:
@kanderson250 no, it is good to go.  Just need someone to push it since I don't have permissions to do so

Issue #2748, opened 2026-02-10
Generalize metric and event ingest URIs.


jeffalder:
## Is your feature request related to a problem? Please describe.
The EU-region metric and event ingest hostnames are hardcoded, which will make it difficult to expand into future regions.

## Feature Description
We should revise the hostnames to follow the standard region-aware ingest domain pattern, which is `<region>.nr-data.net`, making the default region-aware hostnames `metric-api.<region>.nr-data.net` and `insights-collector.<region>.nr-data.net`. These hostnames are currently provisioned for `eu01`.

## Describe Alternatives
1. Continue hardcoding
2. Use the connect payload to distribute hostnames

## Priority
Must Have

PR #2749, opened 2026-02-10
Implement region aware event/metric ingest URIs

jasonjkeller:
Resolves: https://github.com/newrelic/newrelic-java-agent/issues/2748

We weren't parsing the region from the `license_key` and applying it to the metric and event ingest URIs. This PR fixes that.

Do not merge until after the 9.1.0 release.


jeffalder:
Can you revise one of the tests to have a shorter prefix? I've been using `czx` as in the cross-agent tests.

deleonenriqueta (approved):

Issue #2777, opened 2026-03-03
Serverless Implementation: Auto-enable serverless mode from Lambda environment


sharvath-newrelic:
### Description
Per the agent configuration section of the Lambda [spec](https://source.datanerd.us/agents/agent-specs/blob/main/Lambda.md#agent-configuration), the agent should auto-detect that it's running inside an AWS Lambda environment and enable serverless mode accordingly. It should also start successfully when no license key is present in that mode.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-530437

Issue #2702, opened 2026-01-16
Serverless Implementation: Implement Distributed Tracing


obenkenobi:
### Description
The lambda instrumentation should accept dt headers via 2 methods.

HTTP invocation and (if possible) direct lambda invocations via the sdk instrumentation.

### Acceptance Criteria
Java apps that did http calls to a lambda function or via a java app should have distributed traces linking the 2 apps.

### Design Consideration/Limitations
Accepting distributed traces via sdk invocation should be optional.

PR #2782, opened 2026-03-04
Serverless transaction naming, web transactions, and DT

obenkenobi:
### Overview
Accepts inbound DT headers, marks a function as a web transaction based on event source type, provides a standard naming scheme for transaction naming.

The instrumentation was split into 2 modules, the core module and the events module.

Also if the environment variable `AWS_LAMBDA_FUNCTION_NAME` is not empty, serverless mode is enabled 

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/2702
https://github.com/newrelic/newrelic-java-agent/issues/2777


instrumentation/aws-lambda-java-core-1.1.0/src/main/java/com/nr/instrumentation/lambda/LambdaHelper.java
```diff
@@ -0,0 +1,164 @@
...
+
+    /**
+     * Captures Lambda metadata.
+     *
+     * @param context The Lambda execution context
+     * @param event The Lambda event object (can be null or any supported event type)
+     * @return true if metadata was captured successfully, false otherwise
+     */
+    public static boolean startTransaction(Context context, Object event) {
+        nameTransaction(context);
```
sharvath-newrelic: Should this be moved below the `context` guard check (lines 52 - 55)?

PR #2792, opened 2026-03-11
Micronaut fixes

dhilpipre:
Fixes problems related to https://new-relic.atlassian.net/browse/NR-523865

Fixed problems related to excessive number of unexpired async tokens.


instrumentation/micronaut-core-4.0.0/src/main/java/com/nr/agent/instrumentation/micronaut/NRBiConsumerTokenWrapper.java
```diff
@@ -0,0 +1,40 @@
...
+ *  * Copyright 2025 New Relic Corporation. All rights reserved.
+ *  * SPDX-License-Identifier: Apache-2.0
+ *
+ */
+
+package com.nr.agent.instrumentation.micronaut;
+
+import com.newrelic.agent.bridge.AgentBridge;
+import com.newrelic.api.agent.NewRelic;
+import com.newrelic.api.agent.Token;
```
sharvath-newrelic: `Token` is imported here, but not used/referenced within the context of this class. I'm curious if this is dead code, because this no longer appears to be a token wrapper.

instrumentation/micronaut-http-1.3.0/src/main/java/com/nr/agent/instrumentation/micronaut/http/MicronautSubresourceInstrumentation.java
```diff
@@ -137,6 +140,8 @@ private static void instrumentation() {
             }
 
             StringBuilder sb = new StringBuilder();
+            sb.append(methodName);
+            sb.append(" - ");
```
sharvath-newrelic, line 144: This change ended up breaking a few AITs (which was easy to reconcile), is there a precedent for this formatting modification?

instrumentation/micronaut-http-1.3.0/src/main/java/com/nr/agent/instrumentation/micronaut/http/MicronautSubresourceInstrumentation.java
```diff
@@ -146,9 +151,8 @@ private static void instrumentation() {
             if (value != null) {
                 sb.append(value);
             }
-
-            sb.append(" (").append(methodName).append(") ");
-            NewRelic.getAgent().getTransaction().setTransactionName(TransactionNamePriority.FRAMEWORK_HIGH, true, "MicronautController", sb.toString());
+            String txnName = sb.toString().replace("//", "/");
+            NewRelic.getAgent().getTransaction().setTransactionName(TransactionNamePriority.FRAMEWORK_LOW , false, "MicronautController", txnName);
```
sharvath-newrelic, line 155: Why has the transaction name priority been downgraded?

sharvath-newrelic:
I noticed two new HTTP client modules were added (`micronaut-http-client-3.5.0` and `micronaut-http-client-4.0`). These are valuable additions, but perhaps worthy of a separate PR that's not fix focused.

sharvath-newrelic:
It's worth noting that building/running a agent JAR off of this branch results in passing Micronaut [draft](https://github.com/newrelic/java-agent-integration-tests/pull/306) AITs, whereas they were failing prior.

instrumentation/micronaut-core-4.0.0/src/main/java/com/nr/agent/instrumentation/micronaut/NRBiConsumerTokenWrapper.java
dhilpipre: missed that when I was cleaning up the code.

instrumentation/micronaut-http-1.3.0/src/main/java/com/nr/agent/instrumentation/micronaut/http/MicronautSubresourceInstrumentation.java
dhilpipre: yeah, it turned out that there was an inconsistency in transaction naming.  depending on the flow some transactions would get method name at the end and some at the beginning.  This one was changed to be the same as the other transaction.  We can go either way, I just think it should be consistent.

instrumentation/micronaut-http-1.3.0/src/main/java/com/nr/agent/instrumentation/micronaut/http/MicronautSubresourceInstrumentation.java
dhilpipre: this code was overriding a previously set that was better suited name to describe the transaction.  At a minimum the override flag needed to be changed to get rid of the override.  Perhaps the downgrade isn't truly needed.  Probably changed it to ensure it didn't override

dhilpipre:
@sharvath-newrelic 
removed unused import of Token
removed the two new modules and will create a new PR for them.
let me know how you think we should proceed on the transaction naming comments

dhilpipre:
moved new Http Client modules to PR: https://github.com/newrelic/newrelic-java-agent/pull/2834

instrumentation/micronaut-http-1.3.0/src/main/java/com/nr/agent/instrumentation/micronaut/http/MicronautSubresourceInstrumentation.java
sharvath-newrelic: Agreed. If this is following a standard transaction convention then this change should be fine.

instrumentation/micronaut-http-1.3.0/src/main/java/com/nr/agent/instrumentation/micronaut/http/MicronautSubresourceInstrumentation.java
sharvath-newrelic: Understood. Was just curious about the rationale. I don't believe the downgrade is a deal breaker, and I would defer to your judgment here.

sharvath-newrelic (approved):

PR #2705, opened 2026-01-20
Add Kotlin coroutines instrumentation tests

sharvath-newrelic:
### Overview
Per this PR, classloading issues resulting from the porting of Kotlin lab code have been resolved. Additionally, unit tests for the Kotlin coroutines instrumentation modules (1.4, 1.7 and 1.9) have been added with assertions to verify NR agent behavior. Tests help to verify:
- Transaction completion counts
- Transaction name presence
- Metric generation for instrumented operations
- Token propagation across async boundaries

It's worth noting `async.kt` and `PauseService.kt` were not ported for the following reasons:

  1. `async.kt:`
    - Contained example code demonstrating async/await patterns but without assertions
    - All patterns have been incorporated into `CoroutinesTest.kt` with Introspector-based verification
    - The new tests cover the same scenarios (async, awaitAll, multiple deferreds) with agent behavior validation
  2. `PauseService.kt`:
    - No longer needed as new tests use `delay()` directly within test methods

### Related Github Issue
Resolves #2675
Resolves #2661

### Testing
To run the instrumentation tests manually, run the following command template via the terminal: 
`./gradlew instrumentation:kotlin-coroutines-1.x:test --parallel`

PR #2802, opened 2026-03-17
Fix coroutines tasks

dhilpipre:
Enhancements to the current set of Kotlin Coroutines instrumentation

- Better tracking of asynchronous behavior by instrumenting gaps that were found where execution can jump to another thread.
- Better naming on Coroutines and Continuations.  Existing naming was too generic.  New naming identifies according to whether it is started via runBlocking, async, withContext or launch.
- removed some utility methods that were no longer used


sharvath-newrelic:
@dhilpipre These changes are looking good on first pass. Currently we do not have a test suite in `main` to run against changes to Kotlin Coroutines instrumentation. I'm inclined to want to run [these](https://github.com/newrelic/newrelic-java-agent/pull/2705) newer tests here before proceeding with approval, but they are in a PR stage and could use some vetting.

dhilpipre:
@sharvath-newrelic Steven, if changes are needed after May 29 then reach out to Gulab Sidhwani for assistance @gsidhwani-nr

sharvath-newrelic (approved):

PR #2834, opened 2026-04-10
support for Micronaut Http Clients versions 3.5.0+

dhilpipre:
Added support for Micronaut Http clients versions 3.5.0 and later


instrumentation/micronaut-http-client-3.5.0/build.gradle
```diff
@@ -0,0 +1,25 @@
+dependencies {
+    implementation(project(":agent-bridge"))
+    implementation("io.micronaut:micronaut-http-client:3.5.0")
+    implementation("io.projectreactor:reactor-core:3.4.8")
+}
+
+jar {
+    manifest {
+        attributes 'Implementation-Title': 'com.newrelic.instrumentation.micronaut-http-client-3.0.0',
```
sharvath-newrelic: Should this implementation title attribute read as `com.newrelic.instrumentation.micronaut-http-client-3.5.0`?

instrumentation/micronaut-http-client-4.0.0/src/main/java/com/newrelic/instrumentation/micronaut/http/client4/Utils.java
```diff
@@ -0,0 +1,52 @@
...
+	public static URI getRequestURI(HttpRequest<?> request) {
+		
+		InetSocketAddress serverAddress = request.getServerAddress();
+		URI reqURI = request.getUri();
+
+		String schema = reqURI != null ? reqURI.getScheme() : null;
+		if(schema == null) schema = "http";
+		String host = serverAddress.getHostName();
+		int port = serverAddress.getPort();
+		String path = "/";   //request.getPath();
```
sharvath-newrelic, line 44: Is the intent to leave `request.getPath();` commented out?

instrumentation/micronaut-http-client-3.5.0/src/main/java/com/nr/agent/instrumentation/micronaut/http/client3/ReactorListener.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public ReactorListener(Transaction t, HttpParameters p) {
+        txn = t;
+        params = p;
+    }
+
+    @Override
+    public void run() {
+        if (segment != null) {
+            segment.end();
```
sharvath-newrelic, line 32: I noticed there is no `null` assignment here for segment and no try/catch as there is in the 4.0.0 variation. Is this intentional?

sharvath-newrelic:
The new `micronaut-http-client-4.0.0` files are missing copyright headers.

instrumentation/micronaut-http-client-3.5.0/build.gradle
dhilpipre: yes, changed

instrumentation/micronaut-http-client-4.0.0/src/main/java/com/newrelic/instrumentation/micronaut/http/client4/Utils.java
dhilpipre: @sharvath-newrelic So I changed it to "/" because if I left it as the path, it would show any parameter values in the URL when it reported the external call.  I'm not aware of any way that I can get around it.
For example, the server processes the path as /hello/{name}, that's how it shows up on the server side but on the client side would display the call as /hello/Doug.   Concern is leaking PMI data plus possible metric explosion. 
Unless you have a suggestion I would like to keep it as "/" and remove the comment

instrumentation/micronaut-http-client-3.5.0/src/main/java/com/nr/agent/instrumentation/micronaut/http/client3/ReactorListener.java
dhilpipre: No, it should be set to null and I have made that change

instrumentation/micronaut-http-client-4.0.0/src/main/java/com/newrelic/instrumentation/micronaut/http/client4/Utils.java
sharvath-newrelic: This approach sounds reasonable.

sharvath-newrelic (approved):

PR #2842, opened 2026-04-16
Add instrumentation for Spring AI completion and embedding clients

jasonjkeller:
Resolves https://github.com/newrelic/newrelic-java-agent/issues/2770

This PR adds two new instrumentation modules:
- `spring-ai-client-chat-1.0.0`
- `spring-ai-model-1.0.0`

These modules add support for Spring AI with the [New Relic AI Monitoring](https://docs.newrelic.com/docs/ai-monitoring/intro-to-ai-monitoring/) functionality.

## `spring-ai-client-chat-1.0.0`

This module instruments the Spring AI `ChatClient` to capture data from LLM chat completions and generate the `LlmChatCompletionMessage` and `LlmChatCompletionSummary` events that drive the New Relic AI Monitoring UI.

## `spring-ai-model-1.0.0`

This module instruments the Spring AI `EmbeddingModel` to capture data from LLM embeddings and generate the `LlmEmbedding` events that drive the New Relic AI Monitoring UI.

---

Instrumentation verifier runs:
- [spring-ai-model-1.0.0](https://github.com/newrelic/newrelic-java-agent/actions/runs/25121840051/job/73624366671#logs)
- [spring-ai-client-chat-1.0.0](https://github.com/newrelic/newrelic-java-agent/actions/runs/25120982296/job/73621313821#logs)

PR #2845, opened 2026-04-21
Update aws-bedrock-runtime-2.20 Module

### LINKED ISSUE #2839: Update aws-bedrock-runtime-2.20 Module
### Description
Update the token counting strategy for each of the LLM models in  `aws-bedrock-runtime-2.20`.

### Acceptance Criteria
- [ ] Implement token extraction via the LlmTokenCountResolver for all 5 LLM Models
- [ ] All tests pass
- [ ] Update docs

### Additional Context
- [Agent Spec](https://source.datanerd.us/agents/agent-specs/pull/737)
- [DACI - Using LLM Supplied Token Counts](https://newrelic.atlassian.net/wiki/spaces/AIO11y/pages/4106649639/DACI+-+Using+LLM+Supplied+Token+Counts).

deleonenriqueta:
Updates the token counting strategy for each of the models. 

Follows the requirements outlined in the [agent spec]([Agent Spec](https://source.datanerd.us/agents/agent-specs/pull/737)). 
Resolves [#2839](https://github.com/newrelic/newrelic-java-agent/issues/2839)

### DIFF
--- a/instrumentation/aws-bedrock-runtime-2.20/src/main/java/llm/models/ModelResponse.java
+++ b/instrumentation/aws-bedrock-runtime-2.20/src/main/java/llm/models/ModelResponse.java
@@ -107,19 +107,19 @@ public interface ModelResponse {
      *
      * @return Integer representing the prompt token count or null if not provided by the model
      */
-    Integer getPromptTokens();
+    Integer getResponseUsagePromptTokens();
 
     /**
      * Get the number of tokens in the completion/response from the LLM response usage metadata.
      *
      * @return Integer representing the completion token count or null if not provided by the model
      */
-    Integer getCompletionTokens();
+    Integer getResponseUsageCompletionTokens();
 
     /**
      * Get the total number of tokens (prompt + completion) from the LLM response usage metadata.
      *
      * @return Integer representing the total token count or null if not provided by the model
      */
-    Integer getTotalTokens();
+    Integer getResponseUsageTotalTokens();
 }
--- a/instrumentation/aws-bedrock-runtime-2.20/src/main/java/llm/models/ModelResponse.java
+++ b/instrumentation/aws-bedrock-runtime-2.20/src/main/java/llm/models/ModelResponse.java
@@ -101,4 +101,25 @@ public interface ModelResponse {
             NewRelic.getAgent().getLogger().log(Level.FINEST, "AIM: Unable to parse empty/null " + fieldBeingParsed + " from ModelResponse");
         }
     }
+
+    /**
+     * Get the number of tokens in the prompt/request from the LLM response usage metadata.
+     *
+     * @return Integer representing the prompt token count or null if not provided by the model
+     */
+    Integer getPromptTokens();
+
+    /**
+     * Get the number of tokens in the completion/response from the LLM response usage metadata.
+     *
+     * @return Integer representing the completion token count or null if not provided by the model
+     */
+    Integer getCompletionTokens();
+
+    /**
+     * Get the total number of tokens (prompt + completion) from the LLM response usage metadata.
+     *
+     * @return Integer representing the total token count or null if not provided by the model
+     */
+    Integer getTotalTokens();
 }
--- a/instrumentation/aws-bedrock-runtime-2.20/src/main/java/llm/models/ai21labs/jurassic/JurassicModelInvocation.java
+++ b/instrumentation/aws-bedrock-runtime-2.20/src/main/java/llm/models/ai21labs/jurassic/JurassicModelInvocation.java
@@ -89,7 +89,7 @@ public class JurassicModelInvocation implements ModelInvocation {
 
     @Override
     public void recordLlmChatCompletionSummaryEvent(long startTime, int numberOfMessages) {
-        // Calculate hasCompleteUsage here for interface compatibility
+
         boolean hasCompleteUsage = LlmTokenCountResolver.hasCompleteUsageData(
                 modelResponse.getPromptTokens(),
                 modelResponse.getCompletionTokens(),
@@ -121,7 +121,6 @@ public class JurassicModelInvocation implements ModelInvocation {
                 .error()
                 .duration(System.currentTimeMillis() - startTime);
 
-        // Only add usage fields if complete (all-or-nothing rule)
         if (hasCompleteUsage) {
             summaryBuilder
                     .responseUsagePromptTokens()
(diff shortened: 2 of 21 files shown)


instrumentation/aws-bedrock-runtime-2.20/src/main/java/llm/models/ModelResponse.java
```diff
@@ -101,4 +101,25 @@ static void logParsingFailure(Exception e, String fieldBeingParsed) {
...
+     * @return Integer representing the completion token count or null if not provided by the model
+     */
+    Integer getCompletionTokens();
+
+    /**
+     * Get the total number of tokens (prompt + completion) from the LLM response usage metadata.
+     *
+     * @return Integer representing the total token count or null if not provided by the model
+     */
+    Integer getTotalTokens();
```
jasonjkeller: Would you mind refactoring these interface methods to be named `getResponseUsagePromptTokens`, `getResponseUsageCompletionTokens`, and `getResponseUsageTotalTokens` so that it follows the existing pattern of matching the `LlmEvent` builder method and attribute names? This will also match the implementation in the SpringAI instrumentation [here](https://github.com/newrelic/newrelic-java-agent/pull/2842/changes#diff-ed95ffa12ccf824bd112ed086c1c623e72d116c6b1dc623ba9a35c9049797438).

jasonjkeller (changes requested):
Couple of questions and small changes requested, otherwise mostly looks good!

instrumentation/aws-bedrock-runtime-2.20/src/main/java/llm/models/ai21labs/jurassic/JurassicModelInvocation.java
```diff
@@ -58,6 +58,12 @@ public void recordLlmEmbeddingEvent(long startTime, int index) {
             reportLlmError();
         }
 
+        boolean hasCompleteUsage = LlmTokenCountResolver.hasCompleteUsageData(
+                modelResponse.getPromptTokens(),
+                modelResponse.getCompletionTokens(),
+                modelResponse.getTotalTokens()
+        );
```
jasonjkeller: With [LlmEmbedding](https://source.datanerd.us/agents/agent-specs/blob/main/artificial_intelligence/LLMs.md#llmembedding-events) events we are only required to collect a value for `response.usage.total_tokens`. The `modelResponse.getPromptTokens` and `modelResponse.getCompletionTokens` methods won't be implemented (or they may just return 0 or `null` for embeddings) as they are related to chat completions and a default value of 0 can just be used in their places in the `LlmTokenCountResolver`. 

This is the case with SpringAI since there are two instrumentation modules, one that handles embeddings and another that handles chat completions. So the SpringAI embeddings instrumentation doesn't even have methods to get prompt/completion tokens. The AWS instrumentation differs in that one instrumentation module handles both and thus all of these `ModelResponse` methods are implemented regardless if they are needed by each LlmEvent type.

In this instrumentation, I assume that `modelResponse.getPromptTokens` and `modelResponse.getCompletionTokens` would return `null` when parsing an embedding response? Wouldn't this cause the `LlmTokenCountResolver.hasCompleteUsageData` method to always return false?

instrumentation/aws-bedrock-runtime-2.20/src/main/java/llm/models/ai21labs/jurassic/JurassicModelInvocation.java
```diff
@@ -70,7 +76,10 @@ public void recordLlmEmbeddingEvent(long startTime, int index) {
                 .input(index)
                 .requestModel()
                 .responseModel()
-                .tokenCount(getTokenCount(modelRequest.getModelId(), modelRequest.getInputText(index)))
+                .tokenCount(LlmTokenCountResolver.getMessageTokenCount(
```
jasonjkeller: Is `Modellnvocation.getTokenCount` still used anywhere after this change? It looks like the logic was essentially replaced by `LlmTokenCountResolver.getMessageTokenCount`. If it's not used anymore, then can we delete that method from `Modellnvocation`?

jasonjkeller: Another thing that is confusing me, I don't think this `tokenCount` method is the correct one to call on the builder for an `LlmEmbedding` event. Perhaps it was altered in the flurry of agent spec changes that have occurred since we first added the Bedrock instrumentation several years ago, but this `tokenCount` method is going to add an attribute named `token_count` to an `LlmEmbedding` event and that is not in the spec: https://source.datanerd.us/agents/agent-specs/blob/main/artificial_intelligence/LLMs.md#llmembedding-events

According to the spec, the correct attribute for `LlmEmbedding` events should be `response.usage.total_tokens` which would be added by the `responseUsageTotalTokens` builder method. So I think that needs to be changed on every `*ModelInvocation` implementation.

deleonenriqueta: Correct. We are no longer using ModelInvocation.getTokenCount anywhere. I'll delete the method.

instrumentation/aws-bedrock-runtime-2.20/src/main/java/llm/models/ai21labs/jurassic/JurassicModelInvocation.java
deleonenriqueta: Great catch! And yes since those fields don't exist in the embedding response `hasCompleteUsageData` will always return false and `total_tokens` will never get set with this implementation. 

Looking back, I treated all event types the same way regardless of whether it was a chat completion summary event or an embedding event. I'll separate the logical paths so that chat completion summary events attach the 3 `response.usage.` fields and embedding events attaches only `response.usage.total_tokens`. Chat completion message events will continue to use `token_count` set to 0 when the summary has complete usage data or the callback value.

instrumentation/aws-bedrock-runtime-2.20/src/main/java/llm/models/ai21labs/jurassic/JurassicModelInvocation.java
deleonenriqueta: This is directly related to the earlier comment of refactoring embedding events in the same way that I was refactoring the chat completion summary events. I'll go through each of the model invocations and replace the `tokenCount()` method with `responseUsageTotalTokens()` on the embedding builder so the correct attributes get set.

jasonjkeller (approved):
LGTM

PR #2858, opened 2026-04-24
When PointCut tracers were created via CompletableFutures, they were …

### LINKED ISSUE #2844: [Spike] Investigate another "Absurd Values" report
We have a customer that is having problems with absurd values and has provided new logs and info.

jbedell-newrelic:
…not getting their own TransactionActivity, and instead used their parent's.  So, if 2 did that and finished in the "wrong" order, we would get Inconsistent State errors.  If there were more than a couple of threads in that situation, it could cause the transaction to be reported up multiple times, every time a new thread finished.  And any of that could cause Negative and/or Absurd Values to be reported.

Resolves: https://github.com/newrelic/newrelic-java-agent/issues/2844

### DIFF
--- a/instrumentation/java.completable-future-jdk11/src/main/java/util/TokenAwareRunnable.java
+++ b/instrumentation/java.completable-future-jdk11/src/main/java/util/TokenAwareRunnable.java
@@ -2,6 +2,7 @@ package util;
 
 import com.newrelic.agent.bridge.AgentBridge;
 import com.newrelic.agent.bridge.Transaction;
+import com.newrelic.api.agent.Trace;
 
 import static util.TokenAndRefUtils.*;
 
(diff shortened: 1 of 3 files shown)


jtduffy (approved):

jasonjkeller:
Needs some investigation as to whether adding `@Trace(async=true)` to the `run` method will cause a lot of unwanted segments in traces (due to them being root tracers).

jbedell-newrelic:
These changes are not quite enough to fix the problem properly, and, in fact, can introduce other issues.  The real fix here is going to be to remove pointcuts for good, or to spend the time updating the pointcut tracer-creation path to account for TokenAndRefCount functionality, much like the weaver's tracer-creation path does.

PR #2859, opened 2026-04-28
Resolution of Mule 4.9 IllegalAccessError issue

### LINKED ISSUE #2836: Create POC to bypass JPMS restrictions and resolve Mule 4.9 IllegalAccessException
## Description
This work is predicated on [this](https://new-relic.atlassian.net/browse/NR-518055) Jira ticket. NR Labs has created an experimental module for Mule 4.9, and found that when running it on the required JDK 17, `IllegalAccessException`s are propagated due to JPMS encapsulation. This task is to explore a solution within the agent where this exception can be mitigated when paired with the experimental module.

## Additional context
[Technical breakdown and reproduction
](https://docs.google.com/document/d/1znPkF1DVyYDIr8VtW0Frdn7JshEU1Evo_jvPrHI1ISo/edit?tab=t.0)

sharvath-newrelic:
### Overview
POC to fix `IllegalAccessError`s when instrumenting Mule 4.9 on Java 17+ by bypassing JPMS. 

Technical outline(s) of issue: [Jira](https://new-relic.atlassian.net/browse/NR-518055?focusedCommentId=1578052), [Google Docs](https://docs.google.com/document/d/1znPkF1DVyYDIr8VtW0Frdn7JshEU1Evo_jvPrHI1ISo/edit?tab=t.0).

Example error: 
```
 java.lang.IllegalAccessError:                                                                                                                                                                                     
    class org.mule.runtime.core.internal.event.AbstractEventContext                                       
    (in module org.mule.runtime.core)                                                                                                                                                                               
    cannot access class                                                                                   
    com.newrelic.weave.org.mule.runtime.core.internal.event.AbstractEventContext_1827233194_nr_ext
    (in unnamed module @0x44ca8399)
    because module org.mule.runtime.core does not read unnamed module @0x44ca8399
```

Lab's experimental Mule 4.9 instrumentation throws `IllegalAccessException`s when attempting to access Mule runtime classes across module boundaries.

This POC adds a `JpmsModuleHelper` class, which serves as a helper called from `WeavePackageManager` right after utility class injection. On the first `org.mule.runtime.*` class instrumented, it scans loaded classes and collects named Mule modules, then calls `Instrumentation.redefineModule` to help grant these modules read access. For Java 8 compatibility these JPMS calls are reflection-based. 

### Related Github Issue
Resolves #2836 

### Testing
15 unit tests have been added to cover new scenarios.

### DIFF
--- a/newrelic-weaver/src/main/java/com/newrelic/weave/weavepackage/JpmsModuleHelper.java
+++ b/newrelic-weaver/src/main/java/com/newrelic/weave/weavepackage/JpmsModuleHelper.java
@@ -1,0 +1,113 @@
+/*
+ *
+ *  * Copyright 2026 New Relic Corporation. All rights reserved.
+ *  * SPDX-License-Identifier: Apache-2.0
+ *
+ */
+
+package com.newrelic.weave.weavepackage;
+
+import java.lang.instrument.Instrumentation;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+class JpmsModuleHelper {
+
+    private static final Method GET_MODULE;
+    private static final Method IS_NAMED;
+    private static final Method GET_ALL_LOADED;
+    private static final Method GET_UNNAMED;
+    private static final Method REDEFINE_MODULE;
+
+    private static final AtomicBoolean mulePatched = new AtomicBoolean();
+
+    static {
+        Method getModule = null, isNamed = null, getAllLoaded = null,
+               getUnnamed = null, redefineModule = null;
+        try {
+            Class<?> modCls = Class.forName("java.lang.Module");
+            getModule = Class.class.getMethod("getModule");
+            isNamed = modCls.getMethod("isNamed");
+            getAllLoaded = Instrumentation.class.getMethod("getAllLoadedClasses");
+            getUnnamed = ClassLoader.class.getMethod("getUnnamedModule");
+            redefineModule = Instrumentation.class.getMethod("redefineModule",
+                    modCls, Set.class, Map.class, Map.class, Set.class, Map.class);
+        } catch (Throwable ignored) {
+            // Java 8: leave null
+        }
+
+        GET_MODULE = getModule;
+        IS_NAMED = isNamed;
+        GET_ALL_LOADED = getAllLoaded;
+        GET_UNNAMED = getUnnamed;
+        REDEFINE_MODULE = redefineModule;
+    }
+
+    /**
+     * Triggers a one-shot Mule module patch on the first {@code org/mule/runtime/} class.
+     * No-op on Java 8, for non-Mule classes, and after the first call.
+     */
+    static void addReadsToUnnamedModule(Instrumentation inst, String className, ClassLoader cl) {
+        if (GET_MODULE == null || inst == null || cl == null || className == null) return;
+        if (!className.startsWith("org/mule/runtime/")) return;
+        if (!mulePatched.compareAndSet(false, true)) return;
+        try {
+            patchMuleModules(unwrap(inst), cl);
+        } catch (Throwable ignored) {}
+    }
+
+    // Single-pass scan: adds reads(unnamed + all named modules) to every named org.mule.runtime.* module.
+    private static void patchMuleModules(Instrumentation inst, ClassLoader cl) throws Exception {
+        Object unnamed = GET_UNNAMED.invoke(cl);
+        Set<Object> reads = new HashSet<>(Collections.singleton(unnamed));
+        Set<Object> muleModules = new HashSet<>();
+
+        for (Class<?> cls : loadedClasses(inst)) {
+            Object module = moduleOf(cls);
+            if (!isNamed(module)) continue;
+
+            reads.add(module);
+            if (cls.getName().startsWith("org.mule.runtime.")) {
+                muleModules.add(module);
+            }
+        }
+
+        for (Object module : muleModules) {
+            REDEFINE_MODULE.invoke(inst, module, reads,
+                    Collections.emptyMap(), Collections.emptyMap(),
+                    Collections.emptySet(), Collections.emptyMap());
+        }
+    }
+
+    private static Instrumentation unwrap(Instrumentation inst) {
+        for (Class<?> cls = inst.getClass(); cls != null && cls != Object.class; cls = cls.getSuperclass()) {
+            try {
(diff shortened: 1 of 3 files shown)


newrelic-weaver/src/main/java/com/newrelic/weave/weavepackage/JpmsModuleHelper.java
```diff
@@ -0,0 +1,113 @@
...
+        REDEFINE_MODULE = redefineModule;
+    }
+
+    /**
+     * Triggers a one-shot Mule module patch on the first {@code org/mule/runtime/} class.
+     * No-op on Java 8, for non-Mule classes, and after the first call.
+     */
+    static void addReadsToUnnamedModule(Instrumentation inst, String className, ClassLoader cl) {
+        if (GET_MODULE == null || inst == null || cl == null || className == null) return;
+        if (!className.startsWith("org/mule/runtime/")) return;
```
sharvath-newrelic, line 57: This namespacing (specifically `runtime`) looks to be restricted to Mule 4.x.

newrelic-weaver/src/main/java/com/newrelic/weave/weavepackage/JpmsModuleHelper.java
```diff
@@ -0,0 +1,113 @@
...
+            try {
+                Field field = cls.getDeclaredField("delegate");
+                field.setAccessible(true);
+                Object delegate = field.get(inst);
+                return delegate instanceof Instrumentation ? unwrap((Instrumentation) delegate) : inst;
+            } catch (NoSuchFieldException ignored) {
+            } catch (Throwable t) { break; }
+        }
+        return inst;
+    }
```
sharvath-newrelic, line 95: This looks a bit unintuitive, and is somewhat fragile as it relies on a field being named `delegate` with the proxy structure not changing. This function is walking down the chain of delegate fields until it hits the object at the bottom (`sun.instrument.InstrumentationImpl`), which is the only one that can execute `redefineModule()`.

newrelic-weaver/src/main/java/com/newrelic/weave/weavepackage/JpmsModuleHelper.java
sharvath-newrelic, line 57: Ideally this becomes extensible/agnostic in the future. Perhaps instead of a single Mule focused prefix, this could be replaced with a string set of known problematic module roots and `JpmsModuleHelper` becomes more of an orchestrator. Maybe this could be tied into the config so adding a prefix without additional code changes is possible:
```
jpms:
  patched_module_prefixes: "org/mule/runtime"
 ```

Comment (jtduffy) [reply newrelic-weaver/src/main/java/com/newrelic/weave/weavepackage/JpmsModuleHelper.java:57]: Yes, I think this is a great idea. Might be worth opening a ticket so we don't forget about it.

**Review by jtduffy — DISMISSED** [review]
Very nice

File: newrelic-weaver/src/main/java/com/newrelic/weave/weavepackage/JpmsModuleHelper.java [hunk 3]
```diff
@@ -0,0 +1,113 @@
...
+            }
+        }
+
+        for (Object module : muleModules) {
+            REDEFINE_MODULE.invoke(inst, module, reads,
+                    Collections.emptyMap(), Collections.emptyMap(),
+                    Collections.emptySet(), Collections.emptyMap());
+        }
+    }
+
```
sharvath-newrelic, line 84: There are a few points worth elaborating on regarding this function. The initial loop conducts a single pass over every loaded class and for each class in a named module, that module is added to reads. If the class is a Mule runtime class, its module is recorded as one that requires patching. 

In the second loop `redefineModule` is called on each Mule module and grants reads access to every named module as well as the unnamed module. So why does reads contain all named modules? Adding a reads edge is one-directional regarding access. In adding all named modules to reads, the patch allows Mule modules to see both the agent's unnamed module as well as other named modules they might need (e.g. access to logging.level, which cropped up as initially inaccessible during testing), which helps to cover cross-module patterns.

newrelic-weaver/src/main/java/com/newrelic/weave/weavepackage/JpmsModuleHelper.java
kanderson250: I also like the idea of broadening this to other modules and for this to be configurable in some way. 

A complication is that the weaver does not have access to configs - if we want to do this in the future, that could be a reason to consider moving this out of the weaver and into the agent.

newrelic-weaver/src/main/java/com/newrelic/weave/weavepackage/JpmsModuleHelper.java
kanderson250: I'd like to push a little further on this - we should try to follow least-privilege when modifying access and only do it where it is strictly required. 

Do we still have evidence that weaved classes in mule modules are trying and failing to read anything besides the unnamed module (like the logging.level example you give)? Is there a way for us to reproduce this with the current fix?

newrelic-weaver/src/main/java/com/newrelic/weave/weavepackage/JpmsModuleHelper.java
kanderson250: Alternatives to consider:

- Get the field from `DelegatingInstrumentation` reflectively, and invoke it on `inst` without iterating up the hierarchy. 
- Move the JpmsModuleHelper out of the weaver and into the agent, where it will have access to `DelegatingInstrumentation`, and `inst` can be cast to `DelegatingInstrumentation` without reflection. 
- ...other possibilities?

If you prefer to leave this method as-is, a comment explaining where the `delegate` field comes from and why we iterate up the class hierarchy would be good.

sharvath-newrelic: I like the initial alternative you've offered for its clarity of purpose (even with the coupling to `com.newrelic.agent.util.DelegatingInstrumentation`). I'll go ahead and tweak this.

newrelic-weaver/src/main/java/com/newrelic/weave/weavepackage/JpmsModuleHelper.java
sharvath-newrelic: Agreed. I'm glad you pushed back, as after retesting with mods to just allow reads for the unnamed module with the new Labs Mule 4.9 instrumentation, I did not encounter any `IllegalAccessError`s. It's entirely possible early testing went awry and this was added defensively. I'll go ahead and update this to account for a narrower grant.

kanderson250 (approved):
Looks good, nice work!

PR #2873, opened 2026-05-11
AWS DAX Support

### LINKED ISSUE #2800: Dynamo instrumentation when using DAX
What is the problem the customer is trying to solve?

Customer would like to add support for AWS DAX. When running AWS Dax with DynamoDB, customers can no longer see DynamoDB-related metrics in the NR APM

Why is it important for users that we fix that problem?

By adding support for AWS DynamoDB DAX client metrics into New Relic APM, customer development team will gain valuable insights to optimize database performance, reduce latency, and enhance user experience. This integration will enable data-driven decision making, improve error detection and troubleshooting, and optimize resource allocation.

What is the objective they are pursuing?

- Improve database performance analysis

- Enhance error detection and troubleshooting

- Optimize resource allocation and cost optimization

As customer services relies more heavily on DynamoDB, it's crucial for our observability tools to provide real-time insights into database performance. This integration will enable customers to scale with confidence and meet business requirements.

jtduffy:
Resolves #2800 

Add supports for AWS [DAX](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DAX.html).

This is accomplished by instrumenting the `software.amazon.dax.MetricAsyncClient` class, which is used by both the sync and async clients.

Config caching and other info is detailed in the README.


instrumentation/aws-java-sdk-dax-2.0.0/src/main/java/software/amazon/dax/MetricAsyncClient_Instrumentation.java
```diff
@@ -0,0 +1,98 @@
...
+
+import java.util.concurrent.CompletableFuture;
+import java.util.logging.Level;
+
+@Weave(originalName = "software.amazon.dax.MetricAsyncClient", type = MatchType.ExactClass)
+abstract class MetricAsyncClient_Instrumentation {
+
+    MetricAsyncClient_Instrumentation(Configuration configuration) {
+        NewRelic.getAgent().getLogger().log(Level.FINEST, "AWSDAX: Construction of new MetricAsyncClient");
+        DAXClientConfigCache.storeConfiguration(this, configuration);
```
kanderson250: Did we consider storing the config via `@NewField`? 

I think the caching strategy here works perfectly well, but this strategy (keying with `this` to internal state of a weaved class) is already what `@NewField` does, and in that case maybe we would want to lean on it rather than build a module-specific cache. If you would prefer not to use `@NewField` or have another concern with it, the cache can stay as is.

instrumentation/aws-java-sdk-dax-2.0.0/src/main/java/com/nr/instrumentation/dynamodb_dax/DAXUtil.java
```diff
@@ -0,0 +1,107 @@
...
+            return null;
+        }
+        String accountId = getAccountId(sdkClient, configuration);
+        if (accountId == null) {
+            NewRelic.getAgent().getLogger().log(Level.FINEST, "AWSDAX: Unable to assemble ARN. Unable to retrieve account information.");
+            return null;
+        }
+        // arn:${Partition}:dynamodb:${Region}:${Account}:table/${TableName}
+        NewRelic.getAgent().getLogger().log(Level.FINEST, "AWSDAX: Returning arn");
+        return "arn:aws:dynamodb:" + region + ":" + accountId + ":table/" + tableName;
```
kanderson250, line 82: I know the dynamodb ARN is intentional per the README, but as a nit, consider adding a small comment here too.

instrumentation/aws-java-sdk-dax-2.0.0/src/main/java/software/amazon/dax/MetricAsyncClient_Instrumentation.java
jtduffy: I was messing with an idea to use a localized cache rather than the `@NewField` mega-cache because of the cache maintenance issues we sometimes run into.

I think you're right, for this, using `@NewField` makes more sense since there will be very entries for this module (maybe 2-4).

kanderson250 (approved):

PR #2874, opened 2026-05-11
Create default CLAUDE file

jasonjkeller:
Adds a simple default Claude file to the project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,12 @@ production/
 instrumentation/play-2.6/src/test/scala/router/*.scala
 instrumentation/play-2.6.13/src/test/scala/router/*.scala
 
+# Always ignore local Claude settings and overrides
+.claude/settings.local.json
+CLAUDE.local.md
+
+# Ignore local application data (if stored within the project root)
+# Note: These are often kept in your home directory (~/.claude/) instead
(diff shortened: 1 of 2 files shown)


jtduffy (dismissed):

jtduffy:
I was looking at this again and had a couple of thoughts...
Would it be worth adding the following to the .gitignore file as part of this PR?

- `CLAUDE.local.md` - This allows each individual to provide instructions to claude that might be specific to their environment.
- the `.claude` folder - This looks like it contains JSON permissions file that is specific to each devs environment? Someone can double check me on this.

jasonjkeller:
> I was looking at this again and had a couple of thoughts... Would it be worth adding the following to the .gitignore file as part of this PR?
> 
> * `CLAUDE.local.md` - This allows each individual to provide instructions to claude that might be specific to their environment.
> * the `.claude` folder - This looks like it contains JSON permissions file that is specific to each devs environment? Someone can double check me on this.

@jtduffy Good call. I've updated the `.gitignore` file. It sounds like there may be some use cases for certain files in the `.claude` folder getting checked in, so I've excluded limited things but not the entire directly. We can certainly continue to refine this.

jtduffy (approved):

Issue #2775, opened 2026-03-03
Support for Kafka Transactions in Apache Camel Kafka Integration


elucus:
Support for Kafka Transactions in Apache Camel Kafka Integration. See internal ticket for details...

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-529812

PR #2878, opened 2026-05-13
Apache Camel support and Support for Kafka Clients 4.x.x

obenkenobi:
### Overview
This PR adds support for Apache Camel 3.9.0 and up and Kafka Clients 4.0.0 and up

#### What is Apache Camel?

Apache Camel is an integration framework that acts as a producer/client for other data sources (i.e. an http client, database access, message brokers, etc) and can act as a generic consumer for data sources and can reroute data to other destinations via its producer/client APIs.

Camel routes function as the cornerstone of the framework as it allows data from one source to another via a fluent API.

#### What is the scope of the new instrumentation?

In order to support the producer/client aspect of Camel, Kafka clients 4.0.0 and up had to be supported.

To support the consumer aspect of camel, transactions are created for Camel routes alongside accepting inbound headers and reporting the inbound data source as an external trace.

#### Camel Exchanges 

The unit of work for camel is represented by the Exchange object. Exchanges are created for producer/client calls and for consumer routes. 

#### How the new instrumentation works

For consumer calls, we have to flag that exchange. We then have to start transactions in the Processor.process(...) method if the exchange is from a consumer and the exchange does not contain a token. 

When starting a transaction, inbound headers are read and transactions are named based on the camel endpoint. Camel exchanges may actually have child exchanges. This occurs when batch consumers are used for Kafka. When that happens, transaction naming and inbound headers are read by iterating into the first child exchange. 

If Processor.process(...) is not starting transactions, this method mainly links tokens inside the Exchange or creates a new one if a transaction already exists. The latter action is done to support the producer/client APIs of Camel in case of a thread hop. 

We expire the token inside the exchange when it is completed.

#### Exclusions

Camel routes that are sourced from logs and internal processes are excluded from starting transactions.

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/2775


instrumentation/apache-camel-3.9.0/src/test/java/com/nr/instrumentation/apache/camel/processors/KafkaExchangeProcessorTest.java
```diff
@@ -0,0 +1,120 @@
+/*
+ *
+ *  * Copyright 2026 New Relic Corporation. All rights reserved.
+ *  * SPDX-License-Identifier: Apache-2.0
+ *
+ */
+
+spackage com.nr.instrumentation.apache.camel.processors;
```
jtduffy: typo in "package", so this test is never going to compile/run

instrumentation/apache-camel-3.9.0/src/test/java/com/nr/instrumentation/apache/camel/CamelUtilTest.java
```diff
@@ -0,0 +1,98 @@
...
+    @Test
+    public void endpointOperation() throws IOException {
+        String url = "kafka:topicA?brokers=localhost:9092";
+        try (Endpoint endpoint = mockEndpoint(url)) {
+            String operation = CamelUtil.endpointOperation(endpoint);
+            assertEquals("kafka", operation);
+        }
+    }
+
+    public void testKafkaProcessor() throws IOException {
```
jtduffy, line 42: This method and two others are missing the `@Test` annotation

instrumentation/apache-camel-3.9.0/src/main/java/com/nr/instrumentation/apache/camel/processors/ExchangeProcessor.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+package com.nr.instrumentation.apache.camel.processors;
+
+import com.newrelic.agent.bridge.Transaction;
+import org.apache.camel.Exchange;
+
+public interface ExchangeProcessor {
+    void nameTransaction(Transaction transaction, Exchange exchange);
+    boolean shouldStartTransaction();
```
jtduffy, line 15: I doesn't look like this method is ever called, except in tests?

jtduffy (changes requested):

obenkenobi:
Thanks @jtduffy! I just updated the PR with the corrections

obenkenobi:
Changes are pushed to a new branch

PR #2880, opened 2026-05-14
Add support for AI Model Context Protocol

### LINKED ISSUE #2479: Add support for AI Model Context Protocol
### Description
[Model Context Protocol (MCP)](https://modelcontextprotocol.io/docs/getting-started/intro) is a specification that defines how Large Language Models (LLMs) can securely interact with external data, tools, and systems. MCP defines a structured interface for tools, apps, and data resources that models can use to access information in a controlled manner. Given the rapid increase in adoption of MCP to build servers and clients, language agents should produce named spans to represent MCP calls and their results.

[Agent spec](https://source.datanerd.us/agents/agent-specs/blob/main/artificial_intelligence/MCP.md)

deleonenriqueta:
Resolves [#2479](https://github.com/newrelic/newrelic-java-agent/issues/2479)

This PR adds a new instrumentation module: `mcp-java-sdk-1.0.0 `, adding support for the [Model Context Protocol (MCP) Java SDK](https://github.com/modelcontextprotocol/java-sdk).                                                                                                                       
                                                                                                                                                                              
This module instruments `McpAsyncClient` to create Llm/ segments for MCP operations when `ai_monitoring.enabled` is set. The following methods are instrumented across both clients: 
                                                                                                                                                           
  - `callTool → Llm/tool/MCP/call_tool/{tool_name}`
  - `readResource → Llm/resource/MCP/read_resource/{uri_scheme}`
  - `getPrompt → Llm/prompt/MCP/get_prompt/{prompt_name}`
---
Instrumentation verifier run:                                                                                                                                                
  - [mcp-java-sdk-1.0.0](https://github.com/newrelic/newrelic-java-agent/actions/runs/25876815354)


jtduffy:
I wonder if this could be simplified by only instrumenting the async client. I know the sync ends the segment by calling end() and the async calls endAsync(), but I think endAsync() will work fine for the sync client calls.

The only difference that I can see is that the endAsync() calls the finish() method via a runnable submission to a thread pool and end() simply calls finish inline.

If you've already tested this and it didn't work, then ignore me. :-)  We did a similar thing in the DAX module so I thought it might apply here and simplify things a bit (for DAX, we're not using segments, but were able to only instrument the async client).

deleonenriqueta:
Great catch! Thanks for that feedback @jtduffy 

I removed `McpSyncClient_Instrumentation` entirely. Since `McpSyncClient` delegates internally to `McpAsyncClient`, all sync calls are covered. Validated both sync and async paths in APM (correct segment creation and no duplicate reporting).

jtduffy (approved):

PR #2899, opened 2026-05-21
Sql comment metadata updates per spec

jtduffy:
This updates the agent's implementation of SQL metadata comments to match the finalized agent [spec](https://source.datanerd.us/agents/agent-specs/pull/796). Specifically:

- Only the `nr_service_guid` is supported
- The config has changed to only required a simple `enabled` flag: `transaction_tracer.sql_metadata_comments.enabled`
- A transaction is no longer required for the comment to be added


agent-bridge-datastore/src/main/java/com/newrelic/agent/bridge/datastore/JdbcHelper.java
```diff
@@ -430,57 +393,23 @@ private static String generateSqlMetadataComment(Set<String> metadataCommentConf
...
-        if (metadataCommentConfig == null) {
+    private static Boolean isSqlMetadataCommentsEnabled() {
+        if (isSqlMetadataCommentsEnabled == null) {
             synchronized (JdbcHelper.class) {
-                if (metadataCommentConfig == null) {
+                if (isSqlMetadataCommentsEnabled == null) {
                     try {
-                        Set<String> config = parseSqlMetadataCommentsConfig();
+                        Boolean isEnabled = NewRelic.getAgent().getConfig().getValue("transaction_tracer.sql_metadata_comments.enabled");
                         // Only cache if we got a non-null value (config service is ready)
```
obenkenobi, line 402: This is a breaking change in how `transaction_tracer.sql_metadata_comments` is formatted. What is the expected customer impact with this one?

obenkenobi:
I recommend adding an additional tests in JdbcHelperTest (or any other place that seems to be approptiate appropriate). 

The scenarios include:

`transaction_tracer.sql_metadata_comments.enabled=true` adds a comment to a SQL string such that it is formatted correctly.

and

`transaction_tracer.sql_metadata_comments.enabled=false` provides no comment.

agent-bridge-datastore/src/main/java/com/newrelic/agent/bridge/datastore/JdbcHelper.java
jtduffy: There is no customer usage of this feature yet so it won't break anyone

jtduffy:
I tried to do this earlier in a unit test but it was incredibly difficult to get everything mocked properly. I do have an AIT which exercises this.

obenkenobi (dismissed):

jbedell-newrelic (approved):

PR #2900, opened 2026-05-21
chore: Initial work to add agent config schema

mvicknr:
### Overview
Adds a script to automatically generate and update a configuration schema for the agent. This PR has the initial schema generated from the script.

The script will automatically run on local build and on pushes (except main). It will determine if there are config changes and if so, it will update the schema.On local, the dev should include these changes in their commit. In the PR build, if the changes aren't already made in the schema, it will make the changes and add them to the PR.

The schema version will be updated during the release process.

### Related Github Issue
n/a

### Testing
GH workflow tested locally with act

### Checks

Co-Authored-By: Claude <noreply@anthropic.com>


.fleetControl/schemas/config.json
mvicknr, line 1: this is the new schema for fleet control to use

.fleetControl/schemas/config.json
```diff
@@ -73,6 +78,79 @@
             "default": "newrelic_agent.log",
             "description": "The name of the log file. Default is newrelic_agent.log."
         },
+        "datastore_multihost_preference": {
+            "type": "string",
+            "default": "NONE",
+            "description": "When multiple datastore addresses detected, determines which to use. Values: NONE, FIRST, LAST Default: NONE"
```
mvicknr, line 89: If these are the only valid values, this should be an enum like `log_level` on line 42. It can be added to `ENUM_OVERRIDES` in the generator

.fleetControl/schemas/config.json
```diff
@@ -174,6 +324,84 @@
...
+                },
+                "detect_gcp": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "Poll GCP metadata API. Default: true"
+                },
+                "detect_kubernetes": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "Report KUBERNETES_SERVICE_HOST env var. Default: true"
```
mvicknr, line 588: I didn't realize there were k8s-specific configs. If later, we want a schema for on-host, do you know the set of configs that are k8s-specific?

.fleetControl/schemas/config.json
```diff
@@ -201,6 +463,28 @@
...
+            "properties": {
+                "enabled": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "Enable custom event collection. Default: true"
+                },
+                "max_attribute_value": {
+                    "type": "integer",
+                    "default": 255,
+                    "description": "Max attribute value length (max 4095). Default: 255"
```
mvicknr, line 677: We should be able to have a [maximum](https://json-schema.org/draft/2020-12/json-schema-validation#name-maximum) value

.fleetControl/schemaGeneration/generate-schema.groovy
mvicknr: shouldn't this use the new `reference-newrelic.yml`?

.fleetControl/schemas/config.json
jtduffy: Good call out, I'll update the script

.fleetControl/schemas/config.json
jtduffy: This is it as far as I know. If this is `true` (the default) we attempt to detect the container ID from the a couple of places and send that up as part of the utilization payload.

jtduffy (approved):

PR #2907, opened 2026-05-27
Update site information for Micronaut, Kafka, and Camel modules

obenkenobi:
### Overview
Update site blocks for Micronaut, Kafka, and Camel modules for compatability docs generation.


sharvath-newrelic (dismissed):

instrumentation/micronaut-http-1.3.0/build.gradle
```diff
@@ -20,6 +20,6 @@ verifyInstrumentation {
 }
 
 site {
-    title 'Micronaut'
+    title 'Micronaut HTTP'
     type 'Framework'
```
kanderson250: Should this (and other micronaut http modules) have `type: Http` instead?

kanderson250 (changes requested):

instrumentation/micronaut-http-1.3.0/build.gradle
obenkenobi: @sharvath-newrelic thoughts?

obenkenobi: @kanderson250 I updated the site blocks but micronaut-http is a serverside library.

Micronaut-HTTP-Client is however an HTTP client so I updated the type to be HTTP

kanderson250 (approved):

PR #2911, opened 2026-05-28
Netty replacement

dhilpipre:
While working with a customer, we discover that some RASP (Runtime Application Self-Protection) products view the way the Java Agent inserts the NettyDispatcher class in between pipeline handlers as a security risk and blocks it.

The code in this pull request handles it somewhat differently.   Because the Netty framework API allows you to insert and remove a ChannelHandler, it uses a custom New Relic ChannelHandler.   In the decoder class, instead of calling NettyDispatcher, it gets the current ChannelHander and inserts the NRNettyChannelHandler after it. So when the current ChannelHandler finishes, the pipeline invokes the channelRead method on NRNettyChannelHandler.  When the channelRead method finishes on NRNettyChannelHandler, the ChannelHandler that was originally next is invoked.  RASP does not block this action because it is an allowed action.
The NRNettyChannelHandler class channelRead method mimics the code in NettyDispatcher.
When the channel read is complete, the channelReadComplete method is called in the NRNettyChannelHandler class and it removes itself from the pipeline.


dhilpipre:
After May 29th, please reach out to Gulab Sidhwani for assistance to resolve any issues or to answer any questions

Issue #2263, opened 2025-03-10
Investigate an alternative to Caffeine weak ref map


meiao:
### Description
It has happened that the weak ref map used to cache values in the agent has caused thread locks, preventing customer requests from completing.

https://github.com/raphw/weak-lock-free may provide an alternative.
It seems it would also cause locking if multiple threads try to write to the map, but this may alleviate some of the problems seen with Caffeine.

ben-manes:
If you are running into locking, then you might consider setting a high `initialCapacity`. ConcurrentHashMap will create a lock per bin, meaning the size of the underlying table array. If your map never grows very large, e.g. 16 entries, then there is a higher chance of collisions for a long-running computation. As this is merely an array. presizing it is very inexpensive for memory and provides a higher degree of write concurrency.

I don't see how `weak-lock-free` is any different since it is not lock-free. While they mention that `ConcurrentMap` may optionally use locks, like ConcurrentHashMap does, they fail to mention that their call to `ReferenceQueue` on every operation is also lock-based. Prior to Java 8 this was [fully synchronized](https://github.com/openjdk/jdk/blob/77c86a964655e3586a404d3e0803bd8cd6dbbc01/jdk/src/share/classes/java/lang/ref/ReferenceQueue.java#L97-L101) and now uses an optimistic volatile check if empty but locks if populated. That is a fast lock but an incorrect claim of being lock-free and impacts every operation. In the Java 5 days (Guava), I solved this by reusing the amortization approach to make a concurrent LRU. This uses buffers to record accesses and replay them under a tryLock to maintain the policy. Caffeine uses an updated version of that scheme, and it means the `ReferenceQueue.poll()` is only accessed by a single thread during maintenance cycles.

That project might be a better fit if you decide to use a lock-free ConcurrentMap like `ConcurrentSkipListMap` or `NonBlockingHashMap`, with the necessary caveats involved (such as non-atomic computes). I'd recommend first considering using an `initialSize` or decoupling the computation from the hash table to use `ConcurrentHashMap` more effectively.

PR #2917, opened 2026-06-02
Remove linking metadata blob NewFields from log4j-2.11 module

jtduffy:
### Overview
Relates to #2263

This removes the linking metadata `@NewField` annotations from the (now deleted) `LogEvent_Instrumentation` class. The extension classes were created regardless if the `NewField`s were actually used. This could cause thrashing on the extension class cache and result in high CPU usage or deadlocks for very high logging volume.

Instead, the module will now store the linking metadata blob in the Log4j MDC (ThreadContext) if app logging and local decorating are enabled.

It accomplishes this by instrumenting the `AbstractLogger.logMessageSafely` method, which is the final "sink" for all loggable messages. The blob is then pulled out in the `AbstractJacksonLayout.toSerializable` method. This is the location where the `NewField` values we're previously read.

All existing logging related AITs pass.


instrumentation/apache-log4j-2.11/src/main/java/org/apache/logging/log4j/spi/AbstractLogger_Instrumentation.java
```diff
@@ -0,0 +1,30 @@
...
+import org.apache.logging.log4j.message.Message;
+
+import static com.newrelic.agent.bridge.logging.AppLoggingUtils.*;
+
+@Weave(originalName = "org.apache.logging.log4j.spi.AbstractLogger", type = MatchType.BaseClass)
+public abstract class AbstractLogger_Instrumentation {
+    private void logMessageSafely(final String fqcn, final Level level, final Marker marker,
+            final Message message, final Throwable throwable) {
+        if (isApplicationLoggingEnabled() && isApplicationLoggingLocalDecoratingEnabled()) {
+            System.out.println("DUF- Thread: " + Thread.currentThread());
```
jbedell-newrelic: debugging code

instrumentation/apache-log4j-2.11/src/main/java/org/apache/logging/log4j/core/layout/AbstractJacksonLayout_Instrumentation.java
```diff
@@ -40,16 +39,17 @@ public String toSerializable(final LogEvent_Instrumentation event) {
             if (isApplicationLoggingEnabled() && isApplicationLoggingLocalDecoratingEnabled()) {
                 // It is possible that the log might already have NR-LINKING metadata from JUL instrumentation
                 if (!writerString.contains(BLOB_PREFIX)) {
+                    System.out.println("DUF- here1");
```
jbedell-newrelic: debugging code

jbedell-newrelic (approved):

PR #2972, opened 2026-07-07
Update the Scala API cross build flow to support gradle 8.3

jtduffy:
Corrects issues related to publishing to maven (snapshots and releases).

Updates the Maven publication setup for all Scala API modules to be compatible with Gradle 8. and bumps the gradle-crossbuild plugin from 0.14.1 to 0.17.0

To support the new cross build plugin, the publishing config needed to be updated so that components properly resolved and the resulting POMs were correct. In Gradle 8.3, publications must declare the full software component via `from(components[...])` rather than referencing individual task artifacts. This switches each module to derive the component name dynamically from the Scala version string and uses `from(components[componentName])` and removes the explicit task variable bindings.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id 'com.github.prokod.gradle-crossbuild' version '0.14.1' apply false
+    id 'com.github.prokod.gradle-crossbuild' version '0.17.0' apply false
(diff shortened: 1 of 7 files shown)


kanderson250 (approved):
LGTM!

Issue #2962, opened 2026-06-25
Suppress the NR_LAMBDA_MONITORING payload from Java agent for AWS Lambda


jasonjkeller:
Resolves: https://new-relic.atlassian.net/browse/NR-578407

The Java agent for AWS Lambda is writing the json payload of `NR_LAMBDA_MONITORING` to Console log.

We should add a configurable mechanism for suppressing this.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-581974

obenkenobi:
Configuration is not needed. According to the agent spec, writing the extension payload takes precedence and if it is not written to, then STDOUT is written to instead.

PR #2975, opened 2026-07-07
Serverless mode sends payload to console if file pipe not written (#2975)

obenkenobi:
### Overview
Resolves an issue where ServerLess mode writes to both stdout and to the lambda extension file. 
Now in ServerLess mode, writing to file takes precedence and if it fails to write to such a file,  stdout is used instead.

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/2962

### DIFF
--- a/newrelic-agent/src/main/java/com/newrelic/agent/transport/serverless/ServerlessWriterImpl.java
+++ b/newrelic-agent/src/main/java/com/newrelic/agent/transport/serverless/ServerlessWriterImpl.java
@@ -5,27 +5,38 @@ import com.newrelic.agent.logging.IAgentLogger;
 import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileWriter;
-import java.io.IOException;
 import java.util.logging.Level;
 
 public class ServerlessWriterImpl implements ServerlessWriter {
     private final IAgentLogger logger;
-    private final File pathFile;
+    private File pathFile;
+    private boolean fileCreated;
 
     public ServerlessWriterImpl(IAgentLogger logger, String filePath) {
         this.logger = logger;
-        pathFile = new File(filePath);
+        try {
+            pathFile = new File(filePath);
+            fileCreated = true;
+        } catch (Throwable t) {
+            fileCreated = false;
+            pathFile = null;
+        }
+
     }
 
     @Override
     public synchronized void write(String filePayload, String consolePayload) {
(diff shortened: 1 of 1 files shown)


newrelic-agent/src/main/java/com/newrelic/agent/transport/serverless/ServerlessWriterImpl.java
```diff
@@ -5,27 +5,38 @@
...
 
     public ServerlessWriterImpl(IAgentLogger logger, String filePath) {
         this.logger = logger;
-        pathFile = new File(filePath);
+        try {
+            pathFile = new File(filePath);
+            fileCreated = true;
+        } catch (Throwable t) {
+            fileCreated = false;
+            pathFile = null;
```
sharvath-newrelic, line 23: Is this `catch` block worthy of a logger statement?

newrelic-agent/src/main/java/com/newrelic/agent/transport/serverless/ServerlessWriterImpl.java
```diff
@@ -5,27 +5,38 @@
...
     @Override
     public synchronized void write(String filePayload, String consolePayload) {
-        if (pathFile.exists()) {
+        boolean fileWritten = false;
+        if (fileCreated && pathFile.exists()) {
             try (BufferedWriter pipe = new BufferedWriter(new FileWriter(pathFile, false))) {
                 pipe.write(filePayload);
-            } catch (IOException ignored) {
-                logger.log(Level.FINEST, "Failed to write payload", ignored);
+                fileWritten = true;
```
sharvath-newrelic, line 35: Should `fileWritten` be set to `true` only after we are assured no exception occurred (e.g. write succeeded)?

obenkenobi: Yes. This is to ensure that STDOUT is only written to if we confirm the file is NOT successfully written.

newrelic-agent/src/main/java/com/newrelic/agent/transport/serverless/ServerlessWriterImpl.java
obenkenobi: Good call, I will make a log statement.

sharvath-newrelic (approved):

PR #2982, opened 2026-07-09
New Patcher instance to remove problem enum values from the shadowed jspecify NullMarked class

### LINKED ISSUE #2976: Spring boot 2.8.x fat jars compiled under Java 8 and run in Java 8 runtime crash
### Description
This is from a #java-agent issue. 

When running a spring boot application compiled under Java 8 and running under a Java 8 run time, the app fails to start. This happens regardless if the app is run as a fat jar or an exploded jar.

When launching a spring boot app, the JarLauncher configures the classpath for Spring's PathMatchingResourcePatternResolver class, and adds the agent jar as a candidate for component scanning. The agent shadows both version 2 and 3 of the Caffeine cache framework. Version 3 only supports Java 11 and up, so when Spring scans the agent jar when running under Java 8, it encounters annotations values a class that don't exist in Java 8, which causes the crash. So something in the customer's app is declaring a some sort of "catch-all" component scan pattern that is picking up "com.newrelic" namespaced classes.

The problem class is the `NullMarked` class from [jspecify](https://jspecify.dev/docs/api/org/jspecify/annotations/NullMarked.html), which is a Caffeine dependency. In the version of jspecify that Caffeine 3 uses, the `MODULE` value was added to the `@Target` annotation of that class. `MODULE` doesn't exist in Java 8, which cause the crash when Spring scans the class metadata.

The possible workarounds are:
- Drop to version 9.1.0 of the agent (no caffeine 3)
- Correct the ComponentScan pattern. It's technically an error that a runtime agent is being component scanned by Spring
- Run the app under Java 9+ (Java 11 would be recommended since it's an LTS release)

Simple boot 2.7.18 [repro](https://github.com/jtduffy/spring-boot2-j8-sample).

Possible agent solution:
During the shadow process, do a bytecode rewrite of the jspecify `NullMarked` class to remove the ``MODULE` value from the `@Target` annotation.  Probably also check for/remove `RECORD_COMPONENT` to future proof since it was added in J 16.

jtduffy:
Resolves #2976 

Adds a new Patcher instance that runs during the shadowJar task that removes ElementType.MODULE and ElementType.RECORD_COMPONENT from the @Target annotation on the shaded jspecify NullMarked class. 
Both enum values were introduced after Java 8 and cause ArrayStoreException in Java 8's AnnotationParser 
when Spring scans the agent JAR because of the presence of the shaded Caffeine v3 jar.

Also adds a README that summarizes every Patcher instance's functionality.

### DIFF
--- a/buildSrc/src/main/java/com/nr/builder/patcher/RemoveUnsupportedAnnotationTargets.java
+++ b/buildSrc/src/main/java/com/nr/builder/patcher/RemoveUnsupportedAnnotationTargets.java
@@ -1,6 +1,6 @@
 /*
  *
- *  * Copyright 2020 New Relic Corporation. All rights reserved.
+ *  * Copyright 2026 New Relic Corporation. All rights reserved.
  *  * SPDX-License-Identifier: Apache-2.0
  *
  */
--- a/buildSrc/src/main/java/com/nr/builder/patcher/RemoveUnsupportedAnnotationTargets.java
+++ b/buildSrc/src/main/java/com/nr/builder/patcher/RemoveUnsupportedAnnotationTargets.java
@@ -1,0 +1,82 @@
+/*
+ *
+ *  * Copyright 2020 New Relic Corporation. All rights reserved.
+ *  * SPDX-License-Identifier: Apache-2.0
+ *
+ */
+
+package com.nr.builder.patcher;
+
+import com.nr.builder.Const;
+import com.nr.builder.Patcher;
+import org.objectweb.asm.AnnotationVisitor;
+import org.objectweb.asm.ClassVisitor;
+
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/java/com/nr/builder/patcher/RemoveUnsupportedAnnotationTargets.java
```diff
@@ -0,0 +1,79 @@
+/*
+ *
+ *  * Copyright 2020 New Relic Corporation. All rights reserved.
```
sharvath-newrelic: Minor nit to bump the copyright date.

sharvath-newrelic (approved):

obenkenobi (approved):

Issue #2637, opened 2025-12-10
Add "http.request.method" as a span event attribute


jtduffy:
### Description
The agent [spec](https://source.datanerd.us/agents/agent-specs/blob/8b7a4b36080aca5a0c756e979498be0d561611a4/Span-Events.md#http) calls out several attributes that should be added to an HTTP span event. The agent seems to the missing the `http.request.method` attribute.

Add the `http.request.method`  to external call spans. The method should be the standard HTTP verb, in uppercase : `GET`, `POST`, `PUT`, etc.

Related [FR](https://new-relic.atlassian.net/browse/NR-361711)

PR #2985, opened 2026-07-10
Http request method

obenkenobi:
### Overview
All HTTP external traces now report an agent/span attribute named "http.request.method". As per spec requirements, this attribute reports uppercase HTTP Verbs (i.e. GET, POST, PUT, DELETE, etc).

Every instrumentation modules that reports HTTP external calls (excluding AWS SDK and GRPC instrumentation) are updated to support that attribute.

This is accomplished via the new agent bridge APIs:

```
AgentBridge.getAgent().setHttpMethod(segment, method);
```

and

```
AgentBridge.getAgent().getTracedMethod().setHttpMethod(method);
```

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/2637


instrumentation/async-http-client-2.0.0/src/main/java/com/nr/agent/instrumentation/asynchttpclient/NRAsyncHandler.java
```diff
@@ -43,6 +44,7 @@ public class NRAsyncHandler<T> {
     public Segment segment;
     @NewField
     public URI uri;
+    public String method;
```
jtduffy, line 48: It's hard to follow the flow in a PR, but was this meant to also be a `@NewField`?

jtduffy (changes requested):

jtduffy (dismissed):

jtduffy (approved):

PR #2988, opened 2026-07-12
Use containerd prefix

jeffalder:
### Overview

AL2023 cgroup v2 no longer has the cgroup v1 file. And containerd uses a different format in the `mountinfo` file compared to dockerd.

I've verified that the updated regex matches the container ID reported by the infra agent for this pod.

### Related Github Issue
n/a

### Testing

### Checks


jasonjkeller:
Potentially requires cross agent test and agent spec updates.

jeffalder:
I'm not sure this is actually accurate. I was running a container via docker, not via Kubernetes, so the path appears different. I don't know what the right answer to get the container actually is here.

I'm going to close this for now; sorry I don't have a better answer.

PR #3010, opened 2026-07-21
Resolve verifier issues with Ning and Spray instrumentation modules

obenkenobi:
### Overview
Resolve verifier failures wrt the new http method Agent Bridge API


jtduffy:
Can you give a brief summary of how the previous changed caused a verifier error? I'm genuinely curious.

jtduffy (approved):

PR #3011, opened 2026-07-21
Add clear_return_stacks default for coroutines modules.

### LINKED ISSUE #2981: [Spike] Investigate mechanism to support variable system properties by module
### Description
Assess the viability of supporting the ability to switch a default system property on a per module basis. The Ktor instrumentation is currently blocked pending a means to turn the `clear_return_stacks` flag to `true` for coroutines specific modules. In lieu of a solution along those lines, propose an alternative approach.

sharvath-newrelic:
### Overview
This PR defaults `clear_return_stacks` (the fix for AIOOBE when weaving Kotlin coroutine bytecode, e.g. Ktor's `invokeSuspend` continuations) to on for our Kotlin-coroutines/Ktor instrumentation modules (total of 8), so it applies automatically at runtime instead of requiring `-Dnewrelic.config.class_transformer.clear_return_stacks=true` to be set manually.

An explicitly-set system property (`true` or `false`) still overrides the per-module default in either direction, so existing behavior for anyone already using the global flag is unchanged. Every other instrumentation module keeps its current default (`false`).

Each of the 8 modules self-declares its default via a new `Clear-Return-Stacks-Default: true` jar manifest attribute (alongside its existing `Implementation-Title`), rather than being named in a central allow-list. This follows the same pattern already used by `Weave-Violation-Filter`. `WeavePackageConfig.Builder.jarInputStream(...)` reads the attribute directly into `clearReturnStacksDefault`, and that config is threaded through `ClassWeave` -> `MethodProcessors` -> `MethodCallInlinerAdapter`. Module identity/defaulting is now entirely manifest-driven, so a future per-module default consumed at the same depth just needs a new getter on `WeavePackageConfig` and a manifest attribute on the relevant module(s).

### Related Github Issue
Resolves #2981 

### Testing
5 new tests have been added to attend to permutations on defaulting system properties.


kanderson250 (approved):
Excellent work. LGTM!

Issue #2728, opened 2026-02-02
Add instrumentation support for Redisson Redis Java client


jasonjkeller:
Currently for Redis, we support the Jedis and Lettuce clients. We should also add instrumentation support for the [Redisson Redis Java client](https://github.com/redisson/redisson).

[OTel Redisson instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/redisson) for reference.

Related: 
- https://new-relic.atlassian.net/browse/NR-380613
- https://new-relic.atlassian.net/browse/NR-501478
- https://new-relic.atlassian.net/browse/NR-501243

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-519076

PR #3050, opened 2026-08-10
Support for Redisson 3.28.0 and up

obenkenobi:
### Overview
Adds support for a redis client named Redisson. This is based on the NR Labs approach to the module but with heavy modifications to instead report low level redis operations as they execute.

The module also reports for batch operations in the format:
`BATCH-EXECUTE_{Operation1}_{Operation2}` and so on.

For example: 
`BATCH-EXECUTE_HSET_HSET_INCR`

This is to workaround the fact that redisson batch operations work by caching redisson operations and then executing all at ounce. To avoid multiple segments per each operation, it made more sense to just make a segment for the batch execution and include all redis operations that are executed.

This approach for batching is subject to change based on further reviews of this pull request.

### Related Github Issue
https://github.com/newrelic/newrelic-java-agent/issues/2728


jtduffy (commented):
Missing copyright headers on new files

jtduffy (approved):

PR #3054, opened 2026-08-13
Report Cluster Ids in Kafka Metrics

### LINKED ISSUE #2984: [Spike] Prototype Kafka per-cluster produce/consume metrics
The Java agent team has been asked to prototype adding Kafka per-cluster produce/consume metrics to help inform the following agent spec draft: https://source.datanerd.us/agents/agent-specs/pull/817

The author of the agent spec draft has provided a PR (https://github.com/newrelic/newrelic-java-agent/pull/2914) with a potential Java agent implementation that should be vetted to determine if it is a viable solution, a good starting point, or if a different approach should be taken. On first glance, this PR heavily relies on reflection which we would prefer to avoid where possible.

The desired outcome of this story is to have a fully working Java agent prototype that can be used to inform the agent spec and help drive implementation for other language agents.

Related slack conversation: https://newrelic.slack.com/archives/C119L74HL/p1782158844151989

kanderson250:
Resolves #2984 

This fulfills a spec change request to add per-cluster produce/consume metrics to the agent. 

## Details

According to the spec change, we need to add optional capture of per-cluster, per-topic metrics for Kafka clients. These are meant to exist as a **supplement** to the existing per-node, per-topic client metrics and will be used by the backend to build entity relationships. 

- **Existing Node Topic Metric**: `MessageBroker/Kafka/Node/{server_address}/{Produce | Consume}/{topic_name}`
- **New Cluster Topic Metric**: `MessageBroker/Kafka/Cluster/{cluster_id}/{Produce | Consume}/{topic_name}`

## What's changed

The cluster topic metric has been added to **all eight** kafka clients metrics modules that currently report the analogous Node Topic Metric. 

The implementation is slightly different between `kafka-clients-metrics` and `kafka-clients-node-metrics`, but it is exactly the same across different versions of the same module (I had to backport it to all versions due to current adoption rates for earlier modules being high). It follows the implementation of the analagous `NodeTopicRegistry` as closely as possible.

For all versions, the bulk of the logic for fetching the cluster id and building the metric names lives in `ClusterTopicRegistry`. 

## Important requirements

These changes **MUST**:
- Gate the capture of the metric behind a configuration `kafka.metrics.cluster.metrics.enabled` which defaults to `false`.
- Not report anything if a clusterId is unavailable. 
- After successful fetch of the clusterId, continue to reuse it for the lifetime of the client. 
- Capture the cluster metric **in addition to** (not instead of) the node topic metric. 
- Report the cluster topic metric at the same frequency as the node topic metric (when available). 

## Background (why this change was necessary)

Prior to this change, entity relationships were based on the combination of **server address + topic** from the Node Topic Metric. However, for self-hosted Kafka instances utilizing VPCs, it is possible for separate clusters to report the same broker address AND same topic name, in which case their identities will be conflated. 

This change allows a new relationship to be formed based on **clusterId + topic**. Since cluster ids are always unique to the cluster, this prevents NR from mixing up identities.


instrumentation/kafka-clients-metrics-3.0.0/src/main/java/com/nr/instrumentation/kafka/ClusterTopicRegistry.java
```diff
@@ -0,0 +1,123 @@
...
+public class ClusterTopicRegistry {
+    private static final String CLUSTER_METRIC_FORMAT = "MessageBroker/Kafka/Cluster/{0}/{1}/{2}"; // {clusterId}/{clientType}/{topicName}
+    private final boolean clusterIdMetricsEnabled;
+
+    private final Set<String> recordedTopics = ConcurrentHashMap.newKeySet();
+    private final Set<String> convertedNames = ConcurrentHashMap.newKeySet();
+
+    private final ClientType clientType;
+    private final Metadata metadata;
+    private volatile String clusterId;
```
obenkenobi, line 52: Is it correct that Kafka Consumers and producers can only connect to one cluster at a given time?

obenkenobi (approved):

instrumentation/kafka-clients-metrics-3.0.0/src/main/java/com/nr/instrumentation/kafka/ClusterTopicRegistry.java
kanderson250: Yes. The client metadata is final, and metadata is scoped to a single cluster (the client can supply multiple broker addresses as bootstrap addresses when initialized, but all broker addresses must belong to the same cluster).

PR #3081, opened 2026-08-27
Add collector connection TTL and recovery tests

PongPong:

### Overview

Adds a configurable maximum lifetime for pooled collector connections to prevent reuse of connections that have been silently terminated by an intermediary such as a proxy, firewall, or load balancer.

The new `collector_connection_ttl` setting:

- Accepts a non-negative number of seconds.
- Defaults to `0`, preserving the existing unlimited connection lifetime.
- Can be configured through `newrelic.yml`, the `newrelic.config.collector_connection_ttl` system property, or the `NEW_RELIC_COLLECTOR_CONNECTION_TTL` environment variable.
- Applies only when leasing pooled connections and does not interrupt active requests.

The PR also adds parameterized connection-lifecycle and recovery tests:

- Embedded Jetty covers HTTP and TLS keep-alive reuse, idle connection termination, explicit connection closure, active request abortion, long-running requests, and repeated recovery.
- Toxiproxy covers TCP resets, blackholes, latency, and truncated responses.
- Toxiproxy tests automatically detect Docker and can be explicitly disabled.

### Related GitHub Issue

N/A

### Testing

The following focused test suites were run locally:

- Embedded Jetty connection-lifecycle tests: 11 passed.
- Toxiproxy network-fault tests: 4 passed.
- Apache HTTP client TTL and CRaC tests: 8 passed.
- Collector connection TTL configuration tests: 8 passed.
- Toxiproxy opt-out behavior: 4 tests skipped successfully.
- `git diff --check` passed.

The complete repository test suite was not run locally. GitHub Actions will run the broader suite when the pull request is opened.

### Checks

This PR adds Testcontainers 1.21.3 and its Toxiproxy module as test-only dependencies. They are used exclusively for TCP fault-injection integration tests and are not included in the Java agent runtime or distribution. Jetty was already present as a test dependency.


newrelic-agent/src/main/java/com/newrelic/agent/config/DataSenderConfig.java
```diff
@@ -65,4 +65,8 @@ public interface DataSenderConfig {
     String getLicenseKey();
 
     int getTimeoutInMilliseconds();
+
+    default long getCollectorConnectionTtlInMilliseconds() {
```
jtduffy: This should be a regular interface method, not a `default` for consistency

newrelic-agent/src/main/java/com/newrelic/agent/transport/apache/ApacheHttpClientWrapper.java
```diff
@@ -89,14 +96,20 @@ private static String initUserHeaderValue() {
...
                 RegistryBuilder.<ConnectionSocketFactory>create()
                         .register("http", PlainConnectionSocketFactory.getSocketFactory())
                         .register("https", sslContext != null ?
                                 new SSLConnectionSocketFactory(sslContext) : SSLConnectionSocketFactory.getSocketFactory())
-                        .build());
+                        .build(),
+                null,
+                null,
+                null,
+                collectorConnectionTtlInMillis > 0 ? collectorConnectionTtlInMillis : -1,
```
jtduffy: This can simply be set to the `collectorConnectionTtlInMillis` value. If you look at the eventual use of the TTL value, it's used in the [PoolEntry](collectorConnectionTtlInMillis) class, which does a check for `> 0`, so the default of 0 will be fine here.

newrelic-agent/src/main/resources/newrelic.yml
```diff
@@ -188,6 +188,11 @@ common: &default_settings
   #proxy_password: password
   #proxy_scheme: https
 
+  # Limits the maximum lifetime of a persistent connection to the New Relic collector.
+  # Set this below the shortest proxy or network connection timeout to avoid reusing stale connections.
+  # A value of 0 allows connections to remain open indefinitely. Default is 0 seconds.
+  #collector_connection_ttl: 0
```
jtduffy: I think this is fine to uncomment in the example yml file.

newrelic-agent/src/test/java/com/newrelic/agent/instrumentation/AgentWrapperTest.java
```diff
@@ -88,8 +90,11 @@ private void setupServices(boolean coreServiceEnabled, boolean withTracerFactory
 
         ServiceFactory.setServiceManager(serviceManager);
 
-        PointCutClassTransformer classTransformer = new PointCutClassTransformer(instrProxy, false);
-        target = AgentWrapper.getAgentWrapper(classTransformer);
+        try (MockedStatic<SecurityAgentConfig> securityAgentConfig = Mockito.mockStatic(SecurityAgentConfig.class)) {
```
jtduffy: These changes will need to be reverted. They are not relevant to the TTL change

README.md
```diff
@@ -30,6 +30,41 @@ See the following documentation for specific use cases of the Java agent:
 - [Troubleshooting](https://docs.newrelic.com/docs/agents/java-agent/troubleshooting)
 - [Java agent Javadocs](https://newrelic.github.io/java-agent-api/javadoc/index.html?com/newrelic/api/agent/package-summary.html)
 
+### Collector connection lifetime
```
jtduffy: We rely on the reference yml and public docs for info like this, so it can be removed from the README

newrelic-agent/src/test/java/com/newrelic/agent/transport/apache/ApacheHttpClientWrapperConnectionTtlTest.java
```diff
@@ -0,0 +1,248 @@
+/*
```
jtduffy: As much as we appreciate the addition of the connection tests, we request that they be removed, including the Jetty tests, since we don't use Jetty anywhere in the agent.  It's a lot of code, and we'd rather not add an additional test dependency to support the tests.

This also includes removing the associated `newrelic-agent/build.gradle` changes

jtduffy (changes requested):
Thank you for the PR. I've requested some changes below.

PongPong:
Follow-up on the remaining failed check: `aits / client/http4s.py` is also failing on `main` at `287a3ab` with the same `test_http4s_ember_client` assertion (`Expected {count: 2}`, observed one matching transaction metric): https://github.com/newrelic/newrelic-java-agent/actions/runs/33775746891/job/100718354977

The corresponding failure for this PR commit is here: https://github.com/newrelic/newrelic-java-agent/actions/runs/33777015267/job/100782407867

The HTTP4s AIT exercises the separate HTTP4s Ember client instrumentation; this PR changes the collector transport based on Apache HttpClient. The unit, functional, and instrumentation matrices—including Java 11 `nonForkedTests`—are now passing. Based on the identical baseline failure on `main`, this remaining red check does not appear to have been introduced by the collector connection TTL change. Could the HTTP4s job be rerun or treated as a known baseline failure?

jtduffy (approved):

PongPong:
hi @jtduffy , Do you have a timeline or plan in mind for when the next release tag will be created? No rush at all, just checking in so I can align my downstream tasks.

jtduffy:
Hi @PongPong 

We don't have an exact date, but I think we're tentatively shooting for v9.5.0 to be out no later than the end of September.

This is the [issue](https://github.com/newrelic/newrelic-java-agent/issues/3108) we have to track the progress of the release activities. You can also keep an eye on our release [page](https://github.com/newrelic/newrelic-java-agent/releases).

